Issue #61, opened 2018-12-07
Import optimizations, select series (2. step):


michaelkain:
* Make it more obvious, that details and preview can be displayed
* Add details display for study as well (missing today)
* Correct the alignment: series are on the same level as the study
* Change „Protocol“ in details to „Protocol name:“ (to better relate to the dicom tag)
* Send only selected series to the back, remove the selected boolean from the json

PR #85, opened 2018-12-18
Import optimizations (#61) and css upgrade

Nunien:
* See #61 
* Css upgrades might affect every page. The main risk is with the new select-boxes.


Nunien:
I have to re-merge, then Yao can test it

PR #227, opened 2019-10-24
Migration scripts: neurinfo and ofsep

MSimonInria:


michaelkain:
Hi Mathieu, I do my tests with the branch develop and your cool scripts. MS Datasets nearly finishes only at the end there is a small error, do you maybe have any insights?

Import echo_time: start
Traceback (most recent call last):
  File "./4_dataset_migration.py", line 357, in <module>
    GROUP BY et.ECHO_TIME_ID HAVING count(md.DATASET_ID) = 1""")
pymysql.err.InternalError: (1055, u"Expression #4 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'neurinfo.md.DATASET_ID' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by")

PR #265, opened 2020-03-02
MS Datasets: c-store and wado-uri introduced + bug fixes

michaelkain:
* Integrates c-store and wado-uri into ms datasets to assure backward compatibility with
dcm4chee2 (for production Neurinfo we agreed with the admin to keep for the moment
the current dcm4chee2 to reduce change overhead for installation of new version).
The flag „ dcm4chee-arc.dicom.web“ can be used to switch between c-store+wado-uri
and stowrs+wadors.
* By default now dicom images are sent to the pacs with c-store as tests showed, that
160 images with c-store require 3 secs and 160 images with stowrs require 4 secs in average
* Integrates a new exception handling and stops the „silent“ exceptions in ms datasets
* Installs dcm4che-5.21.0 into ms datasets (Dockerfile), to be used on the command line
* Fixes a major bug in DatasetAcquisitionApiController.java (setImportJob on ImporterService):
-> not multi-thread safe and would lead to bad errors in production
* Fixes a major bug in DatasetsCreatorAndNIfTIConverterService.java in ms import:
-> diffusion datasets (doNotSeparateFlag) have been created with one DatasetExpression
for each DatasetFile (big error), now fixed

### DIFF
--- a/docker-compose/datasets/Dockerfile
+++ b/docker-compose/datasets/Dockerfile
@@ -12,6 +12,15 @@
 
 FROM openjdk:8-jdk-alpine
 
+# Installation of dcm4che to manage the store-scu into the PACS
+# dcm4che3, used last version 5.21.0 as available on the 2020-02-14
+# https://sourceforge.net/projects/dcm4che/files/dcm4che3/
+RUN wget -O dcm4che-5.21.0-bin.zip http://downloads.sourceforge.net/project/dcm4che/dcm4che3/5.21.0/dcm4che-5.21.0-bin.zip
+RUN unzip dcm4che-5.21.0-bin.zip
+
+# take care of path
+ENV PATH /dcm4che-5.21.0/bin:$PATH
+
 RUN mkdir -pv /var/log/shanoir-ng-logs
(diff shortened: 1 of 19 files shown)


yyao1231:
Hi @michaelkain Could you please show me how to test the first point since I don't have dcm4chee2 installed locally? And for other points, if you have some description of special tests to do, it will be great! Thanks a lot!

michaelkain:
Hi Yao,
please test with dcm4chee 5 arc light. Assuming both servers implement well the dicom protocol,
for the moment I see no need to test with 2. No special test, do an import and download and check
everything is the same as for old, concerning the files.
Best regards,
Michael

yyao1231:
Ok. So I misunderstood. Now the c-store and wado-uri will work with both dcm4chee 2 and 5 server if I understand now. And this mode is active by default (it worked only with stor rest? + dcm4chee5 server before)?

yyao1231:
@michaelkain everything is fine. I created 2 issues (#272 #271) after comparing files download between sh-old and sh-ng. But nothing about your pull request :)

PR #266, opened 2020-03-04
selectable rows in tables

Nunien:
**Usage :**

Add ` [selectionAllowed]="true"` to the `<shanoir-table>` tag and then catch selected rows with the `(selectionChange)` event.

**Exemple :** 

* in html

```
<shanoir-table #table
        [getPage]="getPage.bind(this)" 
        [columnDefs]="columnDefs" 
        [customActionDefs]="customActionDefs" 
        (rowClick)="onRowClick($event)"
        [browserSearch]="false"
        [selectionAllowed]="true"
        (selectionChange)="onSelectedChange($event)">
</shanoir-table>

```
* in component

```
onSelectedChange(selected: Dataset[]) {
	console.log(selected);
}
```

**Warning :**

When clicking 'select all' on a backend paging entities table, a request will fetch every single entity ( getAll() ). Else, how could we know what is selected? We probably will set a limit.

### DIFF
--- a/shanoir-ng-front/src/app/coils/coil-list/coil-list.component.html
+++ b/shanoir-ng-front/src/app/coils/coil-list/coil-list.component.html
@@ -15,8 +15,8 @@ along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 <h2 i18n="Manage coils|Title@@coilListTitle">Manage coils</h2>
   
 <shanoir-table #table
-    [getPage]="getPage.bind(this)" 
-    [columnDefs]="columnDefs" 
-    [customActionDefs]="customActionDefs" 
-    (rowClick)="onRowClick($event)">
+        [getPage]="getPage.bind(this)" 
+        [columnDefs]="columnDefs" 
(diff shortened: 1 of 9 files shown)


Nunien:
Now, here is a few questions before going further : 

* When you click 'select all', do you assume everything will be selected or only the items of the current page?
* Is it possible that some users would want to download hundreds or thousands datasets at once?

Nunien:
So, for now :
* select all / none is related to the current page
* filtering reset the selection

yyao1231:
@Nunien j'ai mergé avec ta version. L'erreur que j'ai "app.js:1 Error: TableComponent : if you are going to use the selectionAllowed input your items must have an id. (it's like in a night club)"
lol

Nunien:
@yyao1231 Salut, je viens de faire des test et je pense que ceci devrait fonctionner : 

Il faut que tu implémentes la méthode suivante dans la classe qui représente une ligne des résultats Solr : 

public get id(): number {
    return this.datasetId;
}

Dis moi si ça fonctionne.

PR #275, opened 2020-03-27
massive download back-end init

yyao1231:
- Init massiveDownload API
- Check that the connected user has the given right for at least one of the given dataset list
- Remove from the datasetIds list the ids the not found datasetIds and the not allowed datasetIds

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApi.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApi.java
@@ -20,12 +20,14 @@ package org.shanoir.ng.dataset.controler;
 
 import java.io.IOException;
 import java.net.MalformedURLException;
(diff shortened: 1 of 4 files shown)


yyao1231:
Tested with solr results in the front-end on yyao1231:solr branch (already merged with fli-iam:develop).
The pull request could also be tested with unit tests only in the back-end if you don't want to be updated with the solr front-end committed on solr branch.

michaelkain:
Hi, a really good job! Excellent, exactly what I was thinking. I have made a visual review only until now, but it looks good. Maybe in your loop on the datasetIds, when a datasetId could not be found in the database directly throw your exception: if (!notFoundDatasetIds.isEmpty()) throw new EntityNotFoundException("Cannot find dataset with id: " + notFoundDatasetIds.toString()); This could be faster?

yyao1231:
Yes I did like you proposed but it could not throw an exception in the loop, that's why I did like this :)

PR #276, opened 2020-03-30
Mass upload - back end + bugfix

jcomedouteau:


michaelkain:
Hi Jean-Come, when I checkout your PR for testing, I get the following error: ERROR: Couldn't find env file: /Users/mkain/git/shanoir-ng/docker-compose/keycloak/variables.env, when calling docker-compose command

jcomedouteau:
Actually, these changes on keycloak and .env shouldn't be in that branch. Anyway I pushed what was missing.
Jean-Côme

Issue #16, opened 2018-10-15
Saving takes some time... sometimes


Nunien:
TODO : Have a loader on save buttons. A boolean can be set in modeSpecificSave().

Nunien:
Added loaders

PR #305, opened 2020-05-25
Studycards - Working

Nunien:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/security/DatasetSecurityService.java
```diff
@@ -63,7 +75,69 @@ public boolean hasRightOnStudy(Long studyId, String rightStr) {
...
+    /**
+	 * Check that the connected user has the given right for the given study card.
+	 * 
+	 * @param studyCardId the study card id
+	 * @param rightStr the right
+	 * @return true or false
+     * @throws EntityNotFoundException 
+	 */
+    public boolean hasRightOnStudyCard(Long studyCardId, String rightStr) throws EntityNotFoundException {
+    	if (KeycloakUtil.getTokenRoles().contains("ROLE_ADMIN")) return true;
```
jcomedouteau, line 88: On pourra en discuter pour les prochains dev (là je pense que c'est un peu tard :) )
https://github.com/fli-iam/shanoir-ng/wiki/Code-convention-and-best-practices

michaelkain:
Hi Julien, I have started to test your PR and clean everything (I use a script to do it, so I start from scratch). I can not test easily, as ms users does not start with the following error, please see below. The weird thing is, that in .env I have set spring.jpa.hibernate.ddl-auto=create and
spring.datasource.initialize=true, but it does not seem to apply? Do you have any idea?
By the way, could you take care of the conflicts coming from the merge with new search?

Caused by: **org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: missing table [account_request_info]**
	at org.hibernate.tool.schema.internal.SchemaValidatorImpl.validateTable(SchemaValidatorImpl.java:67)
	at org.hibernate.tool.schema.internal.SchemaValidatorImpl.doValidation(SchemaValidatorImpl.java:50)
	at org.hibernate.tool.hbm2ddl.SchemaValidator.validate(SchemaValidator.java:91)
	at org.hibernate.internal.SessionFactoryImpl.<init>(SessionFactoryImpl.java:475)
	at org.hibernate.boot.internal.SessionFactoryBuilderImpl.build(SessionFactoryBuilderImpl.java:444)
	at org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl.build(EntityManagerFactoryBuilderImpl.java:879)
	... 30 common frames omitted

michaelkain:
Hi Julien, sorrrryyy, I guess I found the problem on my side

michaelkain:
Hi Julien,
I solved my env problems and could finally test. The good side of this, now with my script I really start testing always from scratch. First I am quite a fan of the new design, for me really cool! On the other hand my testing results of the studycards are not as expected:

I can not create a study card as I am not able to select a dicom tag to finish a rule, the box remains always red. And as I can not create empty studycards I can not test the import etc.
Bildschirmfoto 2020-06-12 um 08 40 33
I have an error in the console: Http failure response for https://shanoir-ng-nginx/shanoir-ng/datasets/studycards: 503 Service Temporarily Unavailable

Furthermore there seems to be an overall bug with empty tables:
Bildschirmfoto 2020-06-12 um 08 42 03

michaelkain:
Failed to execute SQL script statement #16 of class path resource [scripts/import.sql]: INSERT INTO dataset (id, dataset_acquisition_id, origin_metadata_id, subject_id) VALUES (1, 1, 1, 1, 1), (2, 2, 2, 1, 1), (3, 3, 3, 1, 1), (4, NULL, 4, 1, 1); nested exception is java.sql.SQLException: Column count doesn't match value count at row 1

michaelkain:
Hi Julien, thx for the fix. I will start testing again and I will try to directly create Selenium tests. Could you please fix the merge issues? I would like as soon as my tests work accept your PR, that we have this big PR integrated. Thank you and a good week to you, Michael

Nunien:
Hi, I'm on it !

michaelkain:
Hi Julien, thx for the merge! Summary: I could create a study card and finish an import using this studycard, what is already a big step. Therefore I recommend, to discuss, to merge this monster PR now into develop. On the other hand: the study card had no effect, so it does not seem to work. I configured a rule on the SeriesDescription of the dicoms, see screenshot:
Bildschirmfoto 2020-06-15 um 11 04 26
 but the created dataset did not have Mr Dataset Nature set as configured:
Bildschirmfoto 2020-06-15 um 11 04 48
Mr Dataset Nature remains the same result as without this studycard.

Nunien:
@michaelkain Can you send me the zip you imported so I could reproduce the bug and fix it quickly ?

Nunien:
@michaelkain I've tested it with my own zip and it worked. I had to set "localizer" or "tse_vfl_WIP607" as it's the values I have in my zip for SeriesDescription (0008,103E).

michaelkain:
Hi Julien, I tested again with other data, and it works!!! Great! The StudyCards work, the problem is that during my tests, when creating a new subject and a new exam (what is the most common case), the GUI of the import has a bug and resets -> selected study card is lost and the first study card is selected instead, what lead to the problem. I modified and update quite a lot and tested and it worked nice, cool.

PR #308, opened 2020-05-26
Downloader

arthursw:
Add CLI tool and GUI to shanoir-uploader to download datasets from dataset ID, study ID and / or subject ID.


michaelkain:
Hi Arthur,
* Good job, as complicated environment!, but some open points
* Please change the license header to GPL v3.0 and remove the old header
in downloadDataset.sh and please remove the commented lines
* shanoir-uploader/pom_downloader.xml is not such a good style,
please move it to a sub-folder and work with sub-projects, that pom.xml
is always used
    * furthermore a duplicate code pom.xml is not good style neither
* MainWindow.java
    * please move your download Panel in a separate code class,
as it blows up the code of MainWindow
* ShanoirUploaderServiceClientNG.java
    * please remove your system.out.println and use su.log file instead
* pom.xml
    * please copy downloadDataset.sh as well into the target folder to simplify the delivery (lines 239,240)
* Hide your download dialog in case of profile isNotNG, as it will not work with old/soap (if not yet done)
* not related to your dev:
    * Could you please add shanoir-uploader into the „shanoir-ng-parent“ pom.xml, that we have one mvn
script that compiles everything? which will be great for further continuous integration, thank you!
* I will have to test further, as until now I only did the code review

arthursw:
Thanks for you review Michael : )
 
> shanoir-uploader/pom_downloader.xml is not such a good style,
> please move it to a sub-folder and work with sub-projects, that pom.xml
> is always used
>     furthermore a duplicate code pom.xml is not good style neither

Yes, that was my question in my email : how should I organize the pom.xml so that we can use it as a command line tool (and follow shanoir code style for continuous integration) ? I can't make it work and maybe you have a better idea to solve this.

 - Yes I definitely should have created a separate class for the downloader GUI ; I thought about it, but in the begnning the code was smaller, then I forgot about it.

I'll make those changes, tell me what you think for the pom.xml

michaelkain:
Hi Arthur, great, thank you for the changes! For the pom.xml, why do we need a separate pom.xml? And sorry, maybe I am not seeing the obvious.

michaelkain:
Normally when using pom.xml your Downloader is already part of the sh-up-all-deps.jar. You only will have to address it from your downloadDatasets.sh with this jar and the correct class name and it works?

michaelkain:
MAIN_CLASS=classes.org.shanoir.downloader.ShanoirDownloader
MAIN_JAR=shanoir-uploader-6.0.3-jar-with-dependencies.jar
$JAVA_HOME/bin/java -cp $MAIN_JAR $MAIN_CLASS
this should do it?

michaelkain:
could you please provide a .bat file too? most of the users have Windows, I think Julien or Yao could test it later

arthursw:
The problem is that I can't make it working on the command line:

`$JAVA_HOME/bin/java -cp classes.org.shanoir.downloader.ShanoirDownloader ./target/shanoir-downloader-6.0.3-jar-with-dependencies.jar`

I always get the following error:

`Erreur : impossible de trouver ou charger la classe principale classes.org.shanoir.downloader.ShanoirDownloader`


I tried many pom.xml configurations and many variations of the command line, but nothing has worked so far.

michaelkain:
remove the ".classes", I was wondering where it comes from as it makes no sense

arthursw:
Yes I surely planned to make a .bat file but I didn't create it since it's not working yet.

arthursw:
yes I did remove it, with no success

michaelkain:
why do you use -cp as a class path? java -jar "jar-name" "org....Downloader"

michaelkain:
$JAVA_HOME/bin/java  -jar shanoir-uploader-6.0.3-jar-with-dependencies.jar org.shanoir.downloader.ShanoirDownloader (this should do it)

michaelkain:
sorry, I correct: $JAVA_HOME/bin/java -cp shanoir-uploader-6.0.3-jar-with-dependencies.jar org.shanoir.downloader.ShanoirDownloader (this should do it) (assuming called in the target folder)

michaelkain:
the problem is maybe your package name of your class: org.shanoir.uploader.downloader.Downloader, mvn does maybe not integrate what is org.shanoir.downloader?

michaelkain:
so change it to org.shanoir.uploader.downloader.ShanoirDownloader and it should work

michaelkain:
Maven excludes maybe what is outside the package of the pom.xml, and this is why your class never comes into the .jar

michaelkain:
have you unzipped the jar and checked if your class is inside?

arthursw:
Ok I got it working!

michaelkain:
Great, see you this afternoon

michaelkain:
I will test further

michaelkain:
Hi Arthur, have you tried if the Import is still working after your changes? I tried to make an import to test afterwards the download, but when I click on "Importer" in the Imports tab I have a freeze and nothing happens anymore after: ShUp is frozen and I have to kill it. Could you please try this?

michaelkain:
by the way, I could see your nice interface for the download :-)

arthursw:
Hi Michael,

Nope I didn't try since I assumed I didn't change anything related to the import.
Do you have any idea why it crashes? 
I'll try.

arthursw:
I just tried, I didn't have any problem. Although I didn't really understand what I was doing... ^^

michaelkain:
Hi Arthur, I restarted my machine and now the ImportDialog opens, what is great. When I click on "Exporter", I have an exception:
2020-06-02 08:46:18 ERROR [org.shanoir.uploader.service.rest.HttpService] [52] shanoir-ng-nginxnull: unknown error
**java.net.UnknownHostException: shanoir-ng-nginxnull: unknown error**
	at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
Could you please delete your folder /.su_v6.0.3 and start from scratch?
Maybe there is a property value on your machine, that has not been commited into the profile properties? Could you please have a look? Thank you, Michael

arthursw:
Indeed, it seems I mistakenly removed the line:

`service.subjects.create=/shanoir-ng/studies/subjects`

in `profile.sh-ng-localhost/profile.properties`

Sorry about that, I just fixed it!

michaelkain:
Hi Arthur, my feedback:
- ShUp imports are now back working
- I could download a dataset on using the GUI, with datasetId
- I could download a dataset, nifti or dicom, via the command line, great
I will accept the PR and put my optimisation ideas into a GitHub issue.
Well done and see you later

arthursw:
Ok great!

PR #312, opened 2020-06-03
Continuous Integration with GitHub workflows: bug fix for ShUp

michaelkain:
Bug fix for mvn compilation of ShUp: 3 missing libraries, that are not in any public repository.
This is not the best style, but shanoir-1.0.0.jar and shanoir-ws-0.4.jar will soon be removed,
same later for dcm4che-custom-qr.jar. Solution to make it work for GitHub without adding the legacy libs to any repository.


michaelkain:
Works, now ShUp is auto-compiled to by GitHub workflows with mvn

PR #316, opened 2020-06-04
Remove servlet

arthursw:
Remove ImageViewerServlet.java and use a new API endpoint instead:

```
    @ApiOperation(value = "Get dicom image", notes = "Get dicom image", response = Void.class, tags={ "", })
    @ApiResponses(value = {
        @ApiResponse(code = 200, message = "get dicom image", response = Void.class),
        @ApiResponse(code = 400, message = "Invalid input / Bad Request", response = Void.class),
        @ApiResponse(code = 500, message = "unexpected error", response = Error.class) })
    @GetMapping(value = "/get_dicom/", produces = { "application/dicom" })
        @PreAuthorize("hasRole('ADMIN') or (hasAnyRole('EXPERT', 'USER') and @importSecurityService.hasRightOnOneStudy('CAN_IMPORT'))")
    ResponseEntity<ByteArrayResource> getDicomImage(@ApiParam(value = "path", required=true) @RequestParam(value = "path", required = true) String path) throws RestServiceException, IOException;

```

doing the same thing:

```

	/**
	 * This methods returns a dicom file
	 * 
	 * @param path
	 *            the dicom file path
	 * @throws ShanoirException
	 *             when something gets wrong during the import
	 * @throws IOException
	 *             when IO fails
	 * @throws RestServiceException
	 */
	@Override
	public ResponseEntity<ByteArrayResource> getDicomImage(@ApiParam(value = "path", required=true)  @RequestParam(value = "path", required = true) String path)
			throws RestServiceException, IOException {

		final File userImportDir = getUserImportDir();
		String pathInfo = userImportDir.getAbsolutePath() + File.separator + path;
		URL url = new URL("file:///" + pathInfo);
		final URLConnection uCon = url.openConnection();
		final InputStream is = uCon.getInputStream();

		ByteArrayOutputStream buffer = new ByteArrayOutputStream();
		int nRead;
		byte[] data = new byte[BUFFER_SIZE];
		while ((nRead = is.read(data, 0, data.length)) != -1) {
			buffer.write(data, 0, nRead);
		}
	 
		buffer.flush();
		byte[] byteArray = buffer.toByteArray();
		
		ByteArrayResource resource = new ByteArrayResource(byteArray);

		return ResponseEntity.ok()
				.contentType(MediaType.parseMediaType("application/dicom"))
				.contentLength(uCon.getContentLength())
				.body(resource);
	}


```


michaelkain:
Hi, could you check the errors?

michaelkain:
Hi Arthur, thank you a lot! The code looks great, I finished the code review. I will checkout your PR and do some preview testing.

arthursw:
Thanks Michael :)

michaelkain:
Hi Arthur, my results on the other hand are quite bad. Have you really tested?
1) when I use the sample.zip (40mb) I have no preview at all with import from dicom (error)
(I choose the file, see the tree, click on the series and nothing happens)
2) when I use the big zip archive 500mb, I have a direct error: file name is red and I can not continue

arthursw:
Hi,
 
I have tested and it works fine for me with the 40mb sample.zip
I will retest, but could you give me an error message, response error, or log?

I have the same error with the big zip, but this error is not related to my changes, I talked with Jean Come and he told me it is supposed to work but the .zip he sent us might be broken (I think Jean Come also has this problem with this zip).

jcomedouteau:
For the 500mb archive, the archive in itself may be not correct.

arthursw:
@Jean-Come : Can we be sure?

arthursw:
Or have another zip for our tests?

michaelkain:
okay, in any case it would be great if we would share a working big zip with all developers instead one that is not working. My problem with the sample.zip is, that I see no preview at all, so I have no error to share unfortunately... we agree, when I click on the series name or the "brain" icon I should see the preview below in the dicom tag info box, right? but I have nothing, by the way I am testing with Firefox

michaelkain:
unfortunately I do not have a big one with human data

michaelkain:
same behaviour in Chrome, no preview at all, by the way have you cleared your browser cache with your tests?

arthursw:
The preview is only available from the bruker import; I'll check the code to investigate this further.

michaelkain:
most of the import with Neurinfo will be human

arthursw:
Do you mean that you want the preview to work with regular import as well?  This means changing the implemented behavior, but it should be doable, I'll check.

jcomedouteau:
This one is not that big, but still big, and it works while importing
https://filesender.renater.fr/?s=download&token=c6f481c0-73db-434f-bd24-83ba181ef140

arthursw:
Thanks Jean-Come!

Nunien:
@arthursw @michaelkain 
1. Why have we decided to dump the 'in browser' unzip ?
2. During the import, when visualizing one dataset, I don't understand why there are 100 network exchanges with the backend (the old servlet did the same) while there is only one when visualizing 
3. That exchange should not prevent the metadata to be displayed while it's loading

PR #322, opened 2020-06-09
Correct selenium tests

jcomedouteau:
- Allow preclinical subject to be deleted
- Use select correctly during imports
- Create a preclinical subject during import
- Create a preclinical examination during import
- Add possibility not to click on "create new" button
- Clean fields when adding field value


michaelkain:
Hi, I start to review this pull request. I would like to create already Selenium tests for the testing of the StudyCards PR, that is why I would like to have the latest version of tests on develop, that I can extend from this version.

PR #327, opened 2020-06-11
Filter preclinical examinations in back not in front using pageable

jcomedouteau:
- Use adapted API to return only non-preclinical examinations
- Remove manual filter in front in contradiction with Pageable entity


jcomedouteau:
Not merged, already solved

PR #328, opened 2020-06-11
Migration fixes

a-ba:
a few fixes needed for running the migration on the qualif instance

please review carefully (especially the removal of the subqueries)

### DIFF
--- a/docker-compose/migration_scripts/Neurinfo/1_users_migration.py
+++ b/docker-compose/migration_scripts/Neurinfo/1_users_migration.py
@@ -4,7 +4,12 @@
 import os
 import pymysql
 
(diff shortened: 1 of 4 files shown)


michaelkain:
Hi Anthony,
thx for your work on the mig scripts. I reviewed your pull request and I will accept it.
I have to admit, that I did not dig deeper and retest the sub-query change, as this part of the migration scripts has been developed by Mathieu and changed by Arnauld, some time ago. 
Arnauld agreed with some people of the team to move down the four tables in the model, as data have progressed since dev sh-old.
When qualif is deployed, I will check if the values correspond and discuss this with Elise. Everything else is too time consuming and will not bring us much further too. Cheers, Michael

PR #331, opened 2020-06-12
Equipment compatibilty check

Nunien:
During import in old, you could have an annoying error message saying prisma is not compatible with Prisma.

In this commit the checks are made case insensitive and there is also a trim to remove spaces at start and end of strings.


jcomedouteau:
I think you want to merge this to develop no ?

Nunien:
@jcomedouteau It's not urgent at all. It's a very minor change so it can be merged easily any moment later.

Nunien:
@jcomedouteau I mean, after the studycard merge

PR #332, opened 2020-06-12
Sc migration - NOT TESTED

Nunien:


michaelkain:
Hi Julien, since yesterday we have the StudyCard PR accepted with all its commits. I am not sure, if it is correct, that it remains so many commits on this PR. Could you please reopen a PR with only the migration script changes, that I can start testing? I have fear accepting this PR will double apply the commits of StudyCards and create problems. If a GitHub expert tells me otherwise, he can apply the PR ;-). This should avoid the merging conflicts too, right? With kind regards, Michael

michaelkain:
I found your changes on the mig scripts, I will start testhing them already

michaelkain:
Hi Julien, I think the most easy is: delete your PR, and I create a new one from your changes on the mig scripts, as I have to modify them. Cheers, Michael

PR #337, opened 2020-06-17
add missing migrations for the users db

a-ba:


michaelkain:
Hi Anthony, again it looks completely logic. I have not run it on my machine, but I will compare the users during my "user" tests. Thank you and have a nice weekend, Michael

PR #341, opened 2020-06-18
Massive download - Split if download > 5Gb

jcomedouteau:
Massive download big files.
When files are > 5 gb, split them in tasks / datasets


jcomedouteau:
I'm correcting this.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/exporter/service/BIDSServiceImpl.java
```diff
@@ -424,7 +424,8 @@ private void createDatasetBidsFiles(final Dataset dataset, final File workDir, f
 				// TODO: center_id / comment / weigth / other examination things ?
 				Files.write(Paths.get(scansTsvFile.getAbsolutePath()), buffer.toString().getBytes(), StandardOpenOption.APPEND);
 
-			} catch (IOException exception) {
+			} catch (Exception exception) {
+				exception.printStackTrace();
```
jcomedouteau, line 428: Do a proper logging.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -398,7 +406,27 @@ else if (dataset instanceof EegDataset) {
...
+			// Publish events to download files on by one
+			
+			for (ShanoirEvent event : events) {
+				// publish all events for every datasets
+				eventService.publishEvent(event);
+			}
+
+			// Here delete zip file only
+			FileUtils.deleteQuietly(zipFile);
+			FileUtils.deleteQuietly(tmpFile);
```
jcomedouteau, line 427: Do not delete the file and use it intelligently to be able to re-download the dataset (dataset file by dataset file).
=> Less calculating necessary, and keep it faster.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -461,6 +503,39 @@ public void massiveDownload(String format, List<Dataset> datasets, HttpServletRe
...
+			if (!datasetFile.exists()) {
+				return new ResponseEntity<>(HttpStatus.UNPROCESSABLE_ENTITY);
+			}
+
+			File zipFile = new File(datasetFile.getAbsolutePath() + ZIP);
+			zipFile.createNewFile();
+
+			Utils.zip(datasetFile.getAbsolutePath(), zipFile.getAbsolutePath());
+
+			byte[] data = Files.readAllBytes(zipFile.toPath());
```
jcomedouteau, line 525: This is not good for me => We load all in memory

jcomedouteau:
This pull request is out of touch for me:
- There is a flaw in the way we treat the data (load all data in memory)
- It is not adapted anymore to the way we download data
- Generating  links to download datasets one by one is maybe not the best way to manage this. (We could imagine an iterative zip growing as we add datasets then stop before 5Gb, then restart a new "zip" and then generate links for those.)
This also could be linked to the development of the GME to track data size.

I'm closing it for the moment.

Issue #340, opened 2020-06-17
qualif issues with dataset_acquisition/acquisition_equipement


a-ba:
on the qualif server:

- the page https://shanoir-qualif.irisa.fr/shanoir-ng/dataset-acquisition/list freezes because the XHR to https://shanoir-qualif.irisa.fr/shanoir-ng/datasets/datasetacquisition never replies (in background it generates a *lot* of traffic with the mysql database)

- unable to start an import : the form fails because https://shanoir-qualif.irisa.fr/shanoir-ng/studies/acquisitionequipments returns 404

Nunien:
* The timetout will hopefully be fixed with #346 
* I don't knwo why, but the problem with acquisitionequipments is that there are no equipments in the db. I created one and it works.

michaelkain:
Fixed

PR #346, opened 2020-06-22
Ds acq timeout

Nunien:
#340 : 
There was a recursive json serialization, I had to add @JsonIgnore to mrProtocol (and mrDataset) in DiffusionGradient.java


shanoir-ng-front/src/app/import/clinical-context/clinical-context.component.ts
```diff
@@ -263,6 +263,8 @@ export class ClinicalContextComponent implements OnDestroy {
             this.niftiConverter = this.studycard.niftiConverter;
             this.onSelectNifti();
         }
+        this.importDataService.contextBackup.studyCard = this.studycard;
+        this.onContextChange();
```
jcomedouteau: You may have to apply this to eeg-clinial-context and preclinical import too ?

PR #414, opened 2020-07-31
Sc migration script coils

Nunien:
To test : try to import with a studycard that has coils in the rules. Try with admin, admin to the study and common user as the message is not the same in clinical context.


michaelkain:
Hi Julien, could you solve the merge conflict? maybe with Jean-Come? Jean-Come solved the bug, that all previously selected items (study, subject, etc.) within the clinical context remain selected, what was important to fix, maybe there is a link. I will test in between the db mig sql script. Cheers, Michael

michaelkain:
Hi Julien, your changes on the studycards in the migration script do not throw any error. Let's discuss in the daily, what exactly I should test, when you fixed the merge problem. Thank you and best, Michael

Nunien:
The conflict is resolved.
I believe that my version was fixing the same bug as @jcomedouteau fixed, along with other fixes and improvements so I overrode his.

michaelkain:
Great! I will check it out locally and do some testing.

michaelkain (approved):

michaelkain:
Hi,
- migration script runs without errors
- imported works fine, tested on my machine
- study card created and edited with an action for receiving coil: worked in GUI
--> problem I can not verify it with the GUI as dataset acquisitions are empty in the GUI
--> I did not re-run the mig scripts to test them again, as very time demanding, so we will verify on the qualif

PR #420, opened 2020-08-03
Shanoir issue#411 - dcm2niix

jcomedouteau:


michaelkain:
Hi Jean-Come, I think this file is not used anymore: shanoir-ng-import/src/main/docker/files/dk/container_init.
Same for this one: shanoir-ng-import/src/main/docker/files/opt/cfg/shanoirng_import.sql.
This one should be the right one:  shanoir-ng-import/src/main/resources/scripts/import.sql
I am not sure if Anthony for the qualif deploy does an "init" with import.sql, before the migration.
In any case, the mig scripts clean all converters and migrate, to discuss.
Thank you,
Michael

jcomedouteau:
It is one of the points I had questions about.
Anthony told me that the init script was applied first, then the migration script on the qualif. The fact is that some "old" converters are no more usable in shanoir-NG. => Should we still migrate them ? Or transfer them to an equivalent ? (mcverter_2.0.7 to mcverter_2.1.0).
Whatever we do, it has to be "intelligent" so that the old data can still be converted.

PR #429, opened 2020-08-05
Shanoir issue#409 Dataset Acq protocol display [Do not merge]

jcomedouteau:
@Nunien I think using DatasetAcquisitionDTO, you did not propagate all the protocol part to the front. I added something in the mapper, but it may not be ideal.
There is another problem displaying the protocol: I think the protocol is not transferred in an adapted angular object.


Nunien:
@jcomedouteau You may need to update src/app/dataset-acquisitions/shared/dataset-acquisition.dto.ts and rename mrProtocol, petProtocol and ctProtocol (into MrDatasetAcquisitionDTO, PetDatasetAcquisitionDTO and CtDatasetAcquisitionDTO) to "protocol".

Nunien (approved):

PR #442, opened 2020-08-20
Keycloak security optimisations

michaelkain:
- sslRequired all in Kc
- BFD enabled by default
- ms users set from confidential to bearer-only, as no need to do otherwise
- empty redirectUri in ms users fixed as security issue


michaelkain:
Hi Jean-Come, please test this PR thoroughly, as the impact is pretty high.
I tested: create new user with the web GUI and login with it, check login date updated, users created during startup of ms users, edit existing user, login with.

jcomedouteau:
I tried to create, edit, connect as users, change expiration date and test connection failure, call some API of every MS to get a token, everything seems fine.

jcomedouteau (approved):

jcomedouteau:
@michaelkain I think you can merge this one.

Issue #444, opened 2020-08-20
Small error messages within empty tables, with new users having no studies assigned


michaelkain:
- create a new user with admin and the web GUI, with user type User (as most of our users)
- change PW and login with this user
- user has access to no study, so his tables are all empty, BUT
there are a lot of errors displayed, instead of empty tables:
- Manage examinations
- Manage dataset acquistions
- Datasets
- StudyCards
- Preclinical all entities
- same for Solr search

It can take some time to add an user to a study, when the account is accepted.
All these errors give the impression, that something is not well implemented.

Nunien:
@michaelkain 
* I couldn't reproduce any problem with the preclinical lists
* The problem with Solr is inherent to the Solr implementation in Shanoir I think we should fix that independently.

jcomedouteau:
=> Should I create an issue for the solr problem ?

Issue #431, opened 2020-08-10
Qualif: with ShUp: filterStudyCards exception


michaelkain:
2020-08-10 15:00:08.419 ERROR 1 --- [http-nio-9904-exec-9] o.s.n.s.c.GlobalExceptionHandler         : Unexpected error in the rest service. 

java.lang.NullPointerException: null
	at org.shanoir.ng.dataset.security.DatasetSecurityService.filterStudyCardList(DatasetSecurityService.java:480)
	at sun.reflect.GeneratedMethodAccessor208.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.expression.spel.support.ReflectiveMethodExecutor.execute(ReflectiveMethodExecutor.java:113)
	at org.springframework.expression.spel.ast.MethodReference.getValueInternal(MethodReference.java:129)
	at org.springframework.expression.spel.ast.MethodReference.access$000(MethodReference.java:49)
	at org.springframework.expression.spel.ast.MethodReference$MethodValueRef.getValue(MethodReference.java:347)
	at org.springframework.expression.spel.ast.CompoundExpression.getValueInternal(CompoundExpression.java:88)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:170)
	at org.springframework.expression.spel.ast.OpAnd.getBooleanValue(OpAnd.java:56)
	at org.springframework.expression.spel.ast.OpAnd.getValueInternal(OpAnd.java:51)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:170)
	at org.springframework.expression.spel.ast.OpOr.getBooleanValue(OpOr.java:55)
	at org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:50)
	at org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:36)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getTypedValue(SpelNodeImpl.java:131)
	at org.springframework.expression.spel.standard.SpelExpression.getValue(SpelExpression.java:299)
	at org.springframework.security.access.expression.ExpressionUtils.evaluateAsBoolean(ExpressionUtils.java:26)
[...]

PR #446, opened 2020-08-24
fix several empty list bug messages

Nunien:
see #444 & #431


jcomedouteau (approved):
Ok for me. I create an issue for solr results

PR #462, opened 2020-09-01
Bids import

jcomedouteau:
Import from BIDS - adapt to study card

- Reset menus
- Change description
- Configure study card (equipement, center, nifti converter)
- Use rabbitMQ as much as possible


michaelkain:
Hi Jean-Come, as I am not fully aware of all details of the BIDS import. Do we already use the JavaScript BIDS Validator in the frontend to verify if the uploaded data are valid BIDS data? https://github.com/bids-standard/bids-validator

michaelkain (changes requested):
Hi Jean-Come, could you maybe externalise the logic in the method importAsBIDS into some kind of service component? The ImportApiController is becoming really big.

jcomedouteau:
That's a good idea to move out the bids import logic.
We don't use the validator for the moment. I've had a look at it, but there was no validation for EEG bids format at the time so I put it aside. But we could now integrate it.

michaelkain:
Hi Jean-Come, if I understand it correctly, the current BIDS import uses the **/sourcedata** folder with the dicom files and a shanoir-import.json as sidecar in the sourcedata folder to import the dicoms for multiple patients with the same studycard. What is fine, but it means at the moment, we do not have an import of the nifti files in the **/rawdata** folder, what is the most common use case of BIDS. I just try to understand and remember back, where we are.

michaelkain:
Great for the validator. I think it will look very serious if we display clearly in the front, that we use the official validator :-).

jcomedouteau:
> Hi Jean-Come, if I understand it correctly, the current BIDS import uses the **/sourcedata** folder with the dicom files and a shanoir-import.json as sidecar in the sourcedata folder to import the dicoms for multiple patients with the same studycard. What is fine, but it means at the moment, we do not have an import of the nifti files in the **/rawdata** folder, what is the most common use case of BIDS. I just try to understand and remember back, where we are.

That's exactly what we are doing now. But if I understand well, we'll need this nifti import as soon as possible (?)

michaelkain:
No hurry, the lady wrote they have niftis in /rawdata and I think corresponding dicoms in /sourcedata. In any case I guess they will like to keep their nifti files generated by them, so we should think of this alternative: A) having a BIDS import with only /rawdata and B) having a BIDS import with both data /rawdata + /sourcedata. I guess we should have a meeting on this?

jcomedouteau:
We have to see how we integrate it with processed dataset too. Maybe we can talk about it during the daily today then plan a meeting to schedule / organize the dev. (I don't want it to be a huge pile of code from only one dev, I think we can cut it intelligently.)

jcomedouteau:
I updated this pull request to move the code to a new API and prepare the further developments.

michaelkain (commented):
I have reviewed the code, and thank you for the change with a separate controller!
Great, the change request is implemented.
I will have to test the PR locally with some imports before approving it finally.

jcomedouteau:
I'll need this PR to start the rest of BIDS import I think.

PR #464, opened 2020-09-02
fix the location of the old nifti files

a-ba:


Nunien:
test : 
* do an import
* check if nifti are on disk
* check if dicom are on disk

PR #477, opened 2020-09-07
Shanoir issue#476 Dicomifier as nifti converter

jcomedouteau:
Add dicomifier as nifti converter
Rest call to convert
Manage nifti files differently


jcomedouteau:
@Nunien Can you have a look at this one ? Thanks :)

Nunien:
I don't know if import.sql is called on deployment but if not we would have to execute manually this into the import ms db : 

INSERT IGNORE INTO nifticonverter (id, name, nifti_converter_type, is_active) VALUES (7, 'dicomifier',5,1);

Nunien (approved):

PR #488, opened 2020-09-11
Shanoir issue#487 - download nifti as simple user

jcomedouteau:
The method used by Yao returned a list of datasets and not a boolean => Change the method to do the check properly.


michaelkain (approved):
This PR fixes the blocking issue for the qualif. We will have to optimise it in a way, that a call to the API allows download of datasetIDs only, where the user has the right on ALL datasets, I guess, to be discussed.

PR #498, opened 2020-09-15
Shanoir issue#487 download datasets

jcomedouteau:


michaelkain (changes requested):
Hi Jean-Come, I have fear that the download of an entire study could crash the server, e.g. we have a study that contains 3.000 patients. What do you think? In the code your limit is only for multi-downloads. Cheers, Michael

jcomedouteau:
That's totally true !

michaelkain:
I am on it.

michaelkain:
Great job! I have tested on my machine. Single file download, multiple, 50 or more, everything worked fine. I will accept the PR for the 22.09., great. Only one point to optimise: the case of downloading one empty dataset could be optimised, as it returns a zip file, that can not be opened.

PR #509, opened 2020-09-21
Shanoir issue#508 examination extra-data + edition

jcomedouteau:


jcomedouteau:
@michaelkain I cannot test a lot localy as I don't have intrsument based assessment, but it shoud be OK

Nunien:
@jcomedouteau Can you explain me what it does and what I could test ?

jcomedouteau:
1) You can now edit an examination if you have the right to import on the study.
2) You can add "extra file" to an examination and manage them (download them, delete them, add new ones)
3) If you have instruments on an examination, it is now displayed.

For the third point, it is a feature that existed in old, we cannot create them anymore, but we only display them. => If you want to display one, you have to manually add one in DB (instrument, instrument_based_assessment). I can give you a SQL if you want.

Nunien:
Sorry, there are a few bugs : 
* As an expert and having every right on the study, I can not edit the examination (403 "Accès refusé" on examination PUT)
* When I download a file, its downloaded twice ! (therefore the browser asks me if I allow to download multiple files from this website)

Also here is a question : How can I create instruments, I have none in my database ?

Nunien:
Also I have commited a fix in your pull request (I have tested the fix). It prevents the save button to remain disabled when adding/removing a file without editing anything else. A better but more complicated solution would maybe to have made a ControlValueAccessor for the attached files field so the form would have managed the clean/dirty thing automatically.

Maybe do a git pull of your pull request before comiting if you don't want to add a merge commit.

jcomedouteau:
Thank's, I'm on it.

jcomedouteau:
To create an instrument, you have to do it directly in DB.
=> docker exec -it database /bin/bash
=> mysql -u root -p datasets
And then create an instrument, and create an instrument_baszed_assesment linked to an examination

jcomedouteau:
@Nunien  I corrected the rights problem, though I could not reproduce the "double download" problem.. Could you show me ? Is it everytime ?

Nunien:
@jcomedouteau I think yes. But maybe you have to put two or more files first. I'm sorry I can't test it for the moment.

Nunien (approved):

PR #522, opened 2020-09-23
Downloader resources in jar

arthursw:
Load the downloader config from resources in the .jar instead of loading from `/home/userName/.suShanoirVersion/`


michaelkain:
Hi Arthur, sorry for being picky, but could you please make a PR dedicated to ShUp and the issue? This one contains the the downloadStats part too, what I would like to test and add separately. Thank you, Michael

arthursw:
I'm sorry I must have messed up my develop branch. I'll fix that.

michaelkain (changes requested):
Hi Arthur, sorry for this feedback, but this will still not work. The user enters a host on the command line, that should be used, and called. So you will have to use this host URL in two places:
1) for Keycloak auth, to address the correct server: host-url-by-user + "/auth" for your method
keycloakAuthentification(String host, String username, String password) == auth-server-url
2) as property for NGClient (that is what you already do, as far as I understand :-)).

arthursw:
Ok I'll fix that!

arthursw:
Should I prefix the host with "https://" when `port` is 433 and with "http://" when `port` is 22?

arthursw:
I guess yes, so I'll do that for now.

michaelkain (approved):
Thank you, looks fine. I will merge and then test on my machine, as this will be part of ShUp v7.0.1.

Issue #540, opened 2020-09-29
Prod: import from pacs: use exam date to sort request answer after Pacs query


michaelkain:
I meant when importing data from PACS : the result of the request sorts the examination without the dates. If they are several examinations the date helps to identify which one is which.

michaelkain:
Result of analysis:
1) the Dicom study date (-> Exam in Shanoir) is not correctly interpreted and displayed in the GUI
2) the Dicom study date is not used to list the Dicom studies by date
@ebannier2018: would you prefer to order chronologically (2009,12,19...) or the newest exam on top?
![Bildschirmfoto 2020-10-01 um 10 29 45](https://user-images.githubusercontent.com/4560288/94786657-889d7200-03d1-11eb-8657-6b26493939b7.png)
![Bildschirmfoto 2020-10-01 um 10 29 20](https://user-images.githubusercontent.com/4560288/94786683-90f5ad00-03d1-11eb-9183-6a46b1b65a19.png)
![Bildschirmfoto 2020-10-01 um 10 28 54](https://user-images.githubusercontent.com/4560288/94786686-918e4380-03d1-11eb-9b17-d13677187fdc.png)

ebannier2018:
Newest exam on top would be better but the explicit date at the examination or dataset level ideal as used to be in shanoir-old. You can try with the following requests :

Import from PACS : Patient Name : PHANTOM QUALITE 20200925 in Shanoir-NG
returns examination with PHANTOM or QUALITE or 20200925 but we want "PHANTOM QUALITE 20200925" as implemented in Shanoir old : 
For comparison : Import from PACS : Patient Name : PHANTOM QUALITE 20200925 in Shanoir-old 

And the date filter does not help (same answer as without date)


Import from PACS : Patient Name : PHANTOM QUALITE 20200925 + date 20200925 in Shanoir-NG

<img width="1680" alt="ShanoirNG_ImportFromPACS_PHANTOM QUALITE 20200925" src="https://user-images.githubusercontent.com/36895591/94793734-3d885c80-03db-11eb-8df3-c255e4b953ed.png">

<img width="1680" alt="ShanoirOld_ImportFromPACS_PHANTOM QUALITE 20200925" src="https://user-images.githubusercontent.com/36895591/94793726-395c3f00-03db-11eb-95aa-7f6e341fa81e.png">

michaelkain:
[...]

PR #559, opened 2020-10-05
Fix for GitHub issues: #540 (Pacs import)+#561: import data issue

michaelkain:
Prod: import from pacs: use exam date to sort request answer after Pacs query #540
- PatientRootQuery now considers dicom study date and study description as additional criteria
- Dicom study level ordered by study date (-> exam in Shanoir) old to newest
- Dicom tree: displays now the date of the Dicom study (-> exam in Shanoir)
- Wrong date display of Dicom study fixed

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/model/mr/MrSequenceVariant.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/model/mr/MrSequenceVariant.java
@@ -20,7 +20,6 @@ package org.shanoir.ng.datasetacquisition.model.mr;
  * @author atouboul
  *
  */
-
 public enum MrSequenceVariant {
 
 	// segmented k-space
@@ -46,6 +45,8 @@ public enum MrSequenceVariant {
 	
 	// no sequence variant
 	NONE(8);
+	
+	private static final String TOF = "TOF";
 	
 	private int id;
 
@@ -87,6 +88,9 @@ public enum MrSequenceVariant {
 	public static MrSequenceVariant getIdByType(final String type) {
 		if (type == null) {
 			return null;
+		}
+		if (TOF.equals(type)) { // see GitHub issue #561
+			return MrSequenceVariant.SS;
 		}
 		return MrSequenceVariant.valueOf(type);
 	}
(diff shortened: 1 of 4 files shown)


michaelkain:
@Nunien Hi Julien, could you just do a code review, to see if I have not "damaged" something in the front. My tests all worked well. For the final Pacs acceptance test, I will ask Elise to test on qualif herself.

michaelkain:
Hi Julien, could you review this please, that we can bring it to qualif asap? Thank you, Michael

michaelkain:
@Nunien Hi, as I did not have answer from you, and as this fixes are quite required I merged them today. With kind regards, Michael

ebannier2018:
Hi, I tried some PACS imports today on qualif. The request Patient Name : PHANTOM QUALITE 20200925 gives me more than the one expected answer. The patient name "PHANTOM QUALITE 20200925" should be considered as a whole and I should only get one examination as an answer. Trying to filter with study date I could get the one examination only which was not working on prod and this is good, but I expect only one answer on the patient name request anyway, since the patient name is unique. It should NOT return the result of PHANTOM or QUALITE or 20200925 as seems to be the case. No wildcard should be possible there either as we agreed. 

Capture d’écran 2020-10-26 à 10 08 23

Also : 
- the multiple examinations are ordered from oldest to newest and we expect the contrary (newest first). 
- they should not show as examinations from the same subject, since the patient names are different. 

Last, I do not agree with the following description : 
Please fill at least one of the five fields below to query the PACS. Blank is not allowed.
If Patient Name, Patient ID or Patient Birth Date is filled, a PatientRoot query is send to the PACS.
You can fill Study Description and Study Date in combination with the three patient values above to affine your search.
If only Study Description or/and Study Date is filled a StudyRoot query is send to the PACS.

To  me Patient Name is mandatory and should be considered as a whole. Patient ID and Patient Birth I do not see the use. Do some users do? 
Study Date can be used to affine the search indeed. Study Description also but not sure is it useful. 
So StudyRoot query should be sent is only Study Description or Study Date are filled since Patient Name is mandatory. 

We can discuss this over the phone if this is not clear. 

Cheers, 
Elise

ebannier2018:
Any news on this to consider the patient name as a whole (e.g. "PHANTOM QUALITE" is "PHANTOM"and "QUALITE" not "PHANTOM" or "QUALITE") ?

ebannier2018:
Short reminder on this issue, if you could take a look

PR #562, opened 2020-10-05
Shanoir issue#531 Solr search logic

jcomedouteau:
Shanoir-issue#531 Correct solr search logic

- Better wildcard management
- Use _str fields for better results also


michaelkain (approved):
Looks completely fine and logic. I have not tested on my machine, but we will test on qualif with the real data.

PR #618, opened 2020-10-23
Shanoir issue#617 Nifti in Dicom download DO NOT MERGE

jcomedouteau:
When consulting a dataset, we load nifti for preview. As we changed dataset download name => we were loading it in the folder that is used to download DICOM too. => So we have both.

I changed:
- file name we use (with date)
- Exception / logging management
- Do not load nifti for preview everytime but only on demand.


jcomedouteau:
Wrong branch

PR #623, opened 2020-10-26
Shanoir issue#611 receive email for import

jcomedouteau:
- Use user_id instead of study_user id
- Use newImport and not anonymization
- Add tests to check all this..


michaelkain (approved):

ebannier2018:
Also seems I cannot configure my own account to receive the import email...could you check that ?
Example screenshot on MS-TRACTS study 
ReceiveImport_Responsible

PR #637, opened 2020-11-03
Shanoir issue#633 Use HttpResponseServlet instead of ByteArray to download

jcomedouteau:
Shanoir-issue#633 

- Do not load in memory
- Delete used files
- Remove useless logs
- Use user directory everywhere
- re-activate study protocol file


jcomedouteau:
To be noted, for extradata and bids direct download, the download is done twice -> This is linked to a bug in tree-node component that'll be correcte by @Nunien 
In tree-node.component.html => buttonClick.emit() is called twice in <span> and <i>

shanoir-ng-datasets/src/main/java/org/shanoir/ng/exporter/controller/BidsApiController.java
```diff
@@ -49,39 +52,52 @@
...
 		// Get file, zip it and download it
     	File fileToBeZipped = new File(filePath);
     	if (!fileToBeZipped.exists()) {
-			return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+			response.sendError(HttpStatus.NO_CONTENT.value());
+			return;
     	}
     	
     	// Copy / zip it (and by the way filter only folder that we are interested in)
+    	String userDir = getUserDir(System.getProperty(JAVA_IO_TMPDIR)).getAbsolutePath();
```
michaelkain, line 73: Hi Jean-Come, it could be a little bit risky to use IO_TMPDIR here, as this resolves depending on the OS. As we are always in the same container using docker-compose this is not an issue, but it could happen, that his variable does not point to the /tmp, what is the named volume proposed by docker-compose.

michaelkain (approved):
Hi, only one minor comment. So I approve to be tested tomorrow too on qualif.

michaelkain:
Hi, could you please fix the conflicts? Sorry.

jcomedouteau:
No problem, conflicts merged

PR #639, opened 2020-11-03
Shanoir issue#605 [DO NOT MERGE] Bids update - Session level management

jcomedouteau:
Not completed yet:

- Add session level based on:
    -> If multiple exams for a subject
    -> If comments are logic between subjets (To apply strictly ?)
THEN we base our session level on comment.

- Add button to update bids tree (update session level accordingly)


jcomedouteau:
- TODO:
Manage deletion (dataset, examination)
Better manager import

=> In general matters, just update the bids folder entirely when an update strikes ? (what about "local" files ? => extra-data ?)

jcomedouteau:
Will be included in another PR

PR #652, opened 2020-11-06
Shanoir issue#628 dissallow subject name update

jcomedouteau:
Cannot update subject name anymore (front + back)
Remove BIDS update as it was based on name change only


michaelkain (approved):
very good! thx for having added the very useful unit test!

Issue #630, opened 2020-10-29
Prod: dicom import: error: phantom: use exam date instead of birth date for subject creation


michaelkain:
Something weird happening with the Birth Date, also if we select Phantom instead : 01012020, while we expect the acquisition date, does this explain the error, with the import (see the mail from EB)

ebannier2018:
I was in a rush when I sent you the email. We usually set the phantom to 30 years old and birth date is entered accordingly but this is not the issue. Somehow we could not perform the import for this dataset and we do not know why.

michaelkain:
Hi Julien, could you please start to work on that? Please check in the logs, where the problem comes from, I guess there are some dicom data missing/formatted differently?

Nunien:
Ok but I want to discuss it at the dsu as the problem is not clear to me for the moment

Nunien:
Ok so if I understand, a phantom bithdate should be the acquisition date.

ebannier2018:
No, that is not the problem. 

-------------------------------------------------- 
MRI Physicist, PhD 
Plate-forme NeurInfo, https://www.neurinfo.org/ 
CHU Rennes, https://www.chu-rennes.fr/ 
Unité Empenn (ex-VISAGES) U1228 INSERM-INRIA, IRISA UMR CNRS 6074, https://team.inria.fr/empenn/ 
Membre du Réseau d'Entraide Multicentrique en IRM, https://remi.network/fr/ 

+33 2 99 28 99 06 (Neurinfo) 
+33 2 99 84 22 13 (Empenn) - Fridays

ebannier2018:
And I do not know where the problem comes from either...do the logs give some kind of info? 

-------------------------------------------------- 
MRI Physicist, PhD 
Plate-forme NeurInfo, https://www.neurinfo.org/ 
CHU Rennes, https://www.chu-rennes.fr/ 
Unité Empenn (ex-VISAGES) U1228 INSERM-INRIA, IRISA UMR CNRS 6074, https://team.inria.fr/empenn/ 
Membre du Réseau d'Entraide Multicentrique en IRM, https://remi.network/fr/ 

+33 2 99 28 99 06 (Neurinfo) 
+33 2 99 84 22 13 (Empenn) - Fridays

Nunien:
@ebannier2018 Can you tell me the date + aproximate hour of that failed import ? Also, if you can try the import yourself and if you have a moment, it will be great to have a call so you can show me the bug.

[...]

PR #653, opened 2020-11-06
fix error with subject creation see #630

Nunien:
I think the dicom archive that Elise was importing contains wrong string values for the phantoms sex. It causes the backend to bug. Now that we check the value the bug should disappear.

Not easy to test.


michaelkain:
Hi Julien, I think this is quite easy to test. 1) you can ask Elise for the test data (phantom data, so no legal issues) or 2) you can use the same query params of Elise and test with qualif to see if your fix works (import from pacs). I can not see how your fix in the front, that does not set any sex, if no sex is present in the dicoms could solve the problem. With kind regards, Michael

Nunien:
TODO for me : check in the back if the patient is build with the correct sex

jcomedouteau (dismissed):

michaelkain (approved):

PR #662, opened 2020-11-13
Shanoir-issue#660 Propagate disable user to keycloak

jcomedouteau:
- In scheduledTasks, add expiration user propagation to KC-
- Add tests

### DIFF
--- a/shanoir-ng-users/src/main/java/org/shanoir/ng/user/repository/UserRepository.java
+++ b/shanoir-ng-users/src/main/java/org/shanoir/ng/user/repository/UserRepository.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 
 import org.shanoir.ng.user.model.User;
 import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.redis.core.types.Expiration;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.stereotype.Repository;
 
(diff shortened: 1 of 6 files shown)


shanoir-ng-users/src/main/java/org/shanoir/ng/user/service/UserService.java
```diff
@@ -111,6 +111,14 @@
 	@PreAuthorize("hasRole('ADMIN')")
 	List<User> getUsersToReceiveSecondExpirationNotification();
 
+	/**
+	 * Find users that are now expired.
+	 * 
+	 * @return a list of users.
+	 */
+	@PreAuthorize("hasRole('ADMIN')")
+	List<User> getExpiredUsers();
```
michaelkain, line 120: Hi Jean-Come, thank you for your PR! What happens if a user is already expired, so its expirationDate is yesterday, but the code runs today? Will it be found too? Maybe we have a downtime for one or two days, would the code disable as well users, that expired yesterday?

michaelkain (commented):
Hi Jean-Come, I think an administrator should have the possibility to deactivate an user manually and directly using sh-ng when setting the expiration date to yesterday or before. This could be directly communicated to Keycloak too, to disable the user. Today the admin can only delete an existing user to block his access. I think we should discuss this. By the way, is a user deleted in sh-ng mysql database as well deleted from Keycloak?

shanoir-ng-users/src/main/java/org/shanoir/ng/user/service/UserService.java
jcomedouteau: For the expiration date, we have no real way of tracking the "disabling" of a user. Maybe we take can a week of margin (I hope we'll never an interruption of more that 7 days !).
The "problem" (which is not really really a problem because we don't have millions of users) is that we'll update the same user in keycloak everyday for 7 days. Let's discuss this at the DSU.
Deletion is well pushed into keycloak.

shanoir-ng-users/src/main/java/org/shanoir/ng/user/repository/UserRepository.java
```diff
@@ -20,6 +20,7 @@
 
 import org.shanoir.ng.user.model.User;
 import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.redis.core.types.Expiration;
```
michaelkain, line 23: Hi Jean-Come, not sure if this import is still necessary. Cheers, MK

michaelkain (approved):
Hi Jean-Come, please just remove the import and I merge. Best regards, Michael

jcomedouteau:
Done !

michaelkain (approved):

PR #667, opened 2020-11-19
Shanoir issue#172 Extension request

jcomedouteau:


docker-compose/keycloak/themes/shanoir-theme/login/login.ftl
```diff
@@ -41,6 +41,9 @@
 									<span><a href="${url.loginResetCredentialsUrl}">${msg("doForgotPassword")}</a></span>
 								</#if>
 							</div>
+							<div>
+							    <a href="${properties.requestExtension!}">Request an extension.</a>
```
michaelkain, line 45: maybe "Account expired?" would better match with the email sent by Shanoir-NG?

shanoir-ng-users/src/main/java/org/shanoir/ng/extensionrequest/controller/ExtensionRequestApiController.java
```diff
@@ -0,0 +1,71 @@
...
+	@Autowired
+	private KeycloakClient keycloakClient;
+
+	@Autowired
+	private EmailService emailService;
+
+	@Override
+	public ResponseEntity<Void> requestExtension(@RequestBody final ExtensionRequestInfo requestInfo) {
+		try {
+			SecurityContextUtil.initAuthenticationContext("ADMIN_ROLE");
```
michaelkain, line 47: Hi Jean-Come, could we do it without that admin_role sec_context? From my point of view, everything down to the service methods, should be open for access. I think it is for security reasons more dangerous to enable an admin_access here then allowing an outsider to ask for an account extension for a given account, especially as the account remains disabled until the extension request is accepted. With kind regards, Michael

michaelkain: open for access for this request

michaelkain (approved):
in general approved, after code review, please see my minor comments

jcomedouteau:
Done, I created a non protected method to retrieve the user by mail.

PR #696, opened 2020-12-04
Shanoir-issue#687 Import error message management

jcomedouteau:
- Catch error from async method
- Better display error
- Do not display serie sequence but description


Nunien:
@jcomedouteau How can I test it? Do you have a faulty zip?

Nunien (approved):

PR #712, opened 2020-12-14
Shanoir issue#707 subject order in study detail

jcomedouteau:
- Order subjects and users in study detail
- Correct BIDS tree display btw


Nunien (approved):
not tested but seems right

PR #753, opened 2021-01-07
remove app.routing.ts which is unused and add AnimalExaminationService

arthursw:
remove app.routing.ts which is unused and add AnimalExaminationService in app.module.ts


michaelkain (approved):
THX Arthur!!!

PR #761, opened 2021-01-12
Keycloak migration to v12.0.1

michaelkain:
Keycloak side is finished and tested
CSS and graphical aspects in the *.ftl are mostly todos


michaelkain:
Hi Julien, it would be great if you could please take care of the styling. Thank you and with best regards, Michael

PR #762, opened 2021-01-12
Shanoir issue#754 Allow dataset metadata update

jcomedouteau:
Clean form elements
Force updated metadata to get a new ID when updating


jcomedouteau:
@Nunien This pull request is quite simple to test if you want, so we can merge it fast :)
Jean-Côme

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -120,6 +117,10 @@ private Dataset updateDatasetValues(final Dataset datasetDb, final Dataset datas
 		//datasetDb.setReferencedDatasetForSuperimpositionChildrenList(dataset.getReferencedDatasetForSuperimpositionChildrenList());
 		//datasetDb.setStudyId(dataset.getStudyId());
 		datasetDb.setSubjectId(dataset.getSubjectId());
+		if (dataset.getOriginMetadata().getId().equals(dataset.getUpdatedMetadata().getId())) {
+			// Force creation of a new dataset metadata
+			dataset.getUpdatedMetadata().setId(null);
```
Nunien, line 122: This is fine but is it normal that the updated metadata has the same id at this point?

shanoir-ng-front/src/app/datasets/dataset/common/dataset.common.component.ts
```diff
@@ -51,15 +51,15 @@ export class CommonDatasetComponent implements OnChanges {
             private formBuilder: FormBuilder) {
 
         this.exploredEntityOptions = ExploredEntity.options;
-        this.datasetModalityTypes = DatasetModalityType.options;
+        this.datasetModalityTypes = ['Calibration' , 'Ct' , 'Eeg' , 'Meg' , 'Mesh' , 'Mr', 'ParameterQuantification' , 'Pet' , 'Registration' , 'Segmentation' , 'Spect'  , 'Statistical' , 'Template']
```
Nunien, line 54: I have a proposition in order to avoid listing those dataset types here.

* rename `datasetModalityTypes` to `datasetTypes` as those are two different enums
* replace the content of `datasets/shared/dataset-type.model.ts` with : 
```
export const DatasetTypes = [
  'Calibration',
  'Ct',
  'Eeg',
  'Meg',
  'Mesh',
  'Mr',
  'ParameterQuantification',                  
  'Pet',
  'Registration',
  'Segmentation',
  'Spect',
  'Statistical',
  'Template'
] as const;

export type DatasetType = typeof DatasetTypes[number];
```
* `datasetModalityTypes: Option<string>[];` could be `datasetTypes: Option<DatasetType>[];`
* `this.datasetTypes = DatasetTypes.map(prop => new Option<DatasetType>(prop, prop as DatasetType));`

Nunien (changes requested):

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
jcomedouteau: I think that, just after the import, originMetadata and updatedMetadata are pointing to the same entity.
After this, we want to create a new one.

Nunien: But is it normal? Shouldn't updatedMetadata be at least null ?

jcomedouteau: There are some part in the code exclusively using updatedMetadata, so when it does not exists, it points on originMetadata I guess. I agree with you that this is quite weird

shanoir-ng-front/src/app/datasets/dataset/common/dataset.common.component.ts
jcomedouteau: What is the difference between datasetType & datasetModalityType ?

jcomedouteau:
@Nunien I updated the pull request since last time if you want to test it :)

Jean-Côme

jcomedouteau:
This on is testable too @Nunien, I merged the conflicts

PR #777, opened 2021-01-18
Shanoir-issue#775 Subjects order during import

jcomedouteau:


jcomedouteau:
Addendum: Base subject sorting with subjectStudy.name instead of subject.name

Nunien (approved):

PR #789, opened 2021-01-26
Shanoir-issue#764 Shanoir challenge integration

jcomedouteau:
- Add challenge to study
- Allow edit only to administrators
- DB SQL update


Nunien:
I think wee need to discuss that with the team.

What if we want to save data about the challenge, like a start date, a end date or simply a name?
Then we will probably have a Challenge table. And a challenge will be linked to a study (several?). In this case we wouldn't want to have a boolean as this information will already be stored in the relation between Study and Challenge (as a fk or an intermediate table).

Nunien (changes requested):
with bootstrap.sh --clean : 

```
Caused by: org.springframework.jdbc.datasource.init.ScriptStatementFailedException: Failed to execute SQL script statement #2 of class path resource [scripts/import.sql]: INSERT INTO study (id, clinical, coordinator_id, downloadable_by_default, end_date, mono_center, name, start_date, study_status, study_type, visible_by_default, with_examination) VALUES (1, '\0', NULL, '\0', NULL, 1, 'DemoStudy', '2020-11-02 00:00:00', 1, 3, '\0', 1); nested exception is java.sql.SQLException: Field 'challenge' doesn't have a default value
...
Caused by: java.sql.SQLException: Field 'challenge' doesn't have a default value
```

You must either set a default value to challenge or set it in the inserts

jcomedouteau:
I corrected the script and the conflict @Nunien :)

Nunien (changes requested):
The merge from develop is bad : 
Capture d’écran de 2021-02-10 10-30-01
You should orverride the updated ftl files with the one from develop

jcomedouteau:
Mmm done ? I guess? Did you revert the keycloak change ?

Nunien:
No, the first keycloak pr merge did not contain my fixes. I added them to develop afterward.

Nunien:
* Now the ftl seems fine but you need to merge the develop login.css and common.css (common may be good alreadt I'm not sure)
* Other problem : when I click on "subscribe to a challenge" It just reloads the login page

Nunien (dismissed):

Nunien (approved):

Issue #800, opened 2021-02-04
[Bruker import] Create exam error


michaelkain:
user type: expert
browser: Firefox

I have created my new subject during the import and want to create a new examination, then I get the error:
Error: Uncaught (in promise): $w: {"headers":{"normalizedNames":{},"lazyUpdate":null},"status":403,"statusText":"OK","url":"https://shanoir.irisa.fr/shanoir-ng/datasets/examinations","ok":false,"name":"HttpErrorResponse","message":"Http failure response for https://shanoir.irisa.fr/shanoir-ng/datasets/examinations: 403 OK","error":{"code":403,"message":"Accès refusé","details":null}}

jcomedouteau:
at org.shanoir.ng.examination.controler.ExaminationApiController$$EnhancerBySpringCGLIB$$5ecfcfab.saveNewExamination(<generated>)

I'll check it :)

michaelkain:
solved, to be finally tested by me on qualif+prod

PR #805, opened 2021-02-05
Shanoir-issue#804 #800 Bruker import

jcomedouteau:
- Update dicomifier reference for bruker conversion
- Allow examination creation by expert (use examination DTO)

Excuse me @michaelkain if I had tested the bruker import yesterday instead of trying to solve 'rabbitmq' problem in the tests I would have seen the problem :(

I Updated dicomifier to use the proper command, and tested both dicom2bruker and dicom2nifti.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -203,7 +203,7 @@ services:
   #
   preclinical-bruker2dicom:
     container_name: "${SHANOIR_PREFIX}preclinical-bruker2dicom"
-    image: jcomedouteau/dicomifier.ws:1.2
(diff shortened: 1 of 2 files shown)


michaelkain (approved):
Hi Jean-Come, thank you for your fix! I will directly test it on qualif on Monday.
It can only get better for Bruker import :-). I agree with you, but to be honest the
rabbitmq error is annoying too. Now as I really develop in the back, we should have
solved it months ago, as it pollutes the log and we will miss importants things like this.
Thank you for having worked on a fix. Cheers, Michael

PR #809, opened 2021-02-08
fix SHANOIR_COLORS not working with angular 10

a-ba:
Since the migration to angular 10, the css are minified. The regex
overriding the colors from the `SHANOIR_COLORS` env var had to be updated.


jcomedouteau:
@Nunien What should we do with this ?

jcomedouteau:
@Nunien Do we still need this ?

PR #814, opened 2021-02-11
Shanoir-issue#621 import warning on center/acquisition equipement

jcomedouteau:
- Base DICOM/select check on serial number only first, then name to
avoid misleading error messages


michaelkain:
Hi Jean-Come, I am not convinced, that this fixes the original issue. We could more easily loose the warning, yes, but we risk to have no warning if the serial number is missing or does not match, and I am not sure if this is what we want. With kind regards, Michael

michaelkain:
please see my answer to Elise in the issue

jcomedouteau:
The fact is that the warning is present almost 100% of the time because of this check (the name is never really 'equal'). And some people seems to think that it is an error so they cannot go further for their import. We'll see Elise's response :)

michaelkain (approved):

PR #818, opened 2021-02-11
Shanoir issue#793 Study deletion protection

jcomedouteau:
Protect study from deletion:
- If there is still exams
- Not from the list


michaelkain:
Hi Jean-Come, would it be fine for you to merge the conflicts? I would then review and test the PR. With kind regards, Michael

jcomedouteau:
Conflicts merged, I'm retesting it

jcomedouteau:
Hopefully I retested it..
There is an SQL script to execute on database !

shanoir-ng-studies/src/main/java/org/shanoir/ng/study/service/RabbitMQStudiesService.java
```diff
@@ -75,17 +75,68 @@
 				return Collections.emptyList();
 			}
 			// Filter administrators and map to get only IDs
-			List<Long> ids = study.getStudyUserList().stream()
+			return study.getStudyUserList().stream()
 					.filter(studyUser -> studyUser.isReceiveNewImportReport())
```
michaelkain, line 79: Hi Jean-Come, why is here the flag receiveNewImportReport checked and not the right can_administrate? Maybe I am overseeing something. With kind regards, Michael

michaelkain:
Hi Jean-Come, let me know if after my code review I should start local testing. With kind regards, Michael

shanoir-ng-studies/src/main/java/org/shanoir/ng/study/service/RabbitMQStudiesService.java
jcomedouteau: I just corrected a sonarlint warning here stating that we should not create a new variable "ids" just before returning it.

This method is used in EmailServiceImpl.getStudyAdministrator to get the list of users of the study that needs to receive the import mail. I think the method is just really really badly named.

michaelkain:
Hi Jean-Come, I tested locally, please see my screenshot:
Bildschirmfoto 2021-03-26 um 15 27 00
My import finished and I created a new subject + exam in a new study. I think the display in the tables uses the same Set<Long> of examinationIds you use for your check, but it does not seem to be synchronised (I tried with user type user  and user type admin)?

michaelkain:
yes, and I could delete even with existing exam in ms datasets

michaelkain:
The deletion of an empty study was still working successfully

jcomedouteau:
Did you pass the sql script ? 20210211_study_examination_link.sql  

I made a little change, but it feels like rabbitMQ needs a restart to be effective for the changes :/
(I first tested, it didn't work, then restarted then redo the exact same thing and it worked it all)

michaelkain (approved):
Code review approved, I will test locally again

michaelkain:
Hi Jean-Come, I tested locally and it still does not seem to work yet, so please do not merge this pull request yet.
I created a new study, empty, and could directly delete it after creation, what is correct from my point of view.
Then I created a study and imported some data. My first import try was strange, I did the import until the exam
creation clicked next and ended up on the home page, with not import done/started, but my subject and exam
had been created. My second import try worked, so I now had two exams in my study: a) empty b) second import
filled with one serie. I tried to delete the study and it did not work: OK! I deleted both subjects to clean the study,
so no exam should remain anymore, see my screenshot, but I could still not delete the study in the end.
Bildschirmfoto 2021-04-23 um 16 53 05

michaelkain:
by the way I had your last commit on my test branch and was up-to-date with develop
and I did a FULL **bootstrap.sh --clean** to wipe out everything

jcomedouteau:
Well seen, deleting a subject does trigger an examination deletion but not its associated event.

michaelkain:
Hi Jean-Come, thank you for your modifications. I try to understand the changes and maybe I am wrong, but I am wondering about something: ms studies contains a mirror of study_exam (used by you to protect delections of studies/subjects if exams exist) and ms datasets contains the exams, why when I delete a subject you do not clean study_exam in ms studies at first locally and then send the subject delete event? This would avoid you after deleting an exam to communicate back to ms studies and could save quite some communication? In case you would worry about data consistency, as ms datasets is the "master" for the relation study_exam, if something would be wrong in ms studies would be an error to live with in a microservice system? Maybe I am wrong.

jcomedouteau:
Michael, if I understand the problem: the fact is that in ms studies we have no information on the link subject / exam.

So when I delete a subject in ms studies, I cannot link them to exams but by communicating with ms datasets.
Study_examination table is only an study_id / exam_id table  with no informations on subjects.

Maybe we could complete this table but I'm not 100% convinced

michaelkain:
forget, what I wrote

michaelkain:
I looked deeper: there is already a sync communication for the DELETE_EXAMINATION_EVENT, as exams can be deleted independently in ms datasets, where we need to update study_exam in ms studies

michaelkain:
thx for the explanation

michaelkain:
Hi Jean-Come, I continued my tests this morning and unfortunately it does still not work on my machine. I have pulled your last commits on the PR and made a boostrap.sh --clean, and I still have the following issue, I can delete a study as admin even having an exam in the study, see my screenshot. I have made an import with dicom.zip into the study and created the exam during the import, what should block me from deleting the study, but it works I can delete. Maybe I am doing something wrong?
Bildschirmfoto 2021-05-04 um 10 46 18

jcomedouteau:
Can you try to check in your DB if the study_examination was correctly created ? I think not if you could delete it.

I encountered a problem with rabbitMQ multiple times, when you make a bootstrap.sh -- clean, it deletes all rabbitMQ configurations/queues.
So when the first event is catched, it only creates the adapted queue (here study-examination-queue) but it does not trigger the real method.
I don't really know how to solve this.

But if you try a seconde time, everything will be working fine I thinks.

michaelkain:
Hi Jean-Come, I would like to finish this PR first, as I worked already on it. Could you please solve the merge conflicts? Thank you and with kind regards, Michael

jcomedouteau:
Yes ! I'll retest it fully as it's been quite some time :)

jcomedouteau:
The rabbitMQ communication seems to fail locally, I'm investigating

jcomedouteau:
@michaelkain That should be better.
Once again, you may have to restart the app twice in ordre to initialize the new rabbitMQ queue.

jcomedouteau:
I thnk the problem with rabbitMQ I had is linked to already existing queue in rabbitMQ that "stops" the binding of the new queue. 
I'm testing it on qualif to see if it works

michaelkain (approved):

PR #835, opened 2021-02-19
Data User Agreement + Challenge

Nunien:


michaelkain:
Hi Julien, sorry for the merge conflict. I guess it is related to the study file deletion PR I accepted right now. Let me know if I can be of any help! THX for your PR, the demo was great!!!

Nunien:
@michaelkain I'm worried about the fact that an admin can change the flag "confirmed" when editing a study. I mean, we may not provide the possibility on the UI but with the right REST call he can do it. Is it really ok? (one consequence is that I have to be careful about what the front sends back to the backend on creating/editing a study)
Maybe it is related to another point : the "confirmed" flag is not absolutely necessary in the study ms database. It's good that we send it in the DTO from the back to the front but it could be calculated (in the back) at this moment from the data we have. It is needed in the other ms replicating study_user though. 
Also if we remove the confirmed flag in the study db, we may not want to do the calculation when we fetch the whole study list, it will be maybe too many joins.

michaelkain (commented):
Hi Julien, my pure code review looks really good! I will try to test it Monday morning already on my machine. Cheers, Michael

michaelkain:
Hi Julien, thanks for your comments. I prefer having a clear replicated flag instead of calculated infos concerning the rights. Less brain challenging when checking the data, even if somehow a duplicated data. I agree on your concern with the confirmed right, that is a good point. Why do we not just check in the backend with the userId of the token, that it is really our user, that confirms? So nobody else can confirm for somebody else? Just an idea. With kind regards and thank you for this cool development!!! Cheers, Michael

michaelkain:
by the way, it is not a showstopper for the go-live, I could fix this next week

shanoir-ng-front/src/app/dua/dua-signing/dua-signing.component.html
```diff
@@ -0,0 +1,23 @@
...
+along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
+-->
+
+<h3>{{dua.studyName}}</h3>
+<embed *ngIf="pdfUrl" class="embed-pdf" [src]="pdfUrl" alt="pdf" pluginspage="http://www.adobe.com/products/acrobat/readstep2.html">
+<div>
+    <checkbox [(ngModel)]="checked"></checkbox> I have read the data user agreement contract for the study <span class="study-name">{{dua.studyName}}</span> and I fully accept it.
+    <div class="buttons">
+        <button class="accept" [disabled]="!checked" (click)="accept()">Accept</button>
+        <!-- <button class="refuse" (click)="refuse()">Refuse</button> -->
```
jcomedouteau, line 21: So if the user does not accept, he'll be asked for the DUA signing every time he connects ?

shanoir-ng-front/src/app/dua/dua.component.html
```diff
@@ -0,0 +1,19 @@
...
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
+-->
+
+<h2 class="header command-zone">Data User Agreements signing</h2>
+<div *ngIf="duas">
+    <dua-signing *ngFor="let dua of duas" [dua]="dua" (sign)="loadDua()"></dua-signing>
```
jcomedouteau, line 17: I did not test yet, but when the users signs a DUA, does it reloads the list of DUAs to sign ?

shanoir-ng-front/src/app/home/home.component.ts
```diff
@@ -30,11 +31,49 @@ export class HomeComponent {
...
+
+    onSign() {
+        this.fetchChallengeStudy();
+    }
+
+    private fetchChallengeStudy() {
+        this.studyService.getAll().then(studies => {
+            if (studies) {
+                for (let study of studies) {
+                    if (study.challenge) {
```
jcomedouteau, line 66: So we hope not to get multiple challenges here ?
It's OK for me too, as normally a user is subscribing to shanoir only for the challenge

shanoir-ng-front/src/app/dua/dua-signing/dua-signing.component.html
Nunien: I think it is ok for now

shanoir-ng-front/src/app/dua/dua.component.html
Nunien: Yes : (sign)="loadDua()"

shanoir-ng-front/src/app/home/home.component.ts
Nunien: You are right, this version supports only one challenge per user

shanoir-ng-front/src/app/shared/side-menu/side-menu.component.html
```diff
@@ -34,23 +34,17 @@
...
 <ul class="menu-list">
     <li routerLink="/" routerLinkActive="active"><i class="fas fa-home"></i>Home</li>
-    <li (click)="tasksOpened = !tasksOpened" class="notifications">
-        <i class="fas fa-tasks"></i>Tasks
-        <span *ngIf="notificationsService.nbProcess > 0" class="notif">( {{notificationsService.nbProcess}} pending - {{notificationsService.totalProgress() * 100 | number:'0.0-0'}}%)</span>
-        <!-- <button class="awesome" (click)="notificationsService.refresh()" >
-            <i class="fas fa-sync"></i>
-        </button> -->
-        <span class="chevron"><i class="fas fa-chevron-down"></i></span>
+    <li *ngIf="duasToSign() > 0" (click)="notifOpened = !notifOpened"><i class="far fa-flag"></i>
```
jcomedouteau, line 37: We need to make an issue about the refactoring of jobs/tasks, to display a new alert for download/import/etc..

shanoir-ng-front/src/app/home/home.component.html
```diff
@@ -12,13 +12,35 @@
...
+        <hr/>
+    </div>
+    <div *ngIf="!challengeStudy">
+        <h1>Welcome to Shanoir, challenger!</h1>
+        <p class="msg">Here you will be able to download the datasets for the {{challengeDua.studyName}} Challenge.
+        But before that that, we ask you to accept the following Data User Agreement : </p>
+        <dua-signing (sign)="onSign()" [dua]="challengeDua"></dua-signing>
+    </div>
+    <div *ngIf="challengeStudy" class="dl-link">
+        <p class="msg">Congratulation, you can now download all the ressources you need for the <span class="challenge-name">{{challengeStudy.name}}</span> Challenge :</p>
```
jcomedouteau, line 38: That would be cool to have a link directly to the study ? :)

michaelkain:
Hi, GREAT!!! Please see my test report below, I found nothing wrong:
Testing of PR830
- start testing without challenge aspect, pure DUA usage only
- Study has no DUA yet, user type user is member, can connect and see (yyao)
- Attach DUA to study and login with yyao
--- mkain, as admin, is directly asked to sign the DUA, what is correct, as in v1.0
we do not separate roles or users from DUA requirement as all can access and use the data,
in the idea, that a study responsible requires a DUA
- Used expert user (jlouis) and user user (yyao), accepted the DUA and tested data access
---> Works, no access to data before DUA accepted, and data access afterwards
- Good point: admin could not administrate the confirmed flag
- Now testing with challenge aspect
--- Challenge flag -> true + one big zip file added
--- Login with yyao: not asked for DUA and access to file on home page, correct
------> Works
- New study created to test DUA combined with big file download, mkain creates study
--- Works: could directly accept DUA as admin and access to challenge files
--- Works: could directly accept DUA as jlouis and access to challenge files
--- Minor thing: I forgot to add the user with download rights, so I had an error in the GUI,
as the user did not have the correct rights

jcomedouteau:
I tested something interesting:
=> An expert user that can manage a study with  no DUA, when the user adds a DUA => he then has no access to the study so an error pops => Maybe he should then be redirected to the DUA signing form ?
Or the user that adds the DUA implicitely signs it ? (and so the 'confirmed' is set to 'true' ?)

michaelkain:
interesting, when I tested with my admin to manage the study, the notification came directly up and I could not really miss it after having added the dua, but it is maybe related that admin always has more power, but good point, testing with expert would have been more realistic, we could directly set the uploading expert to confirmed, why not

PR #839, opened 2021-02-23
dl link for dua pdf while signing

Nunien:


michaelkain:
Hi Julien, I will test on my machine. Thx for the PR, cheers, Michael

michaelkain (approved):
Hi Julien, my local tests worked totally fine. I could download the DUA with the link and accept it. The pdf was displayed in my browser too. Thx and cheers, Michael

PR #844, opened 2021-02-25
DUA: protect acceptance -> only user itself can accept

michaelkain:


Nunien (changes requested):
As we discussed, you should maybe use @PreAuthorize instead

Nunien (changes requested):
As we discussed, you should maybe use @PreAuthorize instead

shanoir-ng-studies/src/main/java/org/shanoir/ng/study/dua/DataUserAgreementService.java
```diff
@@ -44,32 +45,28 @@
...
-	}
-	
-	public void createDataUserAgreementsForStudy(Study study) {
-		List<StudyUser> studyUserList = study.getStudyUserList();
-		for (StudyUser studyUser : studyUserList) {
-			DataUserAgreement dataUserAgreement = new DataUserAgreement();
-			dataUserAgreement.setStudy(study);
-			dataUserAgreement.setUserId(studyUser.getUserId());
+		// assure that only the user itself can accept its DUA
+		if (dataUserAgreement.getUserId().compareTo(userId) == 0) {
```
jcomedouteau, line 51: Why not using "equals" for code clarity ? (change it only if using @PreAuthorize as Julien was stating).

We could add some unit tests at the same time

PR #851, opened 2021-03-04
Shanoir-issue#850 StudyUser new user -> Confirmed set accordingly

jcomedouteau:


shanoir-ng-studies/src/test/java/org/shanoir/ng/study/service/StudyServiceTest.java
```diff
@@ -209,11 +285,12 @@ private Study createStudy() {
 		return study;
 	}
 	
-	private StudyUser createStudyUsers(Long suId, Long userId, Study study, StudyUserRight... rights) {
+	private StudyUser createStudyUsers(Long suId, Long userId, Study study, boolean confirmed, StudyUserRight... rights) {
 		StudyUser studyUser = new StudyUser();
 		studyUser.setId(suId);
 		studyUser.setStudy(study);
 		studyUser.setUserId(userId);
+		studyUser.setConfirmed(true);
```
michaelkain, line 293: Hi Jean-Come, thank you a lot for this great fix! I guess here you want to use the value of confirmed, and not a fix value?

michaelkain: I am wondering, how the UnitTest runs in case of that value per default

shanoir-ng-studies/src/test/java/org/shanoir/ng/study/service/StudyServiceTest.java
```diff
@@ -183,24 +188,95 @@ public void updateTest() throws AccessDeniedException, EntityNotFoundException,
...
 
 		studyService.update(updated);
 	}
 
+	@Test
+	@WithMockKeycloakUser(id = 3, username = "jlouis", authorities = { "ROLE_EXPERT" })
+	public void testUpdateStudyUsersNoDUA() {
+		Study existing = createStudy();
+		existing.setStudyUserList(new ArrayList<StudyUser>());
+		existing.getStudyUserList().add(createStudyUsers(1L, 1L, existing, false, StudyUserRight.CAN_SEE_ALL, StudyUserRight.CAN_IMPORT));
```
michaelkain, line 214: this should be true here

michaelkain (changes requested):
Hi Jean-Come, thank you for this great fix! Could you please adapt your UnitTest, I think there is an error in it. Thank you for having added it, I will try to get more used to this kind of test oriented dev more too. It would be great to deploy it this evening. With kind regards, Michael

shanoir-ng-studies/src/test/java/org/shanoir/ng/study/service/StudyServiceTest.java
jcomedouteau: In this testing I am testing with no DUA:
So I create a user with "false" set  in 'confirmed' field and then verify that it has been updated to true at the end

shanoir-ng-studies/src/test/java/org/shanoir/ng/study/service/StudyServiceTest.java
jcomedouteau: Damn.

michaelkain:
Hi Jean-Come, I tested on my local machine. Your fix works for an existing study, that is updated, what is the most important case. In case of a new study is created, the setConfirmed(true), was still missing. I inserted it to be complete. Hope this is fine. I will test again and then accept the PR. Thank you a lot!

michaelkain (approved):

Issue #642, opened 2020-11-03
Create subject


ebannier2018:
In a previous issue which I cannot find in the list, I mentionned the fact that some fields for the subject creation were not mandatory (subject identifier for this study, Physically Involved) and could be interpreted as mandatory if they where highlighed by the GUI. This was changed to something more discreet which is nice. Yet I had mentionned that the subject type (Healthy or Patient) was an important information which is now hidden. If you cannot separate all 3, we should show them all and add * to the mandatory ones.

Nunien:
@ebannier2018 what if I just move subject type up and make it mandatory ? : 
![Capture d’écran de 2020-12-04 13-12-42](https://user-images.githubusercontent.com/21199767/101162313-77235100-3632-11eb-983f-1af4ad0ff7cd.png)

Nunien:
(with the missing *)

ebannier2018:
sounds good 

-------------------------------------------------- 
MRI Physicist, PhD 
Plate-forme NeurInfo, https://www.neurinfo.org/ 
CHU Rennes, https://www.chu-rennes.fr/ 
Unité Empenn (ex-VISAGES) U1228 INSERM-INRIA, IRISA UMR CNRS 6074, https://team.inria.fr/empenn/ 
Membre du Réseau d'Entraide Multicentrique en IRM, https://remi.network/fr/ et de la SFRMBM, https://sfrmbm.fr/ 

+33 2 99 28 99 06 (Neurinfo) 
+33 2 99 84 22 13 (Empenn) - Fridays

ebannier2018:
Hi, Sorry but I do not see that the subject type (healthy control or patient) is now mandatory 
Elise

PR #878, opened 2021-03-22
make subject type mandatory #642

Nunien:


Nunien:
Sorry it has been previously merged ... but in master. My bad.

PR #897, opened 2021-04-08
Shanoir issue#895 Manage 4D resting state data

jcomedouteau:
4D restng state data is organized with multiple folders containing same dicoms but with a different timing.
The problem is that the dicom during the import are not well organized => Same path for every subfolder/dicom so the download fails in Shanoir.

I re-organized the download so that if there is twice the same path, we just add subfolders.


jcomedouteau:
PM me for data to test

michaelkain (approved):
Code review approved

michaelkain:
Hi Jean-Come, could you share the test data with Julien next week to finish it or with me the week after?

jcomedouteau:
When importing resting state 4D generated by philips, a lot of dicoms are generated following this pattern:

> To create these, click on "Patient/Administration/Disk Files/" on the console which will export the selected scan series to a Windows folder, each image in a separate DICOM file. The maximum number of images in a directory is 2048, so additional images are put in subdirectories called 000001, 000002, etc
 
All these dicoms have exactly the SAME name (000001/001.dcm 000001/002.dcm etc..), so when we import them one by one, they are all copied in a single folder and they all erase the previous dicoms (so only the last "folder" is taken into account during the import).
This means the dataset at the end is completely false.

So I added the "folder" information during the import so that they are not all copied in the same and we keep all dicoms

jcomedouteau:
@Nunien @michaelkain Michel is asking me if it's possible to check this PR quite soon? 
Thanks !
Jean-Côme

michaelkain:
I will test this tomorrow on qualif, as my current env is building right now to test the study deletion protection PR again, if okay for Michel we deploy it next Wednesday?

PR #901, opened 2021-04-09
angular style check in github

Nunien:


jcomedouteau:
The goal is to write some coding rules for angular, so I think we can close this pull request.

jcomedouteau:
I'm closing this too. For the moment, we'll just keep the written coding rules in wiki.

PR #929, opened 2021-05-05
Dua only user itself as annotation

michaelkain:
test protocol, done with local web GUI + SoapUI:
- new study created with DUA attached and two members added into study
- DUA of user 1 validated, logged in with user 1, worked
- tried to validate DUA of user 2, logged in with user 1, forbidden


jcomedouteau:
{"code":500,"message":"Could not validate the data user agreement acceptance. The user of the DUA and the used user for the request do not match."}

OK for me !

jcomedouteau (approved):

PR #941, opened 2021-05-10
Shanoir issue#930 Import email correction

jcomedouteau:


Nunien (approved):
I don't see the examination date as asked in the issue. I'm approving anyway since it works well and the users are waiting. It can be improved later.

Nunien:
Ok I made the fix on develop ( </span> )

It's working.

Issue #947, opened 2021-05-17
Old datasets are not accessible anymore


jcomedouteau:
Every dataset imported before the migration are not accessible anymore:
Cannot find dataset with id 250692

This is due to this commit:
https://github.com/fli-iam/shanoir-ng/commit/4415d83eeda157f5ed9b9ea33edebbf7a81b3069

The FetchType.EAGER loading DatasetAcquisition forces to load all datasetAcquisition when loading a dataset. The problem is located here:
In MrProtocol for OLD datasets:
- originMetadata = NULL & updatedMetadata = XXXX
In MrProtocol for NW datasets:
- originMetadata = XXXX & updatedMetadata = NULL;

This fails for some reason.

So we have to "exchange them:

```
# Update all
UPDATE mr_protocol SET origin_metadata_id = updated_metadata_id WHERE origin_metadata_id IS NULL and updated_metadata_id IS NOT NULL;

# Remove updated_metadata_id where they are equal (when they are different it's not a problem)
UPDATE mr_protocol SET updated_metadata_id = NULL WHERE origin_metadata_id IS NOT NULL and updated_metadata_id IS NOT NULL AND origin_metadata_id = updated_metadata_id;
```

jcomedouteau:
Edit: It's more a problem of data type:
UPDATE mr_protocol_metadata SET dtype = 2 WHERE id IN (SELECT updated_metadata_id FROM mr_protocol WHERE updated_metadata IS NOT NULL);

PR #948, opened 2021-05-17
Shanoir-hotfix#947 Correct access to old datasets

jcomedouteau:
See #947 

Datasets before migration were not accessible anymore. Update database to correct this.


Nunien:
The script has been executed on both prod servers, I think it's over.

jcomedouteau:
I really think we have to merge it to keep track of what happend on the server.

PR #950, opened 2021-05-31
Fix shanoir downloader

arthursw:
- add the datasetIds option to download dataset in batch (path to a textfile containing one dataset id per line)
- properly initialize ShanoirDownloader logs
- added downloadDataset.py the python equivalent to ShanoirDownloader


michaelkain:
waiting for final extensions by Arthur

Issue #931, opened 2021-05-10
Additional metadata : Magnetic Field Strengh


Nunien:
Save this dicom field during import : 
(0018,0087) Magnetic Field Strength

It should be saved in a new property in MrProtocol

PR #955, opened 2021-06-01
Shanoir issue#937 Magnetic field strength + creation date

jcomedouteau:
Store and display creation date on dataset acquisition


jcomedouteau:
I added https://github.com/fli-iam/shanoir-ng/issues/931

jcomedouteau:
How to test this pull request:
- Import MR data in Shanoir
- Get on the newly created dataset acquisition
- Check that the Magnetic field strength is well displayed in accordance with what is in the DICOM

Nunien (approved):

db-changes/datasets/01062021_dataset_acquisition_import_date.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE dataset_acquisition ADD COLUMN creation_date date;
```
jcomedouteau, line 1: We should totally deploy on qualif everytime, because a new column is missing here for mr_protocol.magnetic_field_strength.

PR #965, opened 2021-06-11
Solr new version - 1rst version

Nunien:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/repository/SolrRepositoryImpl.java
```diff
@@ -19,147 +26,125 @@
...
-				} else if (studyName.startsWith(WILDCARD) && studyName.endsWith(WILDCARD)) {
-					criteria.and(Criteria.where(STUDY_NAME).contains(studyName.substring(1, studyName.length() -1)));
-				} else if (studyName.startsWith(WILDCARD)) {
-					criteria.and(Criteria.where(STUDY_NAME).endsWith(studyName.substring(1, studyName.length())));
-				} else if (studyName.endsWith(WILDCARD)) {
-					criteria.and(Criteria.where(STUDY_NAME).startsWith(studyName.substring(0, studyName.length() - 1)));
-				}
-			}
+	
+	private void addAndPredicateToCriteria(Criteria criteria, String fieldName, Collection<String> values) {
```
jcomedouteau, line 59: If we put a wildcard here, does it works ?( I'll test it after code review :) )

shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/repository/SolrRepositoryImpl.java
```diff
@@ -19,147 +26,125 @@
...
-				} else if (datasetNature.startsWith(WILDCARD) && datasetNature.endsWith(WILDCARD)) {
-					criteria.and(Criteria.where(DATASET_NATURE).contains(datasetNature.substring(1, datasetNature.length() -1)));
-				} else if (datasetNature.startsWith(WILDCARD)) {
-					criteria.and(Criteria.where(DATASET_NATURE).endsWith(datasetNature.substring(1, datasetNature.length())));
-				} else if (datasetNature.endsWith(WILDCARD)) {
-					criteria.and(Criteria.where(DATASET_NATURE).startsWith(datasetNature.substring(0, datasetNature.length() - 1)));
-				}
-			}
+		
+		if (facet.getSearchText() != null && !facet.getSearchText().trim().isEmpty()) {
```
jcomedouteau, line 80: You can use StringUtils.isBlank()

shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/repository/SolrRepositoryImpl.java
```diff
@@ -19,147 +26,125 @@
...
-		SolrResultPage<ShanoirSolrDocument> result = (SolrResultPage<ShanoirSolrDocument>) solrTemplate.queryForPage(query, ShanoirSolrDocument.class);
-		FacetPage<ShanoirSolrDocument> result2 = solrTemplate.queryForFacetPage(query, ShanoirSolrDocument.class);
-
-
-		return (SolrResultPage<ShanoirSolrDocument>) result2;
+		try {
+			FacetPage<ShanoirSolrDocument> result = solrTemplate.queryForFacetPage(query, ShanoirSolrDocument.class);			
+			return (SolrResultPage<ShanoirSolrDocument>) result;
+		} catch (UncategorizedSolrException e) {
+			ErrorModel error = new ErrorModel(HttpStatus.UNPROCESSABLE_ENTITY.value(), "solr query failed");
```
jcomedouteau, line 105: The exception is never logged here, it could be interesting for us to keep track of it in the logs.

Nunien: It's supposed to be logged when the RestServiceException is catch

jcomedouteau: Yes but you never use the initial exception, in you want it to be logged, you should do 
throw new RestServiceException(error, e); I think

jcomedouteau:
After tests:
It works really well, it's really better than before, well done @Nunien  :)
A little thing: 
- For the dates, when you remove a date filter from the left panel it does not removes it from the right panel => It can provokes a little bug
1) Put some dates filters with which you have no data (two consecutive days)
2) Remove those dates from the left panels
3) if you remove the filters from the top filter bar, the filter left does not work

Otherwise than this, everything seems to work fine as an expert.

jcomedouteau (approved):

PR #972, opened 2021-06-21
bootstrap.sh: --no-keycloak + --no-dcm4chee flag

michaelkain:
- bootstrap.sh: start with or without Keycloak
- docker-compose.yml: depends_on to Keycloak removed from sh-ng microservices

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -5,7 +5,7 @@ print_help()
 	cat <<EOF
 Build and deploy Shanoir NG
 usage:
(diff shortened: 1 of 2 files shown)


michaelkain:
Local tests executed:
- --no-keycloak flag used: OK, runs up without Keycloak
- --no-keycloak flag removed: OK, runs up with Keycloak (db+app server)

michaelkain:
MK: solution for my problem with Apple Silicon M1 on my MacBook Air and not able to run the Keycloak container anymore inside docker-compose (tested with 12.0.3+13.0.1+ all possible new platform flags introduced into buildx, no success), now installed with external Keycloak

PR #979, opened 2021-07-01
Solr : selection list + deletion

Nunien:
* selection list, what to test
    * select datasets
    * navigate between the 2 tabs : selection list and result list
    * edit search while in selection -> switches to results
    * navigating back from another page should restore saved selection
    * test "clear selection" button
* delete feature
    * try to delete datasets not in a study the current user administrates (forbidden)
    * try to mix authorized and forbidden datasets
...


jcomedouteau:
I tested it, it seems to be working well ! Just checking the code a little bit

jcomedouteau (approved):

PR #980, opened 2021-07-07
Solr : Center + Save State

Nunien:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/model/ShanoirSolrDocument.java
```diff
@@ -76,13 +76,18 @@
 	@Indexed(name="studyId", type="Long")
 	private Long studyId;
 	
+	@Field
+	@Indexed(name="canterName", type="string")
```
jcomedouteau, line 80: canterName ? or centerName ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/repository/SolrRepositoryImpl.java
```diff
@@ -114,8 +116,13 @@ private void addExpertClause(Criteria criteria, String searchStr) {
 	
 	private void addSearchInAllClause(Criteria criteria, String searchStr) {
 		if (searchStr != null && !searchStr.isEmpty()) {
-			String[] fields = {"studyName", "subjectName", "datasetName", "examinationComment", "datasetType", "datasetNature"};
-			String[] searchTerms = searchStr.replace(Criteria.WILDCARD, " ").trim().split(" ");
+			String[] fields = {"studyName", "subjectName", "datasetName", "examinationComment", "datasetType", "datasetNature", "centerName"};
+			String[] specialChars = {"+", "-", "&&", "||", "!", "(", ")", "{", "}", "[", "]", "^", "\"", "~", "*", "?", ":", "/"};
+			String escapedSearchStr = searchStr;
```
jcomedouteau, line 121: What is the goal of these lines ?

shanoir-ng-ms-common/src/main/java/org/shanoir/ng/shared/configuration/RabbitMQConfiguration.java
```diff
@@ -64,6 +64,9 @@
 	/** Subject name updated => notify dataset MS to change database. */
 	public static final String SUBJECT_NAME_UPDATE_QUEUE = "subject-name-update-queue";
 	
+	/** Subject name updated => notify dataset MS to change database. */
```
jcomedouteau, line 67: Update comment please :)

shanoir-ng-studies/src/main/java/org/shanoir/ng/center/service/CenterServiceImpl.java
```diff
@@ -82,9 +96,49 @@ protected Center updateValues(final Center from, final Center to) {
...
+	}
 
 	@Override
 	public Center findByName(String name) {
 		return centerRepository.findByName(name);
 	}
+	
+	private boolean updateName(IdName idName) throws MicroServiceCommunicationException{
+		try {
+			rabbitTemplate.convertAndSend(RabbitMQConfiguration.centerNameUpdateQueue().getName(),
```
jcomedouteau, line 137: Can you use the constant CENTER_NAME_UPDATE_QUEUE here please ? It's how it's done everywhere else and I find it more usefull when looking for queue's usage

shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/amqp/RabbitMQDatasetsService.java
```diff
@@ -89,41 +95,43 @@ public void receiveMessage(String commandArrStr) {
...
-				Subject newSubject = new Subject(receivedSubject.getId(), receivedSubject.getName());
-				subjectRepository.save(newSubject);
+				// create new entity
+				try {
+					T newOne = clazz.newInstance();
+					newOne.setId(received.getId());
+					newOne.setName(received.getName());
+					repository.save(newOne);
+				} catch ( SecurityException | InstantiationException | IllegalAccessException | IllegalArgumentException e) {
+					throw new IllegalStateException("Cannot instanciate " + clazz.getSimpleName() + " class through reflection. It is a programming error.", e);
```
jcomedouteau, line 133: This is quite "dangerous" to use IllegalStateException in a @RabbitHandlerMethod
=> If the catched exception is not AmqpRejectAndDontRequeueException, then the message well be requeued and then it'll loop over and over

jcomedouteau: @Nunien Can you update this to use something else than an IllegalStateException so that it does not requeue in an infinite loop ?

jcomedouteau:
Can you telle me how to test the "save state" part ?

I have this error when I try to index data to solr:
org.springframework.orm.jpa.JpaSystemException: Could not resolve column name in result set [centerName]; nested exception is org.hibernate.HibernateException: Could not resolve column name in result set [centerName]

Edit: I really don't understand this error as you set up all fields correctly in the Metadata class..

Nunien:
* I think the "save state" feature is when, from solr, you go to another page (for instance you open a dataset) and you go back to solr : your search + pagination must be preserved.
* you must execute db-changes/datasets/18062021_copy_center_id_name.sql which creates the center name column.

michaelkain:
Hi Julien, can we merge this to finish the new search for deployment?

Nunien (commented):
@jcomedouteau I have put AmqpRejectAndDontRequeueException instead of IllegalStateException

shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/amqp/RabbitMQDatasetsService.java
jcomedouteau: Yes !

PR #993, opened 2021-07-28
Shanoir-issue#991 Clean users.log from AsyncTask exceptions

jcomedouteau:


jcomedouteau:
I tested it locally, I catch well the exception.

To test it:
- Open a session
- Import some data
- Go to jobs
- Wait ~15mn until user session ends
-> No logs

michaelkain (approved):

PR #1003, opened 2021-08-16
Import extensions: better manage enhanced + etc.

michaelkain:


michaelkain:
Tested with my automatic SOAPUI tests: MR DicomDir + MRS NoDicomDir and both work.
MRS and E (Enhanced) are displayed correctly in 2. Select Series. The MRS dataset is tagged
as SPECTRO and no nifti conversion is done, neither proposed on view dataset details.
I will have to test the pacs import on qualif to assure that it functions for pacs import too.

PR #1007, opened 2021-08-23
Potential fix for issue #990 + detect patientIdentityRemoved (front+b…

michaelkain:
…ack) + deploy.sh


michaelkain:
I have locally tested everything, especially the change in the anonymization module: with debugger all cases tested: birthDate==null,birthDate=="", birthDate==1980 and birthDate==19790101, all worked

PR #1011, opened 2021-08-25
Shanoir issue#996 Mail when import fails

jcomedouteau:
Make an import fail (add a "throw new exception" in ImporterManagerService or ImporterService and check that a mail is well sent.


Nunien (approved):
How I tested : 
* start an import process
* select an examination
* go next
* on another tab, delete the examination
* finish the import
-> the import is failed in tasks
-> a mail is sent to the study admin with the details

PR #1016, opened 2021-09-01
shanoir-issue#922 tags on shanoir subjects

jcomedouteau:
How to test:
Open a study you can modify.
There is now an option to create tags on the study. you can add new tags and delete them.
These tag can be applied/removed on subjectStudy
- From study edition
- From subject edition


jcomedouteau:
Not linked to solr yet

jcomedouteau:
Left to do: add tags in import subject creation context

michaelkain (commented):
Hi Jean-Come,
thank you for this pull request.
I have some questions about it:
- Have you written a specification on the GitHub wiki to describe this feature?
- --> very important for later users to know how to apply that feature
- Can you describe a typical example of how to use this feature (coming from Grenoble)?
- I am asking this, as potentially identifiable information could be entered using this way, and we need to carefully introduce this feature. Maybe put some explication and warn messages.
- As the Shanoir import is already quite complex, when no tags are defined for the study is there a way to somehow hide that feature to avoid that users will be confused?
- One important question is: are the tags unique on the study level, or maybe on the system/server instance level? We will have to discuss this with our users and the consortium. If an user defines interesting tags, do other studies want to reuse them?
- Have you noted a personal test report? so, what exactly you already tested yourself?
- Currently in SubjectStudyDTO.java you search tags for the subject + the tags for the study as availableTags. It could be more clean to have separate REST interfaces for both and that the frontend is working with both. The advantage: the REST API is open for outside usage, so we could have another system that populates the tags for a study using the REST API.
- In any case, this is an interesting extension, I will integrate into my presentation for the consortium, but I will better have to understand the "clinical" need for that.
- For the search: I prefer the duplication of the tags into ms datasets, same as for the subjectName. Tags are managed by ms studies. If a tag is added to a subject, what happens sometimes, we update ms datasets with this small information using rabbitmq. The indexation will remain fast like this as we can use native sql on the same database table. If ms datasets goes down: tags can still be added and managed, ms studies is the master. If ms studies goes down, ms datasets can still update the search index.
Michael

michaelkain:
Hi Jean-Come, can I have a look again and test? Or are you still working on it?

jcomedouteau:
I think I need to clean some little things, merge the conflict and ask Julien some advices for the front part.

jcomedouteau:
Combined with https://github.com/fli-iam/shanoir-ng/pull/1036

PR #1018, opened 2021-09-02
Shanoir-issue#970 Progress bar for upload + error when file too big

jcomedouteau:
1) Upload dicom file and see the progress
2) Upload a file with size  > 5Gb and see the error


Nunien:
@jcomedouteau 
Do you know what takes time after the uploading is 100% ? Is it the processing of it by the backend service ?

jcomedouteau:
Yes it's the processing that takes time. We could imagine a "Processing data..." message when the upload is at 100% ?

Nunien:
> Yes it's the processing that takes time. We could imagine a "Processing data..." message when the upload is at 100% ?

Yes, but it can be done in another development, this pr is an improvement.

Nunien:
> Can you check the error that comes back from the backend ?

Yes sorry it's 413 Request Entity Too Large

jcomedouteau:
I corrected it @Nunien

PR #1026, opened 2021-09-23
solr custom config

Nunien:
In order to be able to set a precise custom configuration in Solr (ie : configure the tokenisation to fix the "-" problem), the Solr config will no longer be set by annotations (which can't work in our architecture) but by givin to the solr docker a solr core config folder.

There are certainly useless files, I didn't delete them as I don't know if we could use them in the future. And they can't weight that much. But let's discuss it.

To test
* before delpoly, clean the solr-data volume
* check that searching for valyues with a - inside now work as expected


jcomedouteau:
maybe we can provide our own image

jcomedouteau (approved):
Ok for me, search with "-" works after solr reset

PR #1036, opened 2021-09-30
Tags - css updated

julien-louis:


jcomedouteau:
I have a compilation problem on this one, in SolrServiceImpl if you can correct it ? (saveAll instead of save l.84) @Nunien

jcomedouteau:
Just one little thing, can you call 'updateFormAndValidity' to display the update button correctly when updating a subject_study from study edition ?

jcomedouteau:
Something else to redo I think, if no studySubject of a study already has a tag, the "hasTags" is never true so you can never add one..
I think this has to be changed in subject-study-list-component.ts:
            this.hasTags = !!(this.model as SubjectStudy[]).find(subStu => subStu.study.tags && subStu.study.tags.length > 0);

PR #1037, opened 2021-10-01
Shanoir-issue#1033 EEG rework

jcomedouteau:
- Better download naming
- Database initialization
- Set updated metadata and sorting index for EEG datasets
- Consider EEG datasets in solr
- Clean EEGImportJob


shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/ImporterService.java
```diff
@@ -447,7 +450,7 @@ public void createEegDataset(final EegImportJob importJob) {
 				datasetToCreate.setCreationDate(LocalDate.now());
 				datasetToCreate.setDatasetAcquisition(datasetAcquisition);
 				datasetToCreate.setOriginMetadata(originMetadata);
-				//datasetToCreate.setStudyId(importJob.getStudyId());
+				datasetToCreate.setUpdatedMetadata(originMetadata);
```
Nunien, line 453: I feel like this can lead to errors. Doesn't it means that the two properties will point to the same object even in the database ? (I am not 100% certain of this because here originMetadata is a new entity and has no id)

Is there a reason why it couldn't be null here ? After all at this point there are no updated metadata for this dataset.

jcomedouteau: That's a good question. I agree with you that it's not a good idea to have both metadata to point on the same DB object.I think I did this initially to be able to use it to download dataset name, but it's not in use anymore. I'm removing this and the database update.

Issue #605, opened 2020-10-16
BIDS export with extra "session"


cmaumet:
Hi Shanoir-NG dev team!

It's very nice to see the BIDS export available, thanks for putting this together! 

I think there there is an extra session sub-folder that should not be present: 

<img width="385" alt="Screen Shot 2020-10-16 at 11 02 02" src="https://user-images.githubusercontent.com/5374264/96239214-b96bd280-0f9f-11eb-8800-22dc845d5ffc.png">
 
According to the [BIDS spec](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html): "Defining multiple sessions is appropriate when several identical or similar data acquisitions are planned and performed on all -or most- subjects, often in the case of some intervention between sessions (for example, training)". 

So, in most cases, the session sub-folder should be omitted. Session folders are usually only needed when we have multiple 'visits' (longitudinal studies) and in that case, the session names are common across subjects (or at least across a subset of the subjects).

jcomedouteau:
That's an interesting point, I hadn't read this !
It actually allows us to keep the examination level/logic from Shanoir, but you are right on the fact that it's not necessary on a majority of datasets.

We cannot really know (can we?) when we are in this case: "several identical or similar data acquisitions are planned and performed on all -or most- subjects" => So when in doubt we actually apply the rule more strictly and add the session level, that's how I see it, but it's clearly open to discussion :)

cmaumet:
Thanks for the prompt reply @jcomedouteau!

I completely understand the constraints on the development end and the idea of keeping the same logic for all studies. 

On the analysis end having the extra folder when not required, does add (a little bit) more complexity and I *think* that in most BIDS studies I have seen, the session folder is omitted as often as possible. 

But I think the main issue in the current setup is that we end up with different session names for each subject (which is not consistant with the [definition of BIDS sessions](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html) "Session - a logical grouping of neuroimaging and behavioral data **consistent across subjects**.").

Do you think that checking whether the sessions are available across subjects could help deciding whether or not we should have a session folder?

jcomedouteau:
> But I think the main issue in the current setup is that we end up with different session names for each subject (which is not consistant with the definition of BIDS sessions "Session - a logical grouping of neuroimaging and behavioral data consistent across subjects.").

[...]

PR #1042, opened 2021-10-08
Shanoir issue#1031 Anonymized subject birthDate error

### LINKED ISSUE #1031: create subject bug
* create subject page
* select already anonymized
* complete mandatory fields except date
* save
---> server error with the date and deserializtion

jcomedouteau:
How to test:
- Import EEG data
- Create a new subject 
- Already anonymized, do not add birthDate
- Save
- -> No error, birth date is set to 01/01/1970

Closes #1031


jcomedouteau:
Removing #605 commits

Issue #628, opened 2020-10-27
Edit subject: avoid change of common name


michaelkain:
To that, I think modifying the subject name via the edit menu should be prevented since the DICOM data themselves are not modified when modifying the subject names in Shanoir and this can be misleading when downloading DICOM data. All other information can be modified at less risk so I would say we can grant modification rights to those performing import.

PR #1043, opened 2021-10-08
Shanoir-issue#1031 Anonymized Subject birthDate save

### LINKED ISSUE #568: [front] Allow examination edition from list
Add little edition icon to examination list when available.

### LINKED ISSUE #627: As a complement / duplicate to #624 - do not allow modifying common/subject name, gender and examination date in the GUI after import
Modifying the common/subject name is the Shanoir GUI does not modify the DICOM patient name tags. This can be misleading when downloading the datasets with names mismatch between shanoir GUI and the DICOM files. This should not be allowed in the edit menu. Nor the gender field I would say, and the examination date, also taken from the DICOM tags. Other parameters can be modified/added (manual and language hemispheric dominance).

### LINKED ISSUE #1031: create subject bug

jcomedouteau:
How to test:

Import EEG data
Create a new subject
Already anonymized, do not add birthDate
Save
-> No error, birth date is set to 01/01/1970
Closes #1031
Closes #568
Closes #627

+ examination list => edit button in list is present when possible
+ Subject sex edition is now forbidden.


jcomedouteau:
Closes #568 
Closes #627 
Closes #628

PR #1059, opened 2021-10-19
ShanoirUploader: New Release: clean-up with profiles (no NG anymore, as all NG) + Apache HttpClient 5.1 + removes entire shanoir-old dependency

michaelkain:


michaelkain:
@jcomedouteau I have not yet asigned a reviewer, as I have not finished my tests :-)

michaelkain:
Fixes the ClassNotFound Exception with the DicomTreeNode from Shanoir-old. Strange: the library of shanoir-old (shanoir-1.0.0.jar) was part of the delivered new version, but the .jar seemed to be corrupted, even if it was the original one. Now solved: entire jar dependency removed and last link to shanoir-old cleaned and now completely free of shanoir-old.jar.

shanoir-uploader/src/main/java/org/shanoir/downloader/ShanoirDownloader.java
```diff
@@ -205,35 +205,33 @@ private void initProperties(final String fileName, final Properties properties)
...
 		final File downloadedFile = new File(destDir + "/" + fileName);
-
 		FileUtils.copyInputStreamToFile(response.getEntity().getContent(), downloadedFile);
 	}
 
 	public static String downloadDataset(File destDir, Long datasetId, String format,
 			ShanoirUploaderServiceClient shng) throws Exception {
 		System.out.println("Downloading dataset " + datasetId + "...");
-		HttpResponse response = shng.downloadDatasetById(datasetId, format);
+		CloseableHttpResponse response = shng.downloadDatasetById(datasetId, format);
```
jcomedouteau, line 219: surround executions with try {} catch {} the close the response in finally{} clause ? (I don't really know to which degree  a CloseableHttpResponse HAS to be closed everytime or not..)

michaelkain: Hi Jean-Come,
the main architecture principle here is to re-use the httpClient instance and not close it with each request.
From my point of view one instance of http client is responsible to manage its connections and take care
of all requests. This one instance shall only be closed, when ShUp shuts down. The only exception is the
proxyTest during application startup, where I do not know the target server yet (depends on profile selection).
In this case, I create and close a separate http client to check for the proxy connection.

BUT your comment had me have another look again on ShanoirUploaderServiceClient, and I missed some try statements, to release the connection, after the download-methods and fixed them, what should be more clean now.

With kind regards,
Michael

jcomedouteau (approved):

PR #1077, opened 2021-10-28
add support for the 'check-sso' keycloak authentication method

a-ba:
It is faster but may not work in some setups, so we let it configurable at runtime with the SHANOIR_KEYCLOAK_ADAPTER_MODE environment variable.


Nunien:
@a-ba What do you think should be tested here ?

a-ba:
it is rather a review which is needed (since it may affect security)

michaelkain:
Dear Anthony,
thank you for the above pull request.
I took my time to analyse it and from my point of view there is a misunderstanding and confusion of concepts in your pull request.
Please see below the documentation of Keycloak:

By default to authenticate you need to call the login function. However, there are two options available to make the adapter automatically authenticate. You can pass login-required or check-sso to the init function.
- **login-required** will authenticate the client if the user is logged-in to Keycloak or display the login page if not.
- **check-sso** will only authenticate the client if the user is already logged-in, if the user is not logged-in the browser will be **redirected back to the application and remain unauthenticated**.

From my point of view: in **no case**, we want check-sso for shanoir-ng-front. If the user is not authenticated, we explicitly want him to login to Keycloak for the client shanoir-ng-front. Same for the ohif-viewer, we **never** want to redirect back to the application and letting the user unauthentiated.

Your PR would be helpful and optimise things in case of check-sso, to use it silently to optimise redirects, right, but we do not require check-sso in the JavaScript Adapter, from my point of view.

Maybe I am wrong, but to discuss and with kind regards, Michael

a-ba:
The difference between the two methods is that `login-required`  performs the authentication directly in the browser window (by redirecting to the keycloak server) whereas `check-sso` performs it within an hidden iframe.

The redirection to the login page happens automatically with `login-required` because keycloak controls what's displayed in the browser window. In the case of `check-sso` keycloak cannot do the redirection (it would be pointless to display the login page in an hidden iframe), however the application can implement this redirection and that's exactly the purpose of https://github.com/fli-iam/shanoir-ng/pull/1077/files#diff-1c7b402055b6700cce6320a8d95db7f2493eb35e8049bdbf14795803ccb43a0dR78

michaelkain:
Hi Anthony,
no, the difference is, that both methods check the login-state using a hidden iframe, BUT login-required sends you to the login page and check-sso does it not do, but leaves you un-authenticated. That is why you use JavaScript to send back to the login-page yourself. What is the same behaviour as we had before. With kind regards, Michael

a-ba:
nope

demonstration:

- with 'check-sso', authentication is performed directly within the iframe, total time = 2.29s (for a basic page)
check-sso

- with 'login-required', the browser is redirected to the keycloak authentication page, authentication happens there, then keyclock redirects to the requested page, which is completely reloaded a second time (that includes reloading all assets, parsing, rendering and redoing the token validation with openid-connect), total time = 4.29s
login-required

michaelkain:
Dear Anthony, thank you for your patience with me and your now clear explication. I apologise for my comment above and I had not really understood the depth of your change and what you exactly optimised. Your JavaScript call optimises the redirect, I got it now. I will do one more local test and then directly approve your PR. With kind regards, Michael

a-ba:
actually i think the confusion came from the fact that the PR uses 'check-sso' + the silent check option (without the silent check, the redirect would be needed anyway)

michaelkain (approved):

PR #1093, opened 2021-11-15
Shanoir-issue#1092 Update form validity when updating entities

### LINKED ISSUE #1092: Form update is not well recalculated
When updating a study, removing a DUA file does not trigger the availability of update button.
This is the case on multiple entities of the application.

jcomedouteau:
- Update form validity (update button not grey anymore) when updating a study

Closes #1092


jcomedouteau:
Adding subject_study

Nunien (changes requested):
Imo it's not a good practice to use markAsDirty() inside ngOnChange(). I think dirty is when the user changes manually a form. But ngOnChange triggers on any change, even programmatic, for instance at the component initialization. The consequence here is that the form is always dirty when the page loads up.

 A quick solution here would be to call markAsDirty only when the user manually edit values. Typically it's the (change) event. But in my experience, on custom inputs, implementing (change) mess up with angular, that's why (change) may be called (userChange) or (onChange) on some custom inputs.
Exemple : <select-box (userChange)="parentFormGroup.markAsDirty()" ...>

jcomedouteau:
@Nunien this one is testable too

Nunien (approved):

Nunien:
This should also be implemented in MrDatasetComponent. Before merging we should decide if this should be done here or in another pr.

PR #1096, opened 2021-11-16
OHIF viewer 2.0 + DICOMWeb interface

michaelkain:
This PR will integrate a new viewer into sh-ng. As the version 3.0 is currently under development and does not yet seem to support OpenID connect with Keycloak (see user account control chapter empty), we will start with the v2.0, the React version. As the OHIF viewer requires a dicomWeb interface (-> studies), this PR will also integrate a studies-dicom-web implementation, that will be based on the today /examinations interface of MS datasets and return only exams visible by the current user account.


michaelkain:
As by version 3.0 of OHIF the _Embedded Viewer_ is **DEPRECATED**, this implementation already focuses on the stand-alone viewer installation.

michaelkain:
Hi Jean-Come,
I hope you had a nice weekend.
Could you please start to familiarise yourself with this PR?
I will try to finish it asap, maybe with your help during my holidays.
In the context of the RHU, they will have a group of experts for manual image lesion segmentations.
One solution could be to use the new OHIF-Viewer for that purpose, that allows manual annotations.
Maybe they could use Shanoir for that purpose, but therefore I will try to make asap a demo of the feature.
I have not yet implemented the feature of storing the manual annotations, neither finished the PR, but it
could be interesting to bring it out earlier. In case the annotations could be done on the raw dicom images
and the images do not require any special pre-processing (re-calibration etc.) before, to see.
With kind regards,
Michael

jcomedouteau:
Hi Michael! 
I'll have a look at the code and i'll test it this afternoon  to se how it's done :+1:

Jean-Côme

michaelkain:
Thank you a lot! And thank you for your BIDS development too! Great, that the feature is already seriously asked.

michaelkain:
Remove Google-API calls in entrypoint.sh, as the viewer will connect to Shanoir in any way

michaelkain:
Sorry, now my PRs contains the content of your PR, that I have tested on my branch, what was a mistake...

michaelkain:
I have added a separate entrypoint_common specific for ohif-viewer into docker-compose/ohif-viewer
the original entrypoint_common from shanoir-ng/utils folder is copied via Maven into docker-compose,
what does not work in the case of ohif-viewer, as we have no Maven project for it and I did not artificially
want to create one neither, just to copy the file, so I have some kind of code duplication in my small entrypoint_common
in ohif-viewer

michaelkain:
I have removed the Google-API calls in entrypoint.sh

michaelkain:
I reverted your PR, it should be clean now. I let you take over from here, if okay. Cheers, Michael

michaelkain:
READY for final testing, and prod!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
```diff
@@ -0,0 +1,174 @@
...
+	 * of examinations in Shanoir, in the Json returned.
+	 * 
+	 * @param root
+	 * @param examinationId
+	 * @param studyLevel
+	 */
+	public void replaceStudyInstanceUIDsWithExaminationIds(JsonNode root, Long examinationId, boolean studyLevel) {
+		if (root.isObject()) {
+			Iterator<String> fieldNames = root.fieldNames();
+			while (fieldNames.hasNext()) {
```
jcomedouteau, line 57: Why iterating over all elements if its possible to get elements directly with root.get("0020000D");
Extract constants for these ? 

Also, 

> for (int i = 0; i < studyInstanceUIDArray.size(); i++) {
>       studyInstanceUIDArray.remove(i);
>       studyInstanceUIDArray.add(examinationId.toString());
> }
does not lead to some kind of concurrent modification ?
(isn't it risky to iterate based on the lenth of an "array" from which you add and remove an element ?)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
```diff
@@ -0,0 +1,174 @@
...
+		if (!acquisitions.isEmpty()) {
+			DatasetAcquisition acquisition = acquisitions.get(0);
+			List<Dataset> datasets = acquisition.getDatasets();
+			if (!datasets.isEmpty()) {
+				Dataset dataset = datasets.get(0);
+				List<DatasetExpression> expressions = dataset.getDatasetExpressions();
+				if (!expressions.isEmpty()) {
+					for (DatasetExpression expression : expressions) {
+						// only DICOM is of interest here
+						if (expression.getDatasetExpressionFormat().equals(DatasetExpressionFormat.DICOM)) {
```
jcomedouteau, line 138: If we have an examination with an EEG acquisition as first acquisition, we won't retrieve anything. I think we need to iterated over acquisitions / datasets until we eventually find one containing DICOMs.

Maybe add a check on datasetFile.isPacs() to be sure that we are retrieving a dicom path from the PACS and not from the file system ? (and iterate over files too in this way)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
```diff
@@ -0,0 +1,174 @@
...
+public class StudyInstanceUIDHandler {
+	
+	private static final Logger LOG = LoggerFactory.getLogger(StudyInstanceUIDHandler.class);
+
+	private static final String WADO_URI_STUDY_UID_SERIES_UID = "studyUID=(.*?)\\&seriesUID";
+	
+	private static final String WADO_RS_STUDY_UID_SERIES_UID = "/studies/(.*?)/series/";
+
+	@Autowired
+	private ExaminationService examinationService;
```
jcomedouteau, line 36: I can see no method where this cache is evicted/updated.
If a user deletes/updates an examination (or the data behind it) it should be updated/evicted no ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/dto/mapper/ExaminationToStudyDTOMapper.java
```diff
@@ -0,0 +1,112 @@
...
+	protected SubjectRepository subjectRepository;
+	
+	public StudyDTO examinationToStudyDTO(Examination examination) {
+		StudyDTO studyDTO = new StudyDTO();
+		final String studyInstanceUID = UIDGeneration.ROOT + ".1." + examination.getId();
+		// 5 DICOM-study specific values
+		studyDTO.setStudyInstanceUID(studyInstanceUID);
+		studyDTO.setStudyID(examination.getId());
+		studyDTO.setStudyDescription(examination.getComment());
+//		studyDTO.setStudyDate(examination.getExaminationDate().toString());
```
jcomedouteau, line 40: Why don't we keep exam date here ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/dto/mapper/ExaminationToStudyDTOMapper.java
```diff
@@ -0,0 +1,112 @@
...
+			serie.setSerieInstanceUID(serieInstanceUID);
+			serie.setSeriesNumber(datasetAcquisition.getSortingIndex());
+			serie.setModality(datasetAcquisition.getType().toUpperCase());
+			
+			List<InstanceDTO> instances = new ArrayList<InstanceDTO>();
+			List<Dataset> datasets = datasetAcquisition.getDatasets();
+			for (Dataset dataset : datasets) {
+				InstanceDTO instance = new InstanceDTO();
+				MetadataDTO metadata = new MetadataDTO();
+				metadata.setSopClassUID("1.2.840.10008.5.1.4.1.1.2");
```
jcomedouteau, line 85: Is it normal that this value is fixed ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
```diff
@@ -121,6 +133,21 @@ public void deleteFromRabbit(Examination exam) throws EntityNotFoundException {
 			return examinationRepository.findByPreclinicalAndStudyIdIn(preclinical, studyIds, pageable);
 		}
 	}
+	
+	@Override
+	public Page<Examination> findPage(final Pageable pageable, String patientName) {
+		if (KeycloakUtil.getTokenRoles().contains("ROLE_ADMIN")) {
+			if (StringUtils.isNotEmpty(patientName) && patientName.length() <= 64) {
```
jcomedouteau, line 140: Why this 64 char limit ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
```diff
@@ -121,6 +133,21 @@ public void deleteFromRabbit(Examination exam) throws EntityNotFoundException {
...
+				return examinationRepository.findAllBySubjectName("%" + patientName + "%", pageable);
+			} else {
+				return examinationRepository.findAll(pageable);
+			}
+		} else {
+			Long userId = KeycloakUtil.getTokenUserId();
```
jcomedouteau, line 146: So if the user is not an admin, he can't search a subject name ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
jcomedouteau: We actually have many weird things in the DB (we have an issue for this): 
select count(*) from dataset_file df JOIN dataset_expression de ON df.dataset_expression_id = de.id WHERE de.dataset_expression_format = 6 and df.pacs is false;
+----------+
| count(*) |
+----------+
|    20414 |
+----------+

So the check on "pacs == true" should totally be done :)

jcomedouteau:
When I try to start the app by doing a .:bootstrap.sh --clean I have the following exception:

shanoir-ng-nginx            | sed: can't read /etc/nginx/html/assets/env.js: No such file or directory

i'm trying to  have a look on what's the problem :)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
michaelkain: Hi Jean-Come, please find my answers below:
- root.get("0020000D") does only give you access to a direct child element, not the entire tree, what would be too magic; as 0020000D is a nested element I have to go recursively
- constants extracted, DONE
- for the array, I changed to insert(i,, but please see here: https://stackoverflow.com/questions/31582252/how-to-change-element-value-in-arraynode, but as I found other code examples, it seems to be the way to do

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
michaelkain: - EEG acquisitions are not DICOM and can not be displayed by OHIF-viewer, normally
- great, THX!, I have added the isPacs() check to be protected

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
michaelkain: - if an user deletes an examination, no display of images can be called on this exam via the GUI
- for me, not an issue if the cache until the night would provide an UID, that is never called
- update exam, does not change the dataset files normally, even if deleted, but the UID remains unique for all series of one exam

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/dto/mapper/ExaminationToStudyDTOMapper.java
michaelkain: - as stated in my JavaDoc please ignore DICOMJson, as it is only a draft, that might be interesting for v3 of OHIF-viewer, the issue here is: the shanoir exam date format does not corrspond to DICOM date format

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/dto/mapper/ExaminationToStudyDTOMapper.java
michaelkain: no, but please as stated in my JavaDoc please ignore DICOMJson

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
michaelkain: 64 is the max length for DICOM patient name, see DICOM spec

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
michaelkain: I have to fix that, thank you, I forgot about it

michaelkain:
.gitignore hided the env.js, now I have commited it

docker-compose/ohif-viewer/entrypoint_common
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
michaelkain: I agree, this was just a commit error, not a conceptional one, I am removing the files. I am wondering, why the .gitignore does not seem to work every time..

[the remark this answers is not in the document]
michaelkain: remaining: ./utils/entrypoint_common

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
michaelkain: fixed

michaelkain: just for the notes: I tried a cleaner solution on not-using @query (native query) and changed the type of Long subjectId to Subject subject and I could avoid the usage of @query, BUT this introduced other issues afterwards with Gson in the Unittests, that I was not able to solve within 1 hour, so I stopped to one day finish this PR. The problem lies, that Gson has even be introduced, what should not have been done. Gson annotations are different to the Jackson ones, and to fix my issue, I would have to double annotate Java classes for json with both annotation types, what would be even more confusing than anything else. One solution would be to clean up all Gson usage in the Unit tests, what is quite a task.

docker-compose/nginx/https.conf
```diff
@@ -17,3 +17,42 @@ server {
...
+		proxy_pass	http://SHANOIR_PREFIXkeycloak:8080/auth/;
+		# protect admin web interface from outside access
+		location = /auth/admin/ {
+			return 404;
+		}
+	}
+
+	location / { proxy_pass http://SHANOIR_PREFIXohif-viewer:9905/; }
+	# Directly root to MS-Datasets, to avoid CORS-errors from OHIF Viewer
+	location /dicomweb/	{ proxy_pass http://SHANOIR_PREFIXdatasets:9904/dicomweb/; }
```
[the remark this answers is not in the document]
michaelkain: no, I removed the line from shanoir.template.conf, that must be some legacy code, maybe used for image preview during the import in an older version, no link with real "DICOMWeb"

docker-compose/nginx/https.conf
```diff
@@ -17,3 +17,42 @@ server {
...
+	
+	proxy_set_header	Host $http_host;
+	proxy_set_header	X-Forwarded-For $remote_addr;
+	proxy_set_header	X-Forwarded-Proto $scheme;
+
+	#
+	# Keycloak
+	#
+	# forward (reverse-proxy) all requests for /auth/* to the Keycloak server
+	location /auth/ {
```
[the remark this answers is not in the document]
michaelkain: you are right, I removed it, this was a legacy code of my first tries with OIDC and the viewer

docker-compose/nginx/https.conf
```diff
@@ -17,3 +17,29 @@ server {
 
 	include shanoir.conf;
 }
+
+server {
```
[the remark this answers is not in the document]
michaelkain: done, see my commit

README.md
```diff
@@ -179,6 +179,8 @@ Name                  | Value             | Description
 `SHANOIR_CERTIFICATE` | `auto\|manual`     | auto-generates a self-signed TLS certificate (NOT RECOMMENDED) or use a manually installed certificate |
 `SHANOIR_MIGRATION`   | `auto\|init\|never\|manual\|export\|import` | Normal runs should use `auto` in development and `never` in production. Other values are for controlling deployment and migrations (see below). |
 `SHANOIR_KEYCLOAK_USER`<br>`SHANOIR_KEYCLOAK_PASSWORD` | *username/password* | Keycloak admin account used by shanoir for managing user accounts |
+`VIEWER_URL_SCHEME`  | `http\|https`      | https (over TLS), http (plain text, NOT RECOMMENDED) |
```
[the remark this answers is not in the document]
michaelkain: I renamed both variables to start with SHANOIR, see my commit

docker-compose/datasets/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN apt-get update -qq \
 # Installation of dcm4che to manage the store-scu into the PACS
 # dcm4che3, used last version 5.21.0 as available on the 2020-02-14
 # https://sourceforge.net/projects/dcm4che/files/dcm4che3/
-RUN wget -O dcm4che-5.21.0-bin.zip https://downloads.sourceforge.net/project/dcm4che/dcm4che3/5.21.0/dcm4che-5.21.0-bin.zip
+RUN wget --no-check-certificate -O dcm4che-5.21.0-bin.zip https://downloads.sourceforge.net/project/dcm4che/dcm4che3/5.21.0/dcm4che-5.21.0-bin.zip
```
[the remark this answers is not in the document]
michaelkain: done

michaelkain:
Dear Anthony, normally I have solved all your comments, at least I tried my best. Could you please verify and resolve the conversations? Thank you and with kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
michaelkain: Hi Jean-Come, would you be okay to close this conversation? I think there is really no value from cleaning the cache, except cleaning it via a scheduler e.g. every 24 hours or once a week. It only introduces more code complexity. Our prod and it will be the same for the GME is stopped for security updates all night, so the cache is cleaned in the morning. The info in the cache is so small, even if during one morning somebody we read 500 exams, nothing would impact that on the server. Exams are in practice never deleted once imported. And nobody except typing the URL can open a deleted exam into the viewer and examIds are counted up all the time. The viewer would just throw an error.

michaelkain:
Hi Jean-Come, for me it is fine. Could you please have another look?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
jcomedouteau: If the cache is evicted every 24h it's OK for me. let's keep this in mind if we have a bug with the OHIF viewer one of these day ;)
I'm closing it!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
jcomedouteau: > root.get("0020000D") does only give you access to a direct child element, not the entire tree, what would be too magic; as 0020000D is a nested element I have to go recursively

But you can test for every element
`if (root.get("0020000D") != null) { updateNode();}
`
And keep the recursive method to check at every level of the tree if  this node exists.
If the JSON is not that big, it's not really relevant, otherwise, it avoids us to iterate over all the elements of the tree.

Furthermore, if this change need to be done only in these two fields and these fields are present only once in the whole tree, you can stop the recursion with an adapted condition.

But once more I don't know if this is relevant as I never saw the structure of the json dicomWeb.

docker-compose/nginx/http.conf
```diff
@@ -5,3 +5,24 @@ server {
...
+	proxy_read_timeout 	5000s;
+
+	# increase the buffer size (the default 4kb is occasionally too small for keycloak)
+	proxy_buffer_size	8k;
+	
+	# do not buffer whole responses
+	proxy_buffering 	off;
+	
+	proxy_set_header	Host $http_host;
+	proxy_set_header	X-Forwarded-For $remote_addr;
```
a-ba, line 22: `X-Forwarded-For` and `X-Forwarded-Proto` must be generated only if `SHANOIR_X_FORWARDED=generate`
(in production they are generated by the frontal nginx server)

I fixed it + deduplicated the ohif-viewer config + simplified the entrypoint in michaelkain/shanoir-ng#8

a-ba (changes requested):

docker-compose/nginx/http.conf
michaelkain: Great Anthony, thank you for your help here, what makes it much clearer!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/StudyInstanceUIDHandler.java
michaelkain: Hi Jean-Come,
studyInstanceUIDs and retrieveURLs are multiple elements, that occur often in the json, so I can not stop after having replaced one element. I have no guarantee at which position they come and in which order. So I am not sure to understand, what you mean. Maybe, but not very probably, the depth might change too. And when I found the element I am not going deeper in the tree, as both elements are on the same depths. Maybe a screenshot helps here.
With kind regards,
Michael
Bildschirmfoto 2022-03-30 um 10 49 25

michaelkain:
Hi Jean-Come, as follow up to your comment: I clean now the DICOMWeb cache every morning at 6:00:00.

michaelkain:
For me the PR is ready for PROD!

michaelkain:
Hi Jean-Come, as discussed I have commited the json parsing optimisation proposed by you. Hope it is fine now?

jcomedouteau (approved):
Ok for me !

michaelkain:
As Jean-Come has approved my PR, Anthony closed all conversations and I merged the PR from Anthony, I consider we are fine and I merge the PR. In the worst case, the button "View DICOM" on exam details will not work, but at least this big PR is integrated.

PR #1099, opened 2021-11-16
Shanoir-issue#1087 Simplify SQL scripts

### LINKED ISSUE #1087: [SQL] Clean initialization SQL scripts
Datasets initialized are not pertinent anymore, remove them.
- Create a demo study with a center / equipement and a study card, and that's all.

jcomedouteau:
How to test:
- Deploy from scratch (or set the argument to 'true' in application.yml for each microservice) and check the logic.
ddl-auto: create
Closes #1087


Nunien (approved):
I used bootstrap.sh and I had no user created in Keycloak, I had to do it manually. I don't remember if it's a normal behaviour but anyway I assume it's independent from the current pr.

PR #1100, opened 2021-11-17
Shanoir issue#1070 subject name limitation + download

### LINKED ISSUE #1070: Bug when dataset names contain forward slash /
Datasets can have names with  forward slash (/) which are interpreted as subfolder when creating the zip files during the download process.

Examples : https://shanoir-ofsep.irisa.fr/shanoir-ng/dataset/details/227619
This file is downloaded as 3D AX_dcm.zip

Special characters should be escaped before creating Files, and special characters should be forbidden in dataset, subject and other names.

jcomedouteau:
How to test:
- Before launching test, import data, create a subject with a name with a slash.
--> Data are not accessible anymore
- After deploy
--> Dataset download is now possible
--> It is now impossible to create a new subject with a name with a '/' in it. (and other specific chars)

Closes #1070

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
@@ -327,6 +327,9 @@ public class DatasetApiController implements DatasetApi {
 			if (subjectOpt.isPresent()) {
 				subjectName = subjectOpt.get().getName();
 			}
+			if (subjectName.contains("/")) {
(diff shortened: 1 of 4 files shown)


michaelkain:
Hi Arthur, would it be okay for you to test it, as you will use it from your script? We could then directly merge it if it works with your tests? Cheers and thx, Michael

arthursw (approved):
Ok, everything works fine :
 - it is now impossible to create a subject with a slash in the name
 - it is now possible to download datasets with subject with slash in the name

However, I could create a subject with the name "é"'(§è!çà)-*`%^¨.,=", I don't know if that makes sens...

PR #1104, opened 2021-11-19
Tag fixes

Nunien:
*  study edition
    * the tag list updates when adding a tag in study edition
    * new subject studies using new tags are saved in both studies & datasets
        * tag indexing works here (there has to be datasets for this subject)
    * no tag with same color and name in a study
* fix circular dependencies warns
* fix studycard msg when saving study


jcomedouteau:
Hey Julien ! Can you please merge the conflicts ? Thanks !

jcomedouteau:
Some points I met:
 When I try to delete a newly created tag that was associated with a subject study during the edition, I have some mixed comportements:
- I can delete one even if it is linked to a subject study -> It disappear, but when I recreate one with the same name, it re-appear with the other's color.
- I can't delete the first one that is no more linked to any subject.

Solr Indexing = OK
no tag with same color and name in a study = OK (we accept same color or same name ?)
fix circular dependencies warns = OK
fix studycard msg when saving study = To be merged

Nunien:
@jcomedouteau I think it's fixed

jcomedouteau:
It works better ! :)

I think I have another problem that just appeared:
If we remove all the tags from subject studies and delete this same tag on the same study edition, I have the following problem:
studies                    
| Caused by: java.sql.SQLIntegrityConstraintViolationException: Cannot delete or update a parent row: a foreign key constraint fails (`studies`.`subject_study_tag`, CONSTRAINT `FKcc51xo3yrp74v2yp7df540a03` FOREIGN KEY (`tags_id`) REFERENCES `tag` (`id`))
 
I think we have to delete the subject_study_tag reference then the tag.

jcomedouteau:
@Nunien Could you reproduce my last problem ?

Nunien:
@jcomedouteau Absolutely, a solution that I tried to implement is to cascade the relationship subject_study_tag (not the subject_study) when deleting a trag.

Nunien:
@jcomedouteau I've done it ! 
Also I must say that I have the feeling that it's not the way hibernate is supposed to work.

jcomedouteau:
I currently can't create a subject during the import. feels like the subject creator fails here:
`this.subjectStudyTags = e.tags.map(e=>new xM(new c_(this.id),e))`

This PR seems doomed haha

Nunien:
@jcomedouteau To me it's working now

Nunien:
I'm merging so we will test it at 10

PR #1105, opened 2021-11-19
Shanoir-issue#77 Delete from PACS - delete dataset one by one w REST

### LINKED ISSUE #77: Delete subject: does dcm4che archive 5 support deletion
1. Does dcm4che archive 5 provide (or DICOM protocol in general == DIMSE) a remote delete call? I found N-DELETE, but not sure if it is the right action to use. When we delete a subject in sh-ng we could using this call automatically delete the data in the PACS.
2. Does the web interface of da5 provide a delete button, as dcm4chee2 (used today), to delete patients? If 1. does not work to clean up afterwards with the PACS web interface?

jcomedouteau:
How to test:
- Import some new data
- Check that the data appears in the PACS (http://localhost:8081/dcm4chee-arc/ui2/#/studies)
Delete a newly created dataset:
BEFORE:
- Dataset is still on the local PACS
AFTER
- Dataset is not more on the local PACS

What to test:
- delete datasets one by one
- delete acquisition
- delete examination
- delete full subject

For the laters tests, subject and studies may stay on the PACS, BUT the images should be all deleted.

Closes #77


jcomedouteau:
Merde is OK, it is ready for review.

jcomedouteau:
Yes that's totally true, i didn't think that the path would actually change in the dataset_file.
I need to update this

michaelkain:
Please see here: DicomDatasetExpressionStrategy, where both paths are created, an idea would be to check the link for the char sequence "?requestType=WADO" (== WADO-URI) and WADO-RS as otherwise, but other methods could be used

michaelkain:
Thank you!

michaelkain (changes requested):
Dear Jean-Come,

thank you for this important PR.

I would please require some changes, as the components are not very clearly named now.
We can use two protocols to communicate with a Pacs: DIMSE (old, e.g. c-store) and DICOMWeb == REST (new, e.g. stow-rs).
This has been reflected in the former architecture, see develop today with the interface DicomServiceApi.
To add to a Pacs, we can either use the CStoreDicomService or the StowRsDicomService.
Only the new REST protocol provides a deletion, not the DIMSE protocol, as far as I know.
What I propose please:
- Rename CStore to DIMSEService and declare the deleteDicomFilesFromPacs as NOT available
- Rename StowRsDicomService to DICOMWebService and directly implement the deletion
- Remove your GenericDicomService, that leads to understanding confusion for outside people
- You always delete with DICOMWeb, nevertheless if extracting the info from WADO-URI or WADO-RS links

I will then test on my machine.

Michael

jcomedouteau:
Thanks for the feedback, this new service class was actually stupid..
I'm changing this !

Jean-Côme

jcomedouteau:
I corrected it but I currently have a problem with deletion in hibernate. I'm checking it.

michaelkain (commented):
Code review is perfect now, I will have to test locally.

michaelkain:
- Updated on develop + pull from PR #1105
- ./[bootstrap.sh](http://bootstrap.sh/) --clean
- in ms datasets flag dcm4che-arc.dicom.web remains false
- new imported data are added to the db in WADO-URI and c-store is used to send to the backup pacs
- User mkain, admin
- Import of sample.zip dicom, one serie selected, no2
	- New SC
	- New subject 2022-05-11-01
	- New exam
	- OK, worked
- 96 images in Pacs, use web-gui-dcm4chee arc
- Delete entire exam with id-1, using exam details delete
**- KO: images remain in pacs using web gui of dcm4chee5**
	- web-gui: exam removed and all datasets deleted

Dear Jean-Come,
sorry, but my local test do not work. I deleted the exam and the images remained.
Michael

jcomedouteau:
Did you change your application.yml to dicomweb = true ?

Anyway, that's really weird, I have the same comportment locally. I'll check what's happening.

michaelkain:
Hi Jean-Come,

I changed to dicom.web = true and it WORKS! Sorry for my mistake!
- Import, new exam created, exam deleted: study images removed from pacs -> **OK!**
- Import with entire sample.zip, creates 8 datasets, 6 series in pacs, serie no6 deleted == dataset no8, OK in sh-ng-GUI, OK in pacs-GUI, last serie removed with all its images -> **OK!**
- Import with entire sample1.zip (quite big), new subject + new exam
**- KO: when I try to delete a subject an error is thrown:** Uncaught (in promise): Error: [object Object] e/this.**openDeleteConfirmDialog**/
- not sure, if this error is known and fixed elsewhere
- and there is a conflict now

With kind regards,
Michael

jcomedouteau:
It should be better now.

michaelkain:
Hi Jean-Come,
thank you for that great optimisation. I tested the subject deletion, with a really big exam (sample1.zip)
and it takes a very long time for deletion in the pacs itself, what is quite surprising for 1000 images only.
BUT it works fine now. I will approve and merge, as nothing should change in prod for dicom.web = false.
With kind regards,
Michael

michaelkain (approved):
Great job!

PR #1112, opened 2021-11-24
Tasks table auto refresh

Nunien:
* check that the front doesn't make unnecessary tasks api calls


jcomedouteau (approved):

jcomedouteau:
Very cool ! Ok for me

PR #1114, opened 2021-11-26
Papaya fullscreen

Nunien:
Papaya is used in : 
* dataset details
* select series
* bruker select series

The pr enables the papaya menu and the fullscreen mode


jcomedouteau:
This is very cool !
I do have a small problem, when you put papaya in full screen, you cannot really come back to small screen.

Nunien:
@jcomedouteau where is that ? I deployed the pr to qualif and I can't reproduce the bug. Can you check on qualif ?

jcomedouteau (approved):
Ok for me. When old papaya.js is in cache the bug appears, but it works on qualif.

Issue #1084, opened 2021-10-29
Add concerned study and contact person in manage user menu


ebannier:
The challenge and FLI-GME in waiting studies hosted at Neurinfo bring a lot of account requests that are not administrated by Neurinfo. Would it be possible to add in the manage user menu columns to identify the concerned study (should be specified at account request) and contact person so that we can quickly pick up those that are on our plate ?

Nunien:
I agree, there could be a "Studies" column in the user list.

PR #1117, opened 2021-11-30
Studies in user table

Nunien:
* study list column in user table - https://github.com/fli-iam/shanoir-ng/issues/1084
* fix bug : in another user details, click on profile -> no change
* study-user list separate component
* change study list style in user


jcomedouteau:
@Nunien Theres (already ?) conflict to resolve !
Can you describe this bug -> fix bug : in another user details, click on profile -> no change ?
I don't understand it well.
Thanks!

shanoir-ng-front/src/app/studies/shared/study-user.model.ts
```diff
@@ -32,11 +33,7 @@ export class StudyUser {
...
 
     public static completeMember(studyUser: StudyUser, users: User[]) {
-        for (let user of users) {
-            if (studyUser.userId == user.id) {
-                studyUser.user = user;
-                user.selected = true;
-            }
-        }
+        let user: User = users.find(u => u.id == studyUser.userId);
+        if (user) studyUser.user = user;
```
jcomedouteau, line 37: The user.selected = true is not usefull anymore ?

jcomedouteau:
J'ai eu des soucis avec le nouveau composant study-user:
1) Mon utilisateur jlouis apparait comme faisant parti de l'étude ET dans les options d'ajout (ce qui n'est pas le cas des autres utilisateurs de l'étude)
image
Si j'ajoute de nouveaux utilisateurs, ils ne disparaissent pas de la liste.

2) Si je supprime un utilisateur existant (ici mkain), j'ai une erreur et l'utilisateur n’apparaît pas dans les options:
image

Merci !

Nunien:
@jcomedouteau Sorry I can't reproduce that. I suspect there might be a problem with your users data. All the user users should apprear in the select box and the selected ones should be greyed. Is the response from https://shanoir-ng-nginx/shanoir-ng/users/users [GET] normal ? Does every user have an id ?

jcomedouteau:
I'm re-testing !

jcomedouteau:
You are right, my users are not well initialized in keycloak. i'm updating this.

jcomedouteau:
I just had the little problem that when you re-edit the study, all available users (even if they already are in the study) are displayed as "addable". It does nothing when clicking on "add", but they are not grey.
image

Nunien:
> 

Yes, I think an issue should be created as the problem existed before and it would require a new api service or a rabbitmq transaction to deal with it.

Nunien:
> It's kind of weird, when deleting an user, we don't delete associated study_user object -> This should be done no ?

Yes, I think an issue should be created as the problem existed before and it would require a new api service or a rabbitmq transaction to deal with it.

Nunien:
> I just had the little problem that when you re-edit the study, all available users (even if they already are in the study) are displayed as "addable". It does nothing when clicking on "add", but they are not grey.

I can't reproduce that

jcomedouteau (approved):
Ok could not reproduce on qualif my last bug. I'm creating a new issue for the other one.

PR #1124, opened 2021-12-02
Shanoir-issue#1108 Disable download button in study tree when already downloading

### LINKED ISSUE #1108: [Download] Import from study tree, disable button
When importing from study tree (download all datasets) -> Disable download buttons from the study tree.

jcomedouteau:
When downloading from study tree, disable current button
- dataset level
- examination level

@Nunien I used a callback to be able to do so, but I'm not 100% convinced.
I could not manage to return an Observable from the subscribe method, but there may be a way that I ignore, if you want to have a look :)
Thanks !

Closes #1108


shanoir-ng-front/src/app/datasets/shared/dataset.service.ts
```diff
@@ -81,10 +81,13 @@ export class DatasetService extends EntityService<Dataset> implements OnDestroy
...
                 progressBar.progress = 0;
                 saveAs(event.body, this.getFilename(event));
+                if (callback) {
+                    callback();
+                }
         }
     }
 
-    public downloadDatasets(ids: number[], format: string, progressBar: LoadingBarComponent) {
+    public downloadDatasets(ids: number[], format: string, progressBar: LoadingBarComponent, callback?) {
```
Nunien, line 90: I think you should use a Promise here. Actually promise are essentially callbacks but with a "cleaner" syntax. There might be a difficulty here considering we deal with (righfuly) an observable. But after reading the doc, toPromise() triggers on the completion of the observable. (complete event can be used as the 3rd argument in subscribe(), after error => {}). So I would do that : 

```
public downloadDatasets(ids: number[], format: string, progressBar: LoadingBarComponent): Promise<HttpEvent<any>> {
        const formData: FormData = new FormData();
        formData.set('datasetIds', ids.join(","));
        formData.set("format", format);
        
        let postResponse: Observable<HttpEvent<any>> = this.http.post(
            AppUtils.BACKEND_API_DATASET_URL + '/massiveDownload', formData, {
                reportProgress: true,
                observe: 'events',
                responseType: 'blob'
        });
        this.subscribtions.push(
            postResponse.subscribe(
                (event: HttpEvent<any>) => this.progressBarFunc(event, progressBar),
                error =>  {
                    this.errorService. handleError(error);
                    progressBar.progress = 0;
                }
            )
        );
        return postResponse.toPromise();
    }
```

shanoir-ng-front/src/app/examinations/tree/examination-node.component.ts
```diff
@@ -126,23 +127,30 @@ export class ExaminationNodeComponent implements OnChanges {
...
+                this.downloading = false;
                 return;
             }
         } else {
             datasetIdsReady = Promise.resolve();
         }
         datasetIdsReady.then(() => {
-            this.datasetService.downloadDatasets(this.datasetIds, format, this.progressBar);
+            var self = this;
+            this.datasetService.downloadDatasets(this.datasetIds, format, this.progressBar, function() {
```
Nunien, line 148: then you can do that : 
```
datasetIdsReady.then(() => {
            this.datasetService.downloadDatasets(this.datasetIds, format, this.progressBar)
                .then(() => {
                    self.downlading = false;
                }).catch(() => {
                    self.downlading = false;
                });
        });
```

jcomedouteau: Thanks, I'll have a look

Nunien (approved):

Issue #1128, opened 2021-12-06
Solr search inconsistencies


mgaubert:
Hello,

I found two inconsistencies in Solr, when displaying results. 
First, when sorting a list of images per subject ID, there is one subject who is not where he should be in the list, see below (subject MS-TRACTS 11009 in blue is not organized correctly):

![Pb_1_Shanoir](https://user-images.githubusercontent.com/93723496/144812958-3298c114-2a78-4523-9b50-5b45970b91d0.png)

Second, I can't select multiple subjects, nothing is returned while there should be results. For example, before selecting the subjects (one image per subjects sould be available):

![Pb_2 1_Shanoir](https://user-images.githubusercontent.com/93723496/144813052-4701bc08-7a45-433c-b69b-072cbb613604.png)

And after two subjects are selected, nothing returned:

![Pb_2 2_Shanoir](https://user-images.githubusercontent.com/93723496/144813075-89a29c31-2d0f-4201-857d-a3acae3f57ff.png)

When selecting one subject, one image is returned as it should be.

Thank you for your answers.

Malo

jcomedouteau:
Hello Malo,

Thank you for your feedback !
Concerning the second part of your message, I think the reason is that we choose to make an "AND" between the subjects using the filters on the right of the screen.
If you want to use an "OR" operand, you may want to use the search bar on the top of the screen.
For the first part, I'll have a look.

Have a nice day,
Jean-Côme Douteau

mgaubert:
Hello Jean-Côme,

Thank you for your answer.
I may have missed something but I was not able to use the top search bar appropriately. I mean, I entered the following: 

_subjectName:(MSTRACTS_01009 OR MSTRACTS_01010)_

without any other filters and with lower or upper case for "MSTRACTS".

And got nothing, while there should be >100 results (~80 images per subject).

Did I enter something wrong here?

Have a nice day too,
Malo

jcomedouteau:
Malo,

Your request is correct !
To use the "subjecName" criteria, you have to use the "Expert mode"
![image](https://user-images.githubusercontent.com/7604176/144994109-598f8900-ea83-4151-a80f-f4f9156b045d.png)

The search bar as "beginner" is made just to type key words, or simple names. Expert mode allows to make more complex requests with AND/OR/NOT operators and criterias such as subjectName.

Jean-Côme

mgaubert:
[...]

PR #1131, opened 2021-12-07
solr search no results with multiple values

Nunien:


Nunien:
Fix an important problem with the Solr search. Reported in https://github.com/fli-iam/shanoir-ng/issues/1128

jcomedouteau:
Can you include the column sorting ? (removing the "-str")
thanks !

jcomedouteau (approved):

PR #1142, opened 2021-12-10
added submodule shanoir-downloader 

arthursw:
from https://github.com/Inria-Empenn/shanoir_downloader.git


arthursw:
This is an example of how we could integrate [shanoir_downloader](https://github.com/Inria-Empenn/shanoir_downloader.git) into the main Shanoir repository : as a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules).

When we clone the shanoir-ng repository, the shanoir-downloader/ folder will be empty ; two commands must be run to get the code:
 - `git submodule init`  to initialize your local configuration file
 - `git submodule update`to fetch all the data from shanoir and check out the appropriate commit listed in shanoir_downloader

Then the shanoir-downloader project can be simply managed as a normal git repo (as if it were a separated project) ; meaning that once your are in the shanoir-downloader/ folder, you can just `git pull` to get the latest changes, and commit some changes.

jcomedouteau:
That's quite cool; I didn't knew it existed !
I think if we use this, we have to update the readme, or add a readme on how to use this is the empty folder :)

jcomedouteau (approved):

michaelkain:
THX to both of you for this nice work!

PR #1155, opened 2021-12-15
add missing dependencies

a-ba:
When building the production images this file is updated (to add these missing deps), which makes changes in git's working tree, which makes git-describe add the `-dirty` suffix to the version number.

### DIFF
--- a/shanoir-ng-front/package-lock.json
+++ b/shanoir-ng-front/package-lock.json
@@ -2966,11 +2966,6 @@
         "color-convert": "^1.9.0"
       }
     },
-    "ansi_up": {
-      "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/ansi_up/-/ansi_up-5.0.0.tgz",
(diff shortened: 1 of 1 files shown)


Nunien:
Hi @a-ba, I want to clarify something. package-lock.json is generated : https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json. But, as the doc says : "This file is intended to be committed into source repositories". However I believe it's recommended not to edit it manually but to let npm update it with CLI commands like "install".

The question I'm asking here is : shouldn't package.json be updated as well ? What do you think ?

a-ba:
> it's recommended not to edit it manually but to let npm update it with CLI commands like "install".

it is not a manual edition, these changes are done by npm during the build

> The question I'm asking here is : shouldn't package.json be updated as well ? What do you think ?

No, the point is that these deps are needed only for the production builds (npm did not resolve them during the development builds, thus they were never committed)

Nunien (approved):

PR #1163, opened 2021-12-29
implementing carmin client

alaeessaki:
This pull request is for the carmin implementation in Angular.
I tried to clean the code as much as i can.

All models in shanoir-ng-front/src/app/carmin/models are generated with open API.
I left the URL in the app utils static.
i deleted all the testing mock data, i didnt want to comment them to prevent zombie code.


Nunien:
Hi @alaeessaki , Am I right to say that this PR contains no new UI for now ? Also, does it contain any functional feature that we could test or should we wait for additional developments ? Cheers.

alaeessaki:
Hi @Nunien, For now, it contains only the services and the models used to call VIP for now.
It will be hard to test them since you need to install VIP in the same environment, i did test them locally with a mock UI, but since there is no UI specified, i deleted it before the PR.
But if you manage to install VIP locally with SHANOIR behind the same nginx, you can test them.
If you decide to do that, i can give you the guide with specific mock data.

Best regards.

shanoir-ng-front/src/app/carmin/models/descriptor.ts
```diff
@@ -0,0 +1,70 @@
+/**
```
Nunien, line 1: I think you should remove the "This class is auto generated" mention, as those files are versioned and will not be generated anymore.
Actually, after discussions with @michaelkain , we think you should remove entirely the current headers and use the Inria header that exists in every file in the Shanoir code : 
```

  
/**
 * Shanoir NG - Import, manage and share neuroimaging data
 * Copyright (C) 2009-2022 Inria - https://www.inria.fr/
 * Contact us on https://project.inria.fr/shanoir/
 * 
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 * 
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 */
```

There is a utility to automatically add those headers to the files that are missing it. It's /utils/license/update-license.sh .
It's not used often so it's not really up-to-date. 
To use it, pass the folder in wich you want to apply the headers as a first argument. For exemple in the shanoir root folder : 
``` ./utils/license/update-license.sh ./shanoir-ng-front/src/app/carmin ```
Before that you could edit /utils/license/license-header and set 2022 instead of 2019, it will be even better. Don't worry if you apply the script several times, it will detect where the license have already been added.

Nunien (changes requested):

michaelkain:
Thank you Julien for your feedback!

alaeessaki:
Hi @Nunien, 
I hope you are doing well. 
You can check the new updates.
With best regards.

Nunien (approved):

PR #1164, opened 2022-01-05
Shanoir-issue#1145 MriConverter as nifti converter

jcomedouteau:
- Implement in SQL
- Add jar (to be used directly ?) -> To be discussed.
How to test:
- try to convert different types of data with this new converter.


jcomedouteau:
@youennmerel you may go !
I had some problems with given dicoms, we can discuss about it, but it seems to be linked to the converter in itself.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/ImporterService.java
```diff
@@ -380,6 +380,7 @@ private boolean checkSerieForDicomImages(Serie serie) {
     }
 
     public void cleanTempFiles(String workFolder) {
```
youennmerel, line 382: Maybe this method should be remove altogether ?

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/dcm2nii/DatasetsCreatorAndNIfTIConverterService.java
```diff
@@ -343,6 +344,8 @@ private void convertToNiftiExec(NIfTIConverter converter, String inputFolder, St
 			// dcm2nii
 		} else if (converter.isDicomifier()) {
 			conversionLogs += shanoirExec.dicomifier(inputFolder, outputFolder);
+		} else if (converter.isMriConverter()) {
```
youennmerel, line 347: I think a switch case on "converter.getNIfTIConverterType" should be used here.

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/dcm2nii/NIfTIConverter.java
```diff
@@ -148,23 +148,27 @@ public String getDisplayString() {
 	}
 
 	public boolean isDcm2Nii() {
```
youennmerel, line 150: Except maybe for `isDicomifier` which is used in multiple classes  I don't really see the point of these `is[type]` method, when the converter type is otherwise publicly accessible

shanoir-ng-import/src/main/java/org/shanoir/ng/utils/ShanoirExec.java
```diff
@@ -615,4 +619,49 @@ public String dicomifier(String inputFolder, String outputFolder) {
...
+	 */
+	public String mriConverter(String inputFolder, String outputFolder, boolean reconversion) {
+		String logs = "mriConverter: ";
+		StringBuffer buffer = new StringBuffer();
+		
+		// We force the dataset0 folder here as MRIConverter does not search recursively in the files..
+		if (!reconversion && !inputFolder.contains("dataset")) {
+			inputFolder = inputFolder.concat("/dataset0");
+		}
+		LOG.error(inputFolder);
```
youennmerel, line 636: I think these are debug logs

youennmerel (commented):
Hi @jcomedouteau,

Some small remarks.

youennmerel:
Tested with MR, CT datasets + Bruker import 
File are well readable with Medinria

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/ImporterService.java
jcomedouteau: No, but it should absolutely not be commented ! thanks !

youennmerel (approved):
Checked it again on OFSEP qualif, OK to merge

PR #1165, opened 2022-01-05
implementing carmin Data

alaeessaki:
This pull request is for the carmin Data implementation in Dataset-ms
- All models are generated with OpenAPI using specification in : https://app.swaggerhub.com/apis/CARMIN/carmin-common_api_for_research_medical_imaging_network/0.3.1#/default/. 
- There is a difference between the open api's version used in carmin and the one used in shanoir, so i was forced to edit the generated files (only open api annotations). 
- The download actions are not implemented (only "content" is implemented : action used to download the content of a path).


michaelkain (changes requested):
Dear Alae,

I hope you are doing fine. A very happy new year to you. All the best for 2022!
Thank you for your pull request.

Please see my comment below:
- I am not sure if an empty interface OneOfGetPathResponse is really required, at least it looks strange
- How have you tested your implementation?
- Could you download a dataset zip from sh-ng using a datasetId in a path request?
- The class DatasetApiController is already very long. Could you please move your CARMIN DATA methods into a separate controller, that implements the Api? You might have to put some code lines into a Utils-class to avoid code duplication. This should help us to keep a leaner DatasetApiController class please.

Michael

alaeessaki:
Dear Michael Kain,
Thank you for your response and for your wishes, I hope you are also doing fine and I wish you a happy new year 2022.

For your comments : 

**- I am not sure if an empty interface OneOfGetPathResponse is really required, at least it looks strange**

I did see that, but i couldn't change or delete any of the generated models from open API fearing that it may cause some future unexpected bugs, but we can clear them the moment we finish the upload. And in my opinion, i guess that class is not used.
Should i delete it or we wait the upload implementation ?

**How have you tested your implementation?**

I tested those implementations with Postman. i did upload a dicom file with the help of shanoir UI, and i tried to download it with the endpoint. 
I didn't do any unit test, since i used the same business service.

**Could you download a dataset zip from sh-ng using a datasetId in a path request?**

Yes. The response sent with the POSTMAN is the same response expected (bytes[]). 

**The class DatasetApiController is already very long. Could you please move your CARMIN DATA methods into a separate controller, that implements the Api? You might have to put some code lines into a Utils-class to avoid code duplication. This should help us to keep a leaner DatasetApiController class please.**

I tried that at first, but I thought it may cause a code duplication. I wasn't sure if I have the right to change your download endpoint and use a utility class, but since you gave me the permission, I can now do that.

with best regards,
Alae

alaeessaki:
Hello @michaelkain,

How are you doing ?
I did push some changes about separating carmin data controller.
I tried to create a file utils as you can see, but i still see that there is some duplication. 
What do you think ? 
I'll try to optimize more and I'll let you know.   
I also find that creating a constant class that holds the constants (DCM, NIFTI, DOWNLOAD, etc...) not a bad idea.

ps : these changes are tested in my dev machine.

Have a nice weekend.

With best regards.

michaelkain:
Hi Alae, thank you for your above answers. I will have a look on your PR.
Could you please in the meantime fix the two conflicting files?

michaelkain (changes requested):
Hi Alae,

in general: thank you and we are close to acceptance!

Could you please explain your changes in:
- ShanoirChallengeMigrator/.project (not used anymore)
- shanoir-ng-parent/.project
Why is this filter necessary?

Thank you for the refactor with DatasetFileUtils.
@jcomedouteau Hi Jean-Come, could you please have a look on that, as the downloads
are your speciality?

@alaeessaki + @jcomedouteau I am wondering, if a Spring component, would not be more clean in our context than a static helper class? singleton by default.

With kind regards,
Michael

alaeessaki:
Hello Michael,

Thank you for your response.
I will try to fix the conflict. 
For the .project files, i had a problem with my intellij, and i switched to eclipse for a couple of days before fixing the intellij, and i've found some .project files and i thought that my eclipse is the one who created it. 
The filter is automatically added. 
Since i don't use eclipse anymore, I will stash the changes.

For your question about the spring component. In my opinion, it will be a better design

With best regards,
Alae

michaelkain:
Dear Alae,
thank you for your answer.
I wait for your conflict fixes and the above adaptions to continue my local testing on your PR.
With kind regards and have a nice day,
Michael

alaeessaki:
Dear Michael,

I fixed the conflict. You can now test in your local machine.
Is fixing the .project files required now for your local test ?.

With best regards.
Alae

michaelkain:
Dear Alae,
thank you for your fast answer, and the conflict fixes.
I prefer to do one test on my local machine with a close to final version.
So, please remove the unnecessary files: .project.
I think you made an error in your merge with develop.
It contains now a class, that has been removed from develop:
shanoir-ng-studies/src/main/java/org/shanoir/ng/bids/utils/BidsDeserializer.java

And please take a decision: either you move your DatasetFileUtils code into a Spring
component, what could be nice but not mandatory, OR you keep it as it is and I test it.
You wrote me above, that moving to a Spring component is a good idea, but then you
might have to execute it.

So, please contact me for final testing, when you have a clean stand,
thank you and with kind regards,
Michael

michaelkain:
all your shanoir-ng-studies/.. files should not be included
and please remove shanoir-ng-front/package-lock.json too

alaeessaki:
Hello Michael,
I hope you're doing well.
I'm sorry for the late push. Please read my comments below.

- I deleted the unnecessary filter tag in the .project files (i didn't delete them since i found them in the develop branch).
- I deleted the bids merged files from shanoir-ng-studies.
- I changed the Utility class to a spring component (tested on my local machine). 
But currently i'm working on the unit test mock injection since it's a component now. i have problems running unit test on my machine, so i'm using the pipelines in the github. 

I will contact you when i finish the unit test.

Thank you for your time.
Have a nice day.
And with best regards,
Alae

alaeessaki:
Hello Michael,
How are you doing ?

I failed mocking the utility class methods, the pathUrls always stay empty , then throwing 422. I didn't want to expect any return since I don't know the real behavior. 
For now, i used the constructor injection, just to execute the real methods while mocking (i tried using doRealMethod().when() but it didn't work).
What do you think ? should i leave it this way? If not, how can i mock those methods ?

Thank you.
With best regards,
Alae

michaelkain:
Hi Alae,
thank you for your clean-ups, this helps already.
Now your refactor contains an error: within a Spring DI context, we never use "new" ourselves
and create objects with the constructor, what means by-passing the Spring context.
You will have to use @Autowired or another Spring annotation to initialise your object, not new.
@Autowired
private DatasetFileUtils datasetFileUtils;

michaelkain:
so your object of DatasetFileUtils is never accessible and visible for Spring, but only in your class DSApiController, what might explain the behaviour you are having

alaeessaki:
Hello Michael,

Thank you for your feedback.
For the injection, i used @Autowired at first so it will be visible to spring. 
But the problem i had, is when testing with Mockito, since it's a mock, the methods in my utils class are never used, i should always do when().thenReturn(). But those method should always work when testing. 
I tried executing the real methods with mockito, but nothing helps.

Thats why i was forced to use the plain injection. 
Do you have any suggestions ?

Thank you for your help, have a nice weekend.
with best regards,
Alae

michaelkain:
Hi Alae,
have you tried in your Test class:
@MockBean or @Autowired
private DatasetFileUtils datasetFileUtils;
That your component is available for the test too?
You might add it as separate component somewhere, that your initialised test-api-controller can use it.
With kind regards,
Michael

michaelkain:
I would try @MockBean first

alaeessaki:
Hello Michael,

Yes I did, and indeed it execute those methods, but the pathUrls always stay empty, leading to throw 422. 
It's like the methods are not doing what they suppose to do. 
I will try to solve this. 
Is it ok to use the github pipeline to Test, i have problems running test on my local machine.

Have a nice weekend.
With best regards,
Alae

michaelkain:
Hi Alae, no worries, you can use GitHub pipeline to test. Cheers, Michael

michaelkain:
as discussed, I closed this PR

PR #1175, opened 2022-01-07
Shanoir issue#1173 Seperate EEG import in two steps

### LINKED ISSUE #1173: [EEG import] decorelate archive upload step from data analysis step
At the moment the upload process and "analysis" process are both in the same step.
As analysis can be quite long, and upload too, the keycloak token can be too short.

Decorelate these two
Step one: upload file -> we return "the file is stored here XXX"
Step two: please analyze this file XXX

There will be almost no change in the user's point of view (we can juste update the message)  but actually make two different calls.

jcomedouteau:
EEG import makes now 2 calls instead of one.

Closes #1173


jcomedouteau:
Needs to be checked (build failed)

jcomedouteau:
Is overridden by another one. Closing this

PR #1193, opened 2022-01-31
Shanoir issue#1190 center deletion limitation

### LINKED ISSUE #1190: [Centres] Center can be deleted even if still linked to a dataset (and solr search)
How to reproduce:
- Create a center
- Create an equipment and link it to the center
- Create a study card using this center / equipement
- Import data on this center
- Update the center -> Change equipement
- Update the study, remove study center
- Delete the center
-> The examination is now fantom.
-> Solr is still linked to the old center

jcomedouteau:
How to reproduce:

Create a center
Create an equipment and link it to the center
Create a study card using this center / equipement
Import data on this center
Update the center -> Change equipement
Update the study, remove study center
Delete the center
-> The examination is now fantom.
-> Solr is still linked to the old center

It is no more possible to delete a center when linked to an examination.

Closes #1190

### DIFF
--- a/shanoir-ng-studies/src/main/java/org/shanoir/ng/center/service/CenterServiceImpl.java
+++ b/shanoir-ng-studies/src/main/java/org/shanoir/ng/center/service/CenterServiceImpl.java
@@ -17,6 +17,7 @@ package org.shanoir.ng.center.service;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.shanoir.ng.center.model.Center;
 import org.shanoir.ng.center.repository.CenterRepository;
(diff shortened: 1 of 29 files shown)


michaelkain (changes requested):
Hi Jean-Come,
thank you for this PR. I have been thinking about it.
I would prefer an implementation, where we extend study_examination in MS Studies with a centerId, that we can check on that ID locally, before deleting a center. So MS Studies can delete or not a center as an independent microservice, even if MS Datasets is down.
On the other hand, as already done today, when we create an exam in MS Datasets, we already inform MS Studies about, AFTER the creation of the exam in the db.
I think this approach better respects the separation of concerns principle in a microservice architecture. Your solution will fail if MS Datasets is down, even if MS Studies is the master of the centers.
I hope this sounds logic, with kind regards,
Michael

jcomedouteau:
I understand, this is feasible I think, we just need to create another structure to hold these informations.
Thanks for the feedback !

shanoir-ng-studies/src/main/java/org/shanoir/ng/center/service/CenterServiceImpl.java
```diff
@@ -68,6 +78,12 @@ public void deleteByIdCheckDependencies(final Long id) throws EntityNotFoundExce
 		if (!centerOpt.get().getStudyCenterList().isEmpty()) {
 			errors.add(new FieldError("unauthorized", "Center linked to entities", "studies"));
 		}
+
+		List<StudyExamination> exams = Utils.toList(studyExaminationRepository.findByCenterId(id));
+		if (exams.isEmpty()) {
```
michaelkain, line 83: I think you mean: !exams.isEmpty()?

jcomedouteau: Ouch, yes. It's corrected.

michaelkain (approved):

PR #1232, opened 2022-03-10
Error manager

Nunien:
* Messages now display at the bottom of the sidebar.
* Messages are improved


jcomedouteau:
I started to test a little bit:
It appears all blank sometimes, but it is hardly reproducible.
image

It's a really cool feature to keep track of the errors, but I'm wondering if we're not loosing some clarity.
Actually we could integrate this in a more developed error management feature. A lot of error are not taken into account at the moment and the displayed message is not adapted -> We could check this all better..

let's discuss about this this afternoon :)

Nunien:
I think I will stop here

jcomedouteau (approved):

PR #1241, opened 2022-03-17
Clinical context bug

Nunien:
Fix bug when selecting a non-compatible study card in clinical context


jcomedouteau:
Do you have hints on haw to reproduce this one ?

Nunien:
You could deploy it on qualif and use the same study

Nunien:
I deleted my previous message as I was mixing up with another issue.

To reproduce, try to select a non compatible study card (displayed in red) in clinical context.

jcomedouteau (approved):

Issue #1212, opened 2022-02-18
User role and the Study page


Nunien:
The study page uses getAll() from the user API but the user API forbids a USER to access it. It leads to an error msg ont the front.
- Is the getAll really necessary in this case ?  (it's for building study-user entities)
- should a simple USER be able to see the members of a study ?

PR #1242, opened 2022-03-18
fix user getall bug in study

Nunien:
See https://github.com/fli-iam/shanoir-ng/issues/1212


jcomedouteau:
I still have weird bug
image

Coming from this code:
```
                saveSelection() {
                    this.breadcrumbsService.currentStep.data.tableState || (this.breadcrumbsService.currentStep.data.tableState = []),
                    this.breadcrumbsService.currentStep.data.tableState[this.hash].selection = [...this.selection]
                }

```
It happens when i click in the search bar of members, then click somewhere else

And when I press "Enter" in the search bar, it feels like it tries to download datasets or a study file:
image

```
                getFilename(e) {
                    const t = "attachment;filename=";
                    let n = e.headers.get("Content-Disposition");
                    return n.slice(n.indexOf(t) + t.length, n.length)
                }
```

Nunien:
@jcomedouteau thanks, it's fixed now

PR #1245, opened 2022-03-24
shanoir-issue#1234 dataset download comment contains '/' bug

### LINKED ISSUE #1234: [Dataset] Cannot download dataset with name with '/'
dataset 115896 cannot be downloaded
-> If download name contains '/' -> replace it with a  '_'
-> Disallow '/' char from dataset name ?

jcomedouteau:
Dataset 316887 failing in production -> dataset comment is used in folder path (and not file name)

Closes #1234


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -341,7 +341,9 @@ public void downloadDatasetById(
 		if (dataset.getUpdatedMetadata() != null && dataset.getUpdatedMetadata().getComment() != null) {
 			datasetName += "-" + dataset.getUpdatedMetadata().getComment();
 		}
-
+		if (datasetName.contains("/")) {
```
michaelkain, line 344: Hi Jean-Come, maybe it would be better to be OS-independent and work with File.SEPARATOR?

michaelkain (approved):

PR #1246, opened 2022-03-24
disable study select-box when loading study

### LINKED ISSUE #1202: [Import] Subject list loading time in import context can lead to unlogical state
This happened with the ICAN/UCAN studies. ICAN studies has a lot of subjects.

When importing a dataset, the import context automatically choose a study (ICAN) and a study card -> This leads to the subjects loading.

If we change the study while the subjects loading is not finished, a new study (UCAN) / study card is selected, then UCAN subjects are loaded BEFORE the first request ends.
-> The subjects of the first study are then displayed with the second study -> That leads to an un-logical error. (and import can no longer be OK)

How to correct this ?

Nunien:
this should fix #1202

### DIFF
--- a/shanoir-ng-front/src/app/import/clinical-context/clinical-context.component.html
+++ b/shanoir-ng-front/src/app/import/clinical-context/clinical-context.component.html
@@ -28,7 +28,8 @@ along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
                         (userChange)="onSelectStudy(); onContextChange();" 
                         (onViewClick)="showStudyDetails()" 
                         [viewDisabled]="!study" 
-                        [options]="studyOptions">
+                        [options]="studyOptions"
+                        [disabled]="studyLoading">
                 </select-box>
             </span>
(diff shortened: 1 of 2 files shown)


jcomedouteau:
I tested on qualif and It actually does not work as expected.

The boolean should be set to false only when all objects (study cards, centers, subjects) are loaded.
At the moment if I select ICAN, I select a studycard, then I change to UCAN then select a study card, then the same problem appears.
As the boolean is reset to false at the end of the "changeStudy" method, the method that loads subject finishes after. 

If you want it to work that way, the flag should be set up to true every time we load a list of elements and reset to false when the list loading is over.
We can even imagine not only blocking the study combo box but all combo boxes (loading subject, we don't want the user to change the center or study card). And the boolean reset should be done inside the list loading method "then".
Tell me if this isn't clear :)

Nunien:
@jcomedouteau thank you for that feedback

Nunien:
@jcomedouteau I think it's good now

PR #1250, opened 2022-03-31
reduce the exposed surface of the keycloak UI/API

a-ba:
This new setup blocks all locations by default and only allows the
minimal set needed by the users.


michaelkain:
Thank you Anthony!

PR #1255, opened 2022-04-04
Keycloak: update to new version 16.1.1

michaelkain:


michaelkain:
My local tests worked:
- ./bootstrap.sh --clean worked
- all 3 default users, created in KC
- PW change worked
- could connect with user type admin + user
- with user type admin: all microservices worked: studies, preclinical, import, datasets, search, ohif-viewer, users
- an import worked with sample.zip
- as admin: deletion of user deleted user in KC too

michaelkain:
Hi Julien,
it would be great, if you could test other user features, like password forgotten, reset password, account request, extension request etc if okay for you. With kind regards, Michael

Nunien (approved):

PR #1262, opened 2022-04-07
Move the OHIF viewer into the nginx image

a-ba:
Since the only purpose of the ohif-viewer container is to serve static files, this PR moves them into the nginx image.

The nginx dockerfile is now multi-stage (derived from upstream images `nginx` and `ohif/viewer`).

I also had to change the way of configuring nginx to use the plaintext config (http.conf) or TLS config (https.conf). Previously it implicitely derived that information from SHANOIR_URL_SCHEME, but we cannot no longer do this because this variable is present in the ohif-viewer's app-config.js. The qualif & prod instances whose TLS endpoint is implemented by a separate frontal server need to use the plaintext config but with SHANOIR_URL_SCHEME=https for the app-config.js. I introduced two variables SHANOIR_CERTIFICATE_PEM_CRT and SHANOIR_CERTIFICATE_PEM_KEY which hold the filenames of the server certificate & key. When they are both set to 'none' then the nginx container uses the plaintext config.


jcomedouteau (approved):
Tests working well for me

michaelkain:
@a-ba 👍 very good, I did not know of this double inheritance in docker

PR #1266, opened 2022-04-08
Shanoir issue#1257 Update spring boot to 2.6.6

jcomedouteau:


jcomedouteau:
@michaelkain 
Managed to correct the tests, now we have this exact same problem:
https://stackoverflow.com/questions/70207564/spring-boot-2-6-regression-how-can-i-fix-keycloak-circular-dependency-in-adapte

I tried the solution they propose (moving the keycloak configuration to another @Configuration file) but it didn't seemed to work( error: Caused by: java.io.FileNotFoundException: Unable to locate Keycloak configuration file: keycloak.json )

I'm still investigating it.

Jean-Côme

jcomedouteau:
I managed to get something working, by allowing the circular dependency. It's not really clean, but it seems to be working properly.
Tell me if you find another solution.
Get well and see you soon,

Jean-Côme

Nunien (approved):
It seems that allow-circular-references: true restores the behavior of the previous version.

Nunien:
Tested : 
* study creation
* import

Issue #1261, opened 2022-04-07
ohif viewer unusable with dicom study > 9


a-ba:
When requesting a dicom study with an id greater that 9, eg: https://shanoir-test.irisa.fr/viewer/10
the viewer fails with: *Failed to retrieve study data*

We have an error 500 in the nginx log:
```
"GET /dicomweb/studies/%C3%97/series HTTP/1.1" 500
```
The viewer requests `/dicomweb/studies/%C3%97/series` instead of `/dicomweb/studies/10/series`

On the other hand if we request a study with a lower id: https://shanoir-test.irisa.fr/viewer/9 then the viewer works well, and the log contains:
```
"GET /dicomweb/studies/9/series HTTP/1.1"
```

I suspect this is due to the fact that the id is expected to be StudyInstanceUID (which means starting with a well-formatted org root)

michaelkain:
tested, now working on qualif

PR #1267, opened 2022-04-11
Fix for GitHub issue #1261 Ohif viewer higher 9 issue

michaelkain:
I tested with examIds from 1 to 16.000.000, so it should be fine now


michaelkain:
Fix for GitHub issue: https://github.com/fli-iam/shanoir-ng/issues/1261

Issue #1269, opened 2022-04-13
Admin: create user not working


michaelkain:
Hi Julien,
could you please have a look on that?
On shanoir.irisa.fr, when I manually try to create a new user in Manage users with the +-Button
and enter just normal data, first name, last name, email address and an expiration date in one year
and user type user, an exception is thrown.
Thank you and with kind regards, Michael

PR #1270, opened 2022-04-14
fix error when creating user

Nunien:
see #1269


michaelkain (approved):
tested locally, fixed

Issue #180, opened 2019-08-05
ShUp: Add license header to shanoir-uploader


michaelkain:
Dear Julien, could you maybe run your tool on the folder shanoir-uploader to add the license header into all files of ShUp? Thank you for your answer, cheers, Michael

PR #1273, opened 2022-04-15
Shanoir uploader licenses

julien-louis:
[issue #180](https://github.com/fli-iam/shanoir-ng/issues/180)


michaelkain:
Dear Julien,
thank you for this PR. It would be great if the licenses would be up-to-date with the year 2022,
instead of adding new licenses that contain 2019 as year.
With kind regards,
Michael

PR #1275, opened 2022-04-19
support single sign-out with the 'check-sso' keycloak adapter mode

a-ba:


michaelkain (approved):
Dear Anthony, THANK YOU! I will have to test on qualif. With kind regards, Michael

PR #1276, opened 2022-04-20
fix the favicon

a-ba:
It seems that the build process (angular?) corrupts it if named with
the .ico extension


a-ba:
It does not display on firefox esr

Even the *file* utility does not detect its format:
```
# docker cp prod-nginx:/etc/nginx/html/assets/images/favicon.ico /tmp/
# file /tmp/favicon.ico 
/tmp/favicon.ico: data
```

though it looks like a png file
```
# head -c8 /tmp/favicon.ico 
�PNG
```

Nunien (approved):

PR #1288, opened 2022-05-04
functional rest api for uploading extra data with auto exam creation (wait for Aymeric validation)

Nunien:
Hard to test since no front functionality uses it.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/security/DatasetSecurityService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/security/DatasetSecurityService.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/security/DatasetSecurityService.java
```diff
@@ -87,6 +94,54 @@ public boolean hasRightOnStudy(Long studyId, String rightStr) {
...
+    }
+    
+    /**
+	 * Check that the connected user has the given right for the given subject.
+	 * 
+	 * @param subjectName the study name
+	 * @param rightStr the right
+	 * @return true or false
+	 */
+    public boolean hasRightOnSubjectName(String subjectName, String rightStr) {
```
jcomedouteau, line 128: Maybe refactor the two methods, load subject by subject name then call only one method with subject in entry.

jcomedouteau: Actually, hasRightOnSubjectId is not used no ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
```diff
@@ -260,8 +270,46 @@ public ResponseEntity<Void> addExtraData(
...
+		if (centerRepository.findById(centerId).isEmpty()) {
+			ErrorModel error = new ErrorModel(HttpStatus.UNPROCESSABLE_ENTITY.value(), "Couldn't find center with id " + centerId);
+			throw new RestServiceException(error);
+		}
+
+		Examination examination = new Examination();
+		examination.setComment(file.getOriginalFilename());
+		examination.setCenterId(centerId);
+		examination.setSubjectId(subject.getId());
+		examination.setStudyId(subject.getSubjectStudyList().get(0).getStudy().getId());
```
jcomedouteau, line 297: I guess that in the specifications, all the subjects are only part of one study ?

jcomedouteau:
I think you can remove the unused method in DatasetSecurityService.java then it's OK for me !

PR #1303, opened 2022-05-09
fix objectMapper for study name update

Nunien:


jcomedouteau:
There seems to be tests in error, the objectMapper should be autowired in tests too ! :)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/amqp/RabbitMQDatasetsService.java
```diff
@@ -274,8 +270,6 @@ private <T extends IdName> T receiveAndUpdateIdNameEntity(final String receivedS
 			)
 	@Transactional(isolation = Isolation.READ_UNCOMMITTED,  propagation = Propagation.REQUIRES_NEW)
 	public void createDatasetAcquisition(final String studyStr) {
-		SecurityContextUtil.initAuthenticationContext("ADMIN_ROLE");
```
jcomedouteau, line 277: This line should not be removed!

Nunien: thanks

jcomedouteau (approved):

PR #1309, opened 2022-05-10
Shanoir issue#1294 Add subject therapy/pathology during subject creation

### LINKED ISSUE #1294: [preclinical] When creating a preclinical subject, new pathologies / therapies does not appear in the tables
(but hey are actually created..!)

jcomedouteau:
Closes #1294
How to test:
- Create a new preclinical subject (either import brucker data or directly go to preclinical-subject/create)
- Add a new pathology and therapy
- They are correctly displayed in the table


Nunien (approved):

Nunien:
I set the name to required, there was a bug blocking the creation if omitted.  It's because the name is needed to generate an identifier.

PR #1315, opened 2022-05-19
fix mapper config : update spring mapper

Nunien:


jcomedouteau:
I have this problem on Solr: 
image

```
                        if (this.solrRequest.facetPaging.has(t.facetName)) {
                            let n = e.facetResultPages.find(e=>{
                                var n, i;
                                return (null === (i = null === (n = e.content[0]) || void 0 === n ? void 0 : n.key) || void 0 === i ? void 0 : i.name) == t.facetName
                            }
```

Nunien:
So it's like objectMapper have a custom serialization for page objects in Spring. What I did for that is configure Page serialization to behave as a "default" serialization. The code is not very beautiful but I think it works.

PR #1319, opened 2022-05-23
remove subject back reference

Nunien:


Nunien:
Deployed on qualif : no error, subject create with complete subject study : 
Capture d’écran de 2022-05-23 14-56-18

PR #1327, opened 2022-06-01
Shanoir-issue#1324 shanoir-uploader filters update

jcomedouteau:


michaelkain:
ah, I see, please keep the system entry in the pom.xml, as it will not work otherwise on GitHub, as long as the lib is still required; I think you removed it locally to make it work.

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -453,25 +467,64 @@ protected boolean searchField(String searchedElement, String filter) {
...
+		// NB: It is possible to have AND ";" filters OR OR ";;" filters but not both at the same time for the moment.
+
+		if (filter.contains(";;")) {
+			valid = false;
+			filters = filter.split(";;");
+			for (String filterToApply : filters) {
+				if (filterToApply.startsWith("!")) {
+					valid =  valid || !filterWildCard(searchedElement, filterToApply.replaceAll("!", ""));
+				} else {
+					System.err.println(filterToApply + filterWildCard(searchedElement, filterToApply.replaceAll("!", "")));
```
michaelkain, line 484: Hi Jean-Come, would it be okay to use the logger here, that it comes into the su.log please?

michaelkain (approved):
Hi Jean-Come,
thank you and in general: approved!
Just please fix the pom.xml and the logger issue, and I will merge it.
THX and with kind regards, Michael

jcomedouteau:
It should be good now.

michaelkain (approved):

PR #1338, opened 2022-06-07
shanoir-issue#1335 download dicom failure - add error message file

### LINKED ISSUE #1335: [Download dicom] Download faulty dicom.
It can happen that a dicom is not available anymore on the PACS.
If so, all the associated dicoms of the concerned dataset are now unavailable.

Find a way to:
- Still download the available dicoms
- Warn the user that some dicoms were missing. (add an "error.txt file ?)

jcomedouteau:
How to reproduce:
- Modify a dataset_file from pacs to change the path and lead to an unknown path
- Download a full dataset with multiple dicoms
- BEFORE: none of the dicoms could be downloaded
- AFTER: Every available dicoms are downloaded and an error file is added to the archive
Closes #1335


Nunien:
Sorry I can't test it with qualif since the PACS doesn't respond

jcomedouteau:
Do you know what's happening with the qualif PACS ?

michaelkain:
Hi @jcomedouteau, was it this PR important for Ofsep, where you meant to fix the conflict?

jcomedouteau:
Oh yes, the conflict appeared after the merge of the other PR. I'm on it.

michaelkain:
Hi Julien, could you please have a look on this for the next test session and merge it please?
We will have to deploy this for the OFSEP as well as the other fix. Thank you and with kind regards, Michael

julien-louis:
Ok it works if the pacs url is well formatted but return no result.
Doesn't work if the url is null (doesn't exist in today's production data) or with a random string. I approve anyway.

julien-louis (approved):

PR #1342, opened 2022-06-14
shanoir-issue#1341 use ZipFile instead of zipInputStream -> 

### LINKED ISSUE #1341: Import impossible ( red folder name)
It is impossible to import any of my new folders ( .zip ok with dicom files ok) since last friday.
At the first step: the name of my folder turns
 red and I don't underdand why .

jcomedouteau:
Closes #1341 
This error happens when a zip file is not zipped an appropriate way.
> only DEFLATED entries can have EXT descriptor
[test.zip](https://github.com/fli-iam/shanoir-ng/files/8897532/test.zip)

See test.zip for tests.


michaelkain (approved):
Dear Jean-Come,
thank you for this PR. My code review is fine.
The local tests are fine too: 1) could still import using a sample.zip and OK
2) with your test.zip I get an error with patients == 0, but I guess, that is normal
I will merge it. Thank you and with kind regards, Michael

PR #1355, opened 2022-06-30
shanoir-issue#1268 Import mail - wrong user_id/name in mail

jcomedouteau:
- Get rid of x-user-event when not necessary
- Transmit information as string otherwise


jcomedouteau:
Yes, I agree, we should get rid of this header, I'm on it.

michaelkain (approved):
Dear Jean-Come,
thank you for this good PR!
I will merge it directly and we will test with ShUp etc. all imports during our upcoming test session.
With kind regards,
Michael

PR #1359, opened 2022-07-06
upload endpoint to the tmp folder 

KhalilKes:
This pull request is to integrate upload endpoint in MS-import, so that VIP can upload the results to the tmp folder using the path from URI.

### DIFF
--- a/shanoir-ng-import/src/main/java/org/shanoir/ng/importer/CarminDataApi.java
+++ b/shanoir-ng-import/src/main/java/org/shanoir/ng/importer/CarminDataApi.java
@@ -20,35 +20,32 @@ import org.shanoir.ng.importer.model.carmin.Path;
 import org.shanoir.ng.importer.model.carmin.UploadData;
 import org.shanoir.ng.shared.exception.RestServiceException;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.*;
 
 import javax.validation.Valid;
-
 
 @Api(value = "carmin-data")
 @RequestMapping("/carmin-data")
 public interface CarminDataApi {
 
-    @ApiOperation(value = "Delete a path", notes = "Delete a path and transitively delete all its content if it is a directory.", tags={  })
-    @ApiResponses(value = {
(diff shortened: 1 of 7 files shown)


michaelkain:
Dear Khalil,
by the way, thank you for this great split of pull requests, what make it very good to review and understand!
With kind regards and I will start to have a look on that one,
Michael

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/CarminDataApi.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.validation.Valid;
+
+
+@Api(value = "carmin-data")
+@RequestMapping("/carmin-data")
+public interface CarminDataApi {
+
+    @ApiOperation(value = "Delete a path", notes = "Delete a path and transitively delete all its content if it is a directory.", tags={  })
+    @ApiResponses(value = {
+            @ApiResponse(code = 204, message = "the deletion is successful and finished."),
```
michaelkain, line 34: the response codes are not correct from my point of view

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/CarminDataApi.java
```diff
@@ -0,0 +1,54 @@
...
+    @RequestMapping(value = "/path/{completePath}",
+            produces = { "application/json" },
+            method = RequestMethod.DELETE)
+    ResponseEntity<Void> deletePath(@ApiParam(value = "The complete path to delete. It can contain non-encoded slashes.", required=true) @PathVariable("completePath") String completePath);
+
+
+    @ApiOperation(value = "Upload data to a path", tags = "A request without content creates a directory (an error should be returned if the path already exists). A request with a specific content type (\"application/carmin+json\") allows to upload data encoded in base64. The base64 content (part of a json payload) can either be an encoded file, are an encoded zip archive that will create a directory. All other content (with any content type) will be considered as a raw file and will override the existing path content. If the parent directory of the file/directory to create does not exist, an error must be returned.")
+    @ApiResponses(value = {
+            @ApiResponse(code = 201, message = "The upload is successful and finished.", response = Path.class),
```
michaelkain, line 45: same here: 200 is OK, and e.g. 500 is error or something similar

PR #1373, opened 2022-07-21
execution cron task

KhalilKes:
This pull request contains the cron job to check VIP server if the execution is finished to update the execution status, after that it creates a ProcessedDataset to finish the cycle

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
@@ -9,39 +9,41 @@ import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.nio.file.Paths;
 import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
 import javax.ws.rs.NotFoundException;
-
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.apache.commons.compress.utils.IOUtils;
 import org.shanoir.ng.dataset.modality.ProcessedDatasetType;
+import org.shanoir.ng.dataset.model.Dataset;
 import org.shanoir.ng.importer.dto.ProcessedDatasetImportJob;
 import org.shanoir.ng.importer.service.ImporterService;
 import org.shanoir.ng.processing.carmin.model.CarminDatasetProcessing;
-import org.shanoir.ng.processing.carmin.model.Execution;
 import org.shanoir.ng.processing.carmin.model.ExecutionStatus;
 import org.shanoir.ng.processing.carmin.service.CarminDatasetProcessingService;
 import org.shanoir.ng.processing.model.DatasetProcessing;
 import org.shanoir.ng.processing.service.DatasetProcessingService;
 import org.shanoir.ng.shared.exception.EntityNotFoundException;
+import org.shanoir.ng.processing.carmin.model.Execution;
 import org.shanoir.ng.shared.model.Study;
+import org.shanoir.ng.shared.model.Subject;
 import org.shanoir.ng.shared.repository.StudyRepository;
+import org.shanoir.ng.shared.repository.SubjectRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.scheduling.annotation.Async;
-import org.springframework.security.core.Authentication;
-import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.web.client.RestTemplate;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
-import org.springframework.web.client.RestTemplate;
-
 /**
- * CRON job to request VIP api and create porcessedDataset
+ * CRON job to request VIP api and create processedDataset
  * 
  * @author KhalilKes
  */
@@ -78,6 +80,9 @@ public class ExecutionStatusMonitor implements ExecutionStatusMonitorService {
     @Autowired
     private StudyRepository studyRepository;
 
+    @Autowired
+    private SubjectRepository subjectRepository;
+
     @Async
     @Override
     @Transactional
@@ -90,17 +95,14 @@ public class ExecutionStatusMonitor implements ExecutionStatusMonitorService {
         RestTemplate restTemplate = new RestTemplate();
 
         while (!stop) {
-
             Execution execution = restTemplate.getForObject(uri, Execution.class);
-
             try {
-
                 CarminDatasetProcessing carminDatasetProcessing = this.carminDatasetProcessingService
                         .findByIdentifier(this.identifier)
-                        .orElseThrow(() -> new EntityNotFoundException("entity not found with identifier :" + this.identifier));
+                        .orElseThrow(() -> new EntityNotFoundException(
+                                "entity not found with identifier :" + this.identifier));
 
                 switch (execution.getStatus()) {
-
                     case FINISHED:
                         /**
                          * updates the status and finish the job
@@ -135,7 +137,6 @@ public class ExecutionStatusMonitor implements ExecutionStatusMonitorService {
                     case KILLED:
 
                         carminDatasetProcessing.setStatus(execution.getStatus());
-
                         this.carminDatasetProcessingService.updateCarminDatasetProcessing(carminDatasetProcessing);
                         LOG.info("execution status updated stopping job...");
 
@@ -205,8 +206,8 @@ public class ExecutionStatusMonitor implements ExecutionStatusMonitorService {
     }
 
     /**
-     * 
-     * @param zipFilePath
+     *
+     * @param niiftiFile
      * @param destDir
      * @param carminDatasetProcessing
      */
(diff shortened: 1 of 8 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
```diff
@@ -0,0 +1,265 @@
...
+
+            processedDataset.setDatasetProcessing(datasetProcessing);
+
+            processedDataset.setProcessedDatasetFilePath(niiftiFile.getAbsolutePath());
+            processedDataset.setProcessedDatasetType(ProcessedDatasetType.RECONSTRUCTEDDATASET);
+            processedDataset.setStudyId(datasetProcessing.getStudyId());
+            processedDataset.setStudyName(study.getName());
+
+            // TODO get a subject for the processedDataset, hard coded for the time being
+            processedDataset.setSubjectId(1L);
```
michaelkain, line 238: Hi Khalil, thank you for your PR. The hard coded subject here does not seem to be a good idea.
With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
```diff
@@ -0,0 +1,265 @@
...
+    @Transactional
+    public void startJob(String identifier) {
+
+        this.identifier = identifier;
+        this.stop = false;
+
+        String uri = VIP_URI + identifier + "/summary";
+        RestTemplate restTemplate = new RestTemplate();
+
+        while (!stop) {
```
michaelkain, line 92: Hi Khalil, this while loop is maybe not the best solution. One separate job, that is invoked in regular intervals, for each process to monitor should be used, so a method that is regularly called will check for the finished state and kill itself after instead of having a while loop that runs all the time. For each started processing. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
michaelkain: You should find the link to the subject via the DatasetProcessing, that is already in the database and linked to the subject?

michaelkain:
Dear Khalil,
I hope you are doing fine. Sorry for what I will write now below, but unfortunately there might be quite a missing piece in your PR:
the multi-user management and handling.

The above code works only for a single user, not for all potentially logged users in shanoir. The advantage that all querying to the carmin-api in VIP is done by the Angular frontend, is that we are the current user, for each user that is logged, we get access to all information in VIP available for the current user. If I am not wrong, your current implementation in this PR is in the backend, no information on the token and the user, that called startJob. So how can the server, that runs all async threads know which what user to query the status of the pipeline using its identifier?
I would guess in VIP not everybody can query the status of a pipeline with an identifier, that is e.g. related to an user, that does not want to share its pipeline with everybody.

Two solutions:
A) my preferred one, but more work: put the processing status query into Angular, as the other carmin calls and use JavaScript workers to not block the GUI (should work), use the same frontend code as Jean-Come to complete the ProcessedDatasetImportJob, using the CarminProcessing you should have all information.
B) introduce an user notion into your async threads, but you will have to be careful with the token and its timeouts, when the waiting time increases as longer the pipeline runs (for a beginning we can all keep with 30 secs if easier for you, no prob).


- Good point: that you introduced a sleep time for running state, that VIP will not be victim of a denial-of-service-attack from shanoir :-).
- Furthermore I would propose to increase the sleep time, e.g. by doubling the time, like 10secs, 20secs, 40 secs, 80 secs, 160 secs. I think as longer a pipeline needs as less probable it becomes to have the result directly after? attention here with token timeout, see above
- Please take care of cancelling the thread as well (max timeout/living time), maybe some pipelines will never come to an end?

Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
michaelkain: the above is quite confuse, please read below

alaeessaki:
Hello Michael, 
Allow me to respond to this comment since I am the one who managed this endpoint authorization.

VIP API is secured by the same keycloak as Shanoir, when the user initializes an execution from Angular, the JWT token is sent in the Header.
VIP has the necessary information about the user, and it links the initialized execution with him in the database. so no other user can either manage or see others' executions.

This cron is triggered by Angular at the moment of initializing an execution, and its role is to check the status only, it sends a request to an endpoint called "execution summary", which returns only basic information of the requested execution such as the name, and status. and no other information is sent (no input or output results). I can say that the only security risk here is the workflow id. but since only the microservice is allowed one to call this endpoint, I don't see any objection to calling VIP from the internal URL.
I discussed this matter with Axel and he is ok with this endpoint being open as long as it's only to the microservice's IP.
this is how the endpoint is secured : 
```
.antMatchers("/rest/executions/{executionId}/summary")
.access(String.format("hasIpAddress('%s')", env.getProperty(ShanoirProperties.SendpointHANOIR_HOST_IP)))
```

example of basic information sent  : 
```
{
  "identifier": "string",
  "name": "string",
  "pipelineIdentifier": "string",
  "timeout": 0,
  "status": "Initializing",
}
```

But we are always open to discussing your solutions.

with kind regards, 
Alae

michaelkain:
Dear Alae, thank you for your explanation! Okay, now I know how you made it multi-user accessible, you opened the endpoint in VIP for all users (in the idea the ID might be unique enough, that no other logged user in shanoir might query for a pipeline he is not allowed to see maybe). And protect it by IP address. Just by curiosity, why did you not use the frontend code of Jean-Come to create the job and uses web workers to query the endpoint, when you have the token already in your requests?

alaeessaki:
Dear Michael, we thought if we use the cron in Angular, it will only work if the user is logged in to the browser. 
if the status of the execution is finished, the upload of the result won't happen only if the user is using the shanoir ng front. 
If the browser is completely closed then service workers can not run and will not dispatch any events.

But this is only a matter of user experience, we can agree that using web workers can increase performance since it's running on the user's device. But in my opinion, it's not user-friendly if the user can't see the status updated only if he is using the browser.

What do you think? I may be thinking of unnecessary matters. if that's the case, we can always adapt the code.
with kind regards.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
KhalilKes: Dear Michael, 

We don't really understand this part, we had an idea to select the subject in the front, as the subject is not part of the datasetprocessing, or maybe we didn't quite understand the link between them

looking forward to your reply, and Thank you always!

Khalil

michaelkain:
Dear Alae, thank you for this good and completely logic explanation, again! I should have thought of it myself.
Now it makes much more sense and I am fine with it. Another option would have been to use Window.localStorage,
create an item for all running processes in VIP and remove the item, when the job has been started to import the
processed dataset. When the browser is closed and opened again, and the user logs again, all running items could
be checked and completed, via the Angular frontend. But still not perfect, in a way, that when the user never comes
back some results may remain in VIP, unimported. Not so perfect, I we want in the future that multiple users can run
multiple pipelines in one study and share their results independent of the absence of the other user, as long as the pipeline has been run. So, option B) is definitely to prefer, but for this PR we keep it as it is. Thank you again and with kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
michaelkain: Dear Khalil,
I propose the following: get the processing and its inputDatasets, if their subjectIds are all the same, use it as subjectId for the processed dataset to create, as the information is correct then. If possible leave it empty, when you have multiple input datasets from different subjects, as it makes no sense to link a processed dataset to a subject, when two subjects have been used to process it. Delete the line with subjectName, as it is only part of the Job, but not stored in the dataset created, so we can forget about it. With kind regards and have a nice weekend, Michael

michaelkain:
Dear Khalil, as soon as my subject change proposal above is integrated, I will accept the PR. Thank you and with kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
michaelkain: last open change

KhalilKes:
Dear Michael, 

Thank you, as soon as I fix it and test it, I'll send you a review request and email if you want, have a nice day !

KhalilKes:
Dear Michael, 

First I apologize for the late PR, but I am actually facing an issue while running the bootstrap.sh, either with the force flag or the clean one, it stops at the Keycloak container creation which doesn't create the users, I don't know if you ever encountered this behavior before.

Looking forward for your kind reply

michaelkain:
Dear Khalil,
I hope you are doing fine.
Concerning your local dev environment, not production: the microservice users, that gets the users imported into his database using import.sql, does a synchronize to Keycloak to create the users during its startup procedure. MS users calls Keycloak from outside and creates the users in Keycloak. There is a flag for that behaviour syncAllUsersToKeycloak=true to activate or deactivate that behaviour. Therefore users requires the KC admin account to create the users from outside, that is given to ms users as environment variable. I am using bootstrap.sh on the current develop branch and until last week it worked fine and I am not aware of a change, that might impact that. Hm, maybe see in the logs of ms users, what is happening? Are you up-to-date with develop? There was a bug in bootstrap.sh some months ago, that has been solved by Anthony. With kind regards, Michael

KhalilKes:
Dear Michael, 

I hope this day finds you well.

yes in my dev environment, and I am up to date to the develop branch, for the logs of the microservice users, it just stop at the sync process, it could be that my env is acting up...
Actually i did a docker prune prior to this, but i don't think it'll has any consequences over it.
in any case I'll try to find a way around and test my latest changes today to close this PR. 
Thank you for your kindness and patience !

Khalil

michaelkain (approved):
Approved by code review: very good! Not tested, as no VIP instance.

PR #1387, opened 2022-08-25
OHIF-Viewer: save measurements + performance optimisation

michaelkain:
Contains:
- save measurements of ohif-viewer in backup pacs, version 1.0
- Migration to HttpClient 5 + performance optimization in DICOMWeb API: re-use same httpClient
- Refactoring: DicomServiceApi removed and independent components made + DICOMWeb client
calls unified in one component DICOMWebService
- adds ShUp integration test case, to test stow-sr
- adds View DICOM to each exam in the study tree
- done: change in DICOM SR file patient name/ID + studyInstanceUID, using examinationID, person name
- done: fix for measurement display in viewer
- done: create MeasurementDataset in shanoir database, now displayed in dataset list and details visible
- done: search it in shanoir Solr
- done: display details of MeasurementDataset: important values and deactivate image preview/nifti conversion, as no image
- done: test measurement update management
- DONE: test security of request filter, without valid JWT, no access anymore

LATER: options:
- todo: create multi-measurements in db based on measurement group items

**Additional info:**
- with dicom enhanced perfusion data, the stow-rs works, datasets are created, but the viewer seems to have problems to manage the overlay and show the measurements, even with the SR series are present and can be seen
- with "standard" dicom it works


michaelkain:
As discussed together, I removed the link to any dataset acquisition, as a measurement has no acquisition.
I use now the **referenced_dataset_for_superimposition_id** relation, as a measurment is a superimposition
on the original image, where I reference the original dataset used for the measurement. So many measurements
will not "pollute" the study tree, but can all be found and downloaded via the dataset search.
Each DICOM SR, Image Measurement Report, can be viewed using the ohif-viewer and the details on the SR
icons on the left side of the viewer.

michaelkain:
Dear Jean-Come,
it would be great if you could have a look please. No hurry.
I think now it is quite a round package. I tested, but please retest, the http filter, that manages the multipart/related for DICOM SR, is now protected by Keycloak as well. I could only called it with a valid auth-header from my unit test in shanoir-uploader. You could try it with Postman e.g. The "update" and manipulation of existing measurements creates a new measurement report any time in the viewer as well, so I apply the same philosophy in the backend and in my tests all changes, even deletions remained up to date, even when multiple SR dicoms remained in the pacs. So, to start with, it should be fine. With kind regards, Michael

michaelkain:
ah, I will have to provide the db_changes!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
```diff
@@ -81,7 +81,8 @@
 		@JsonSubTypes.Type(value = SpectDataset.class, name = "Spect"),
 		@JsonSubTypes.Type(value = StatisticalDataset.class, name = "Statistical"),
 		@JsonSubTypes.Type(value = TemplateDataset.class, name = "Template"),
-		@JsonSubTypes.Type(value = BidsDataset.class, name = "BIDS") })
+		@JsonSubTypes.Type(value = BidsDataset.class, name = "BIDS"),
+		@JsonSubTypes.Type(value = BidsDataset.class, name = "Measurement") })
```
jcomedouteau, line 85: MeasurementDataset instead of BidsDataset I guess ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/service/DICOMWebService.java
```diff
@@ -100,33 +139,204 @@ public ResponseEntity findFrameOfStudyOfSerieOfInstance(String studyInstanceUID,
...
+		httpPost.setHeader(HttpHeaders.CONTENT_TYPE, CONTENT_TYPE_MULTIPART+";type="+CONTENT_TYPE_DICOM+";boundary="+BOUNDARY);
+		httpPost.setEntity(entity);
+		try (CloseableHttpResponse response = httpClient.execute(httpPost)) {
+			int code = response.getCode();
+			if (code != HttpStatus.OK.value()) {
+				LOG.error("DICOMWeb: sendMultipartRequest: response code not 200, but: " + code);				
+			}
+		} catch (IOException e) {
+			LOG.error(e.getMessage(), e);
+			throw e;
```
jcomedouteau, line 239: I know you probably didn't touch this code, but the error/exception management is quite weird / un-logical along the class. (sometime we just log something, sometime we throw ShanoirException, sometime we rethrow the catched exception)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
michaelkain: totally right!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomSRImporterService.java
```diff
@@ -0,0 +1,399 @@
...
+	 * inside shanoir-database, to optimize mass data export of all measurements
+	 * e.g. for one study. This code works and has been tested to access the most
+	 * important measurement attributes and was hard to construct. Even if currently
+	 * not used, we keep it for a very high later usage.
+	 * 
+	 * @param datasetAttributes
+	 * @param measurementDataset
+	 */
+	private void completeDatasetFromDicomSR(Attributes datasetAttributes, MeasurementDataset measurementDataset) {
+		Sequence contentSequence = datasetAttributes.getSequence(Tag.ContentSequence);
```
jcomedouteau, line 284: We can discuss about this but we can imagine having a little bigger if
```
if (a != null && a.get(0) && a.get(0).getAttribute() != null [...]) {
    Attribute att = a.get(0).getAttribute()
}
```
I don't really know if it's better or not.
Otherwise 
```
if (a == null) {
return null;
}
```

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomSRImporterService.java
```diff
@@ -0,0 +1,399 @@
...
+	 * @throws IOException
+	 * @throws Exception
+	 */
+	private void sendToPacs(Attributes metaInformationAttributes, Attributes datasetAttributes)
+			throws IOException, Exception {
+		/**
+		 * Create a new output stream to write the changes into and use its bytes
+		 * to produce a new input stream to send later by http client to the DICOM server.
+		 */
+		ByteArrayOutputStream bAOS = new ByteArrayOutputStream();
```
jcomedouteau, line 390: Use a try with ressources ?

jcomedouteau:
I tested a bit this morning:
- I could add and save correctly measurements
- I could find them using solr (but not in study tree as expected)

The problem I encoutered:
- I could not make them re-appear in the viewer, SR dicoms were present in the viewer list of images, but I could not display them.
- I tried to delete one dataset from Shanoir, it seems that it was not deleted in the pacs ( and so still present in ohif viewer after deletion). I could'nt try deletion from ohif viewer.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/service/DICOMWebService.java
michaelkain: Good remark: I will change it.

michaelkain:
Dear Jean-Come, concerning your last comments:
- see my demo video, with my latest tests, when I use the sample.zip the measurements are re-displayed on the series
- I have the impression, that the behaviour changes depending on the image modality maybe, what data have you used?
- the deletion should only work with dicom.web=true, or is it depending on the wado-rs URL structure

jcomedouteau:
So I re-tested locally after the merge:
- The merge is not perfect, but a slight change make the trick in DicomSRImporterService.java
- Now all the measurements are well saved BUT are doubled (see screenshot, i moved the second one so that it is not superimposed) 
image

michaelkain:
Dear Jean-Come, thank you for your fix according to Julien's change on exam.subject/id. For me, concerning the irritating double display of measurements (twice on the same place), this is an error of the viewer, where I hope, that I will be solved by them in the future. So, I would merge my PR, to finish it, as measurements are stored and normally in a completely fresh opened viewer tab I hope it is correctly displayed. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomSRImporterService.java
michaelkain: Hi Jean-Come, currently the above code is not used, see my comment, and only kept in case, we would like to go further on the actual annotation content. On the other hand, even if the if block is unfortunately very deep (related to the weird access to the sub-items in dicom api), I have fear a more compact kind of code would make it even more complicated to understand on which level of item-depth we are operating to access information. I hope it is fine like this. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomSRImporterService.java
michaelkain: done, added, but on the outer stream, that closes the inner as well

michaelkain:
Dear Jean-Come,
I am wondering about your comment above. My code normally has never been applied on qualif, and qualif is cleaned every night - normally - so my new script should be unknown to qualif+prod. Furthermore I did not touch the above file in my PR and I had an build error on GitHub I had to solve as the internal db-changes script to apply on the database was not able to deal with my new file, as the merge with develop had introduced another file with the same number earlier. So please help me to understand the problem. With kind regards, Michael

jcomedouteau:
Michael,
I think I understood, you should merge develop to get the last development of Anthony on this file: studies/0006_study_user_center.sql that currently fails.
Jean-Côme

michaelkain:
Dear Jean-Come, I hope you are doing fine. Has it worked on qualif?

jcomedouteau:
Hi Michael,
The deployment worked well on qualif, but i could not use the ohif viewer, is it normal ?
I think I remember it was a 403 error when opening the viewer, but I need to check again. I'm re-deploying it.
Jean-Côme

jcomedouteau:
Ok it is now working !

jcomedouteau (approved):

michaelkain:
Dear Jean-Come, THANK YOU a lot! Happy to merge this :-)

PR #1393, opened 2022-09-05
fix exam center rights

julien-louis:


julien-louis:
Tested exam display, including preclinical

PR #1395, opened 2022-09-07
Fix for re-apply study cards

michaelkain:
taken from my PR to make work study card re-application


michaelkain:
for the next release, that contains SC re-apply, but not yet the quality control PR

PR #1408, opened 2022-09-14
fix the detection of existing migrations

a-ba:
The "check migration names" stage in the ci pipeline fails to detect existing migrations in the master branch. They appear as `(none)` in https://github.com/fli-iam/shanoir-ng/actions/runs/3046116136/jobs/4908518031

With proper detection this script would have caught the bugs that had to be fixed in ca84e9502b and d25400f693.

Note: #1407 has to be merged before this PR (otherwise this fix will make it fail the test)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -16,6 +16,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: fetch the master branch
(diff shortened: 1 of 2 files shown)


a-ba:
Note : the fix is broken, do not merge

a-ba:
The fix is good now. The PR can be merged.

PR #1420, opened 2022-09-23
clinical context refactoring

julien-louis:


jcomedouteau:
@julien-louis The bug is now resolved, but I have a problem when finishing the processed dataset import... 
image

julien-louis:
> @julien-louis The bug is now resolved, but I have a problem when finishing the processed dataset import... image
> 
> the line of code is:
> 
> ```
>                 startImportJob() {
>                     this.subjectService.updateSubjectStudyValues(this.importDataService.contextData.subject.subjectStudy).then(()=>{
> ```

Fixed !

Issue #1363, opened 2022-07-18
[NG-2022-9] Towards FAIR datasets in Shanoir: step 1 "Findable"


cmaumet:
This issue will contribute to the long-term goal of facilitating data sharing using Shanoir following the [FAIR principles (FAIR=Findable Accessible Interoperable Reusable)](https://www.go-fair.org/fair-principles/).

The goal of this particular issue would be to focus on the F (for Find) of FAIR. To this aim, we could create a new page on Shanoir to list all datasets that are available for sharing. 

This page should include:
 - Basic information about each dataset possibly using an existing schema (such as DATS or dandi datasets), e.g. number of participants, type of acquisitions, number of sites, etc.
 - A link to access the data (that will likely require having a Shanoir account).
 
These would include for instance the datasets shared by Michel Dojat or Hydrops (PI: Arnaud Attyé), or the MICCAI challenges (if publicly available).

michaelkain:
Starting point based on exchange with Alban, introduce "DataCatalog" of schema.org:
The idea is to declare a local Shanoir server instance as a DataCatalog, according to schema.org,
what will allow machines (web bots) to automatically identify our both production servers as data catalog.

albangaignard:
Here is the proposed markup as a first iteration : 

```
{
 "@context": "http://schema.org"/;,
 "@id": "[https://shanoir.irisa.fr/"](https://shanoir.irisa.fr/%22);,
 "@type": "DataCatalog",
 "dct:conformsTo": "[https://bioschemas.org/profiles/DataCatalog/0.3-RELEASE-2019_07_01"](https://bioschemas.org/profiles/DataCatalog/0.3-RELEASE-2019_07_01%22);,
 "description": "TODO",
 "keywords": [
 "Medical Imaging",
 "Neuroimaging",
 "Neuroinfirmatics"
 ],
 "license": "[https://www.gnu.org/licenses/gpl-3.0.en.html"](https://www.gnu.org/licenses/gpl-3.0.en.html%22);,
 "name": "Shanoir",
 "provider": [
 {
 "@context": "http://schema.org"/;,
 "@type": "Organization",
 "dct:conformsTo": "[https://bioschemas.org/profiles/Organization/0.2-DRAFT-2019_07_19"](https://bioschemas.org/profiles/Organization/0.2-DRAFT-2019_07_19%22);,
 "description": "France Life Imaging [...]",
 "name": "France Life Imaging"
 },
 {
[...]

PR #1455, opened 2022-11-14
ISSUE #1363 [NG-2022-9] Towards FAIR datasets in Shanoir: step 1 "Fin…

pierrehenri-dauvergne:
…dable"

### DIFF
--- a/docker-compose/keycloak/entrypoint
+++ b/docker-compose/keycloak/entrypoint
@@ -15,14 +15,16 @@ require  SHANOIR_VIEWER_OHIF_URL_HOST
 
 mkdir -p /tmp/import
 sed "	s/SHANOIR_ADMIN_NAME/$SHANOIR_ADMIN_NAME/g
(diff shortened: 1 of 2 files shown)


michaelkain:
Dear Pierre-Henri, thank you for your first PR! Great! It looks totally fine, but we might have to face one more challenge:
today in your template.ftl file two times the URL "https://shanoir.irisa.fr" is hard-coded. We can not keep it that way, as
we will have different servers, and each server should send the search bots to the correct source URL. Today, the original serverURL is configured in the ".env" file of docker-compose and injected into Keycloak as well. There may be a way to ask Keycloak for that information or to pass the info otherwise into the template.ftl file. We will have to think of that. Thank you and with kind regards, Michael

michaelkain (changes requested):
some minor extension necessary, see my comment

pierrehenri-dauvergne:
Hello Michael, 

Can you come back on the daily call so that I can show you my idea? 

Regards, 

PH

PR #1463, opened 2022-11-16
Adding multiple result extension support 

alaeessaki:
this PR contains adding multiple result extension support while importing the result.
And the possibility to import results from subdirectories from the result tar file.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
@@ -118,7 +118,7 @@ public class ExecutionStatusMonitor implements ExecutionStatusMonitorService {
                                         carminDatasetProcessing.getResultsLocation());
 
                         final PathMatcher matcher = userImportDir.toPath().getFileSystem()
-                                .getPathMatcher("glob:**/*.tgz");
+                                .getPathMatcher("glob:**/*.{tgz,tar.gz}");
                         final Stream<java.nio.file.Path> stream = Files.list(userImportDir.toPath());
 
                         stream.filter(matcher::matches)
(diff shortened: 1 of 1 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/schedule/ExecutionStatusMonitor.java
```diff
@@ -177,19 +177,34 @@ private void decompressTGZ(File in, File out, CarminDatasetProcessing carminData
         try (TarArchiveInputStream fin = new TarArchiveInputStream(
                 new GzipCompressorInputStream(new FileInputStream(in)))) {
             TarArchiveEntry entry;
+
+            File cacheFolder = new File(out.getAbsolutePath() + File.separator + "cache");
```
michaelkain, line 181: Dear Alae, thank you for your PR. When multiple threads run into that method, with the same result file name, what can possibly happen, they will write into the same cacheFolder. I would propose a cache folder either with an UUID or a timestamp with milliseconds to avoid overriding of different results. Thank you and with kind regards, Michael

michaelkain: see microservice import, that generates a random ID for that purpose

KhalilKes: Hello @michaelkain,

We do actually use a timestamp for the results folder, we create the path in the front, and take it back from the carminDatasetProcessing entity resultsLocation line 118

image

[https://github.com/alaeessaki/shanoir-ng/blob/adding-multiple-result-extension-support/shanoir-ng-front/src/app/processing/execution/execution.component.ts](url)
image

to recapitulate : 

- The user launch an execution
- A carminDatasetProcessing is created with a unique result path and saved in the database
- The thread checks if the execution is finished 
- Once the execution is finished we get carminDatasetProcessing with the unique  result path created previously
- Creation of the unique folder and the cache folder inside it 

I hope I explained the workflow, 

With kind regards.

Khalil

alaeessaki: hello @michaelkain, 
Hope you're doing well, 
We have already discussed the code in this PR; what we changed is only the way of importing files from a tar that contains folders, and we added multiple tar file extension support.
Thank you.

michaelkain: Thank you for your answers!

PR #1465, opened 2022-11-17
fixing circular dataset object bug

alaeessaki:


jcomedouteau:
Hello,
Yes, this is a hotfix we decided to do so that they could go on and avoir a circular reference error.
It's OK for me, we can merge.

Jean-Côme

jcomedouteau (approved):

PR #1479, opened 2022-12-05
shanoir-issue#1425-processing-circular-dependency - resolve

### LINKED ISSUE #1425: CarminDatasetProcessing creation error
While trying to save a carmindatasetprocessing from the webapp interface, the object seems to be facing a circular  / cyclic object value 
image

The object in question: 
image

the cyclic values seem to be: 
- subjectStudyList in the study object
- subjectStudyList in the subject object
- studyCenterList

As a hotfix we chose to empty those values as so : 

image

jcomedouteau:
Dear @KhalilKes,

I created this pull request to correct the problem you encountered during carmin dataset processing creation.
I could not test locally, as we havn't VIP on our servers.

The problem was:
1) To load selected datasets, you were doing one call by dataset ID, this is not a great solution for communication and database performance, this should be avoided as much as possible.
2) Each of this call returned a dataset and not a datasetDTO -> these carry all the stuff that was provoking the circular dependency problem

What I did:
- Created a new method in the back to load all datasets from a list of IDs
- This method returns datasetDTOs and not datasets

Closes #1425 

Please note that I could not test it, could you please try it locally so that we can adapt my solution ?
Thanks,
Jean-Côme


michaelkain (approved):
Approved by code review only. Tested by GME FLI.

PR #1482, opened 2022-12-06
Shanoir issue#1200

pierrehenri-dauvergne:


docker-compose/database/db-changes/studies/0010_add_profile_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE `profile` (
+  `id` bigint(20) NOT NULL AUTO_INCREMENT,
+  `profile_name` varchar(255) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+)
+
+ALTER TABLE study ADD profile_id bigint(20);
```
jcomedouteau, line 7: You should also add the foreign key reference -> SQL DB needs to know that this id references a profile.
To do this, show the create table of study and get the foreign key constraint

jcomedouteau (approved):

Issue #1478, opened 2022-12-02
[BIDS] Bids tree isn't compliant + sometime fails


jcomedouteau:
After last week discussion, re-check names in the tree.
+ fails for "IHU_ICAN_TEST_STUDY" for example. double check this.

PR #1487, opened 2022-12-13
#1478 - bids - Subject/study/exam folder name compliance

jcomedouteau:
How to test:
- Create a study with a studycard with a bids data type rule
- Import data in this study with a subject with a name containing a "_" or an empty space
--> Deploy BIDS
Dataset files are well named
Bids data type folder is well organized
scans.tsv file is correct

And the folder is well regenerated every time a change is done on the study.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/bids/service/BIDSServiceImpl.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/bids/service/BIDSServiceImpl.java
@@ -315,6 +315,19 @@ public class BIDSServiceImpl implements BIDSService {
 				buffer.append("session_id").append(TABULATION)
 				.append("acq_time").append(TABULATION)
 				.append(NEW_LINE);
+				
+				for (Examination examination : examinationList) {
+					String sessionLabel = "" + examination.getId();
+					sessionLabel += (examination.getComment() != null ? "-" + examination.getComment() : "");
+			
+					sessionLabel = sessionLabel.replaceAll(" ", "");
+					sessionLabel = sessionLabel.replaceAll("_", "");
+			
+					buffer.append(sessionLabel).append(TABULATION)
+					.append(examination.getExaminationDate()).append(TABULATION)
+					.append(NEW_LINE);
+				}
+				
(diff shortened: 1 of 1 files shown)


jcomedouteau:
Add refresh button to update BIDS folder (+ error message when failing ?)

shanoir-ng-datasets/src/main/java/org/shanoir/ng/bids/service/BIDSServiceImpl.java
```diff
@@ -201,8 +219,10 @@ public void deleteBidsFolder(Long studyId, String studyName) {
 	 * @throws IOException
 	 */
 	@Override
-	public File exportAsBids(final Long studyId, final String studyName) throws IOException {
+	public File exportAsBids(final Long studyId, String studyName) throws IOException {
 		// Get folder
+		studyName = studyName.replaceAll(" ", "");
```
youennmerel, line 224: I think you can do that in one line using regex :
`studyName = studyName.replaceAll("[ _] ", ""); `

shanoir-ng-datasets/src/main/java/org/shanoir/ng/bids/service/BIDSServiceImpl.java
```diff
@@ -392,9 +426,14 @@ private File createExaminationFolder(final Examination examination, final File s
...
+	private void createDatasetBidsFiles(final Dataset dataset, final File workDir, final String studyName, String subjectName) throws IOException {
 		File dataFolder = null;
 
+		String nature = null;
+		subjectName = subjectName.replaceAll(" ", "");
+		subjectName = subjectName.replaceAll("_", "");
+		String datasetFilePrefix = workDir.getName();
+
 		// Create specific files (EEG, MS, MEG, etc..)
 		if (dataset instanceof EegDataset) {
```
youennmerel, line 438: I think, in order to limit the size & complexity of the `createDatasetBidsFiles` method, we should put this if..else if..else if... part into a bigger `createDataFolder` method.

So that we just have to call `dataFolder = createDataFolder(dataset, workdir);` here.

youennmerel (approved):
Tested on qualif, all right now

PR #1491, opened 2023-01-06
Adding keycloak service Account

alaeessaki:
This PR contains the use of keycloak service account to communicate with VIP-API.
It contains also the fix of a thread variable conflict.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/controller/CarminDatasetProcessingApi.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/controller/CarminDatasetProcessingApi.java
@@ -5,6 +5,7 @@ import java.util.List;
 import javax.validation.Valid;
 
 import org.shanoir.ng.processing.carmin.model.CarminDatasetProcessing;
+import org.shanoir.ng.shared.exception.EntityNotFoundException;
 import org.shanoir.ng.shared.exception.ErrorModel;
 import org.shanoir.ng.shared.exception.RestServiceException;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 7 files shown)


michaelkain:
Dear Alae, thank you for your PR and a happy new year!
You are calling a Keycloak client ID, that is not part of the shanoir-ng-realm.json of Keycloak. So this will not work locally on my machine, right? So you might need your client into the configuration. Could you please rapidly explain me the overall architecture? You use always the same accessToken for all executions and your @async annotation creates always a new thread and this one stops, when the ProcessedDataset has been created right? With kind regards, Michael

michaelkain:
Dear Alae, furthermore I am wondering, if there is not a potential issue with the above solution:
- when we have an execution in VIP that runs for days and we stop the Shanoir server in between, the state of this CarminDatasetProcessing will never be updated, as the thread for that dies with the stopping of the microservice
- would it not be better to have a CRON JOB, that runs all 10 minutes to update all processing for each execution, every time the server is up (so when re-started the updates continue to happen)
With kind regards, Michael

alaeessaki:
Hello Michael,I hope you're doing well.
For your first comment. I've prepared a small diagram to simplify the architecture of this concept.

image

to resume, it's a keycloak functionality called service account (it's for a secured communication between the microservices in a service level and not a user level). I've created a keycloak client ID called service account, I activated the service account functionality and I limited the security (access token life 60s, no refresh token and no session, and limited roles). 

For the Async, it uses the service account id and client secret to get an access token and uses it to call VIP API. if the response is unauthorized it gets a new access token and uses it again (in case of an unauthorized error, it's limited to 3 attempts).

You can see my presentation below for more details : 
https://docs.google.com/presentation/d/17IjcyJPUSUe0KLdFp22FTGy_kdHOVcETm23OQvr9lr8/edit?usp=sharing

alaeessaki:
For the second comment : 

I understand the issue, if you stop shanoir before ths execution ends, it will not track the execution anymore.
There is always a space for impromvement, and i think your solution is one of them. 

I can also think of possibly other solutions :

- we can add @PreDestroy method that adds all of the executions with the status "RUNNING" in a fail over table before the shutdown and a commandline runner that rerun all the threads from this table after a restart.

- Or we can add a command line runner that simply reruns all the threads of all executions with the status "RUNNING".

michaelkain:
Dear Alae, thank you for your answers and the diagram, that explains a lot. Fine for me, to keep the point with the Cron Job for a future optimization. BUT it is important to add your client to the configuration.json, that it exists by default, otherwise it will not work in the code (I have not found it as default client).

alaeessaki:
Yes, vip client also should be added in configuration.json. 
We can do this in another PR to seperate reviews.

michaelkain:
Hi Alae, I am merging the PR to not block your progress. As always, I did only a code review, as we do testing on your environment. Please provide us the necessary Keycloak client configurations in a separate PR, to be added to /docker-compose/keycloak/cfg/shanoir-ng-realm.json Thank you and with kind regards, Michael

alaeessaki:
Dear Michael,
Thank you for your review, I've noted the keycloak configuration task, it will be created later.
With kind regards, Alae

michaelkain:
Perfect Alae, thank you too for your pull request!

PR #1503, opened 2023-01-18
improve the oneshot script

a-ba:
- exit as soon as the final pattern is matched (instead of hanging if any pattern is not matched)
- check additional messages in keycloak init/import/export logs


michaelkain (approved):

michaelkain:
Thank you Anthony!

PR #1505, opened 2023-01-19
shanoir-issue#1495

pierrehenri-dauvergne:


shanoir-ng-studies/src/main/java/org/shanoir/ng/tag/model/SubjectTag.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@JsonIgnore
+	@ManyToOne
+	@JoinColumn(name = "study_id")
+	private Study study;
+	
+	@JsonIgnore
+	@OneToMany(mappedBy = "subjectTag", fetch = FetchType.EAGER, cascade = CascadeType.ALL, orphanRemoval = true)
+	@OnDelete(action = OnDeleteAction.CASCADE)
+	private List<SubjectStudyTag> subjectTag;
```
jcomedouteau, line 37: This may have to be renamed subjectStudyTags

PR #1509, opened 2023-01-20
Vip user synchronization

alaeessaki:
Dear Michael,
I hope you're doing well. 
the following PR contains VIP user synchronization. 
For the merge check to the development branch, as discussed earlier, we will create a PR from develop to gme-fli-pre-prod and fix the merge conflict (if it exists) after the first version validation. So you can merge this PR for now,
With kind regards,
Alae


alaeessaki:
Hello Michael,

For the first proposition, since I didn't find how to access properties files, I preferred not to. but since now it's duplicated I guess it's time to move it to ms commons, but do you have any suggestion as to how to access properties (for keycloak URL and client ID and Secret), or I make the method dynamic and use the application properties of each microservice that uses this keycloak util?

With best regards, Alae

michaelkain:
Hi, good point. The need for application.yml properties makes a difference. Have you tried just moving KeycloakServiceAccountUtils to shanoir-ng-ms-common, without any change, and checked if it works? The compiler should not complain, and the values should be solved depending on the microservice where we are (so you keep the duplication of the properties in each ms in application.yml, but avoid the code duplication in Java). The Spring ComponentScan should find the component with the standard package. The good point is, that we will see from the application.yml, what dependencies exist.

alaeessaki:
Hello Michael,

I hope you're doing well,
can you please check the new updates.

With best regards, Alae

michaelkain (approved):
Hi Alae, well done! I approve it via code review only. We will do the functional tests on your environment. With kind regards, Michael

PR #1514, opened 2023-01-25
shanoir-issue#1502-3: remove unused dependencies

pierrehenri-dauvergne:


shanoir-ng-front/src/app/app-routing.module.ts
```diff
@@ -432,10 +431,6 @@ let routes: Routes = [
         path: 'user',
         redirectTo: 'user/list',
     },
-    {
```
julien-louis, line 435: Why did you remove the user list from Shanoir ?

julien-louis: I'm pretty sure it's the user list you get at https://shanoir-ng-nginx/shanoir-ng/user/list . You have to be an administrator though.

pierrehenri-dauvergne: Yes indeed, you're right. This component was defined as unused by depcheck. I knew used dependencies could be listed by depcheck but I didn't pay enough attention here.

Issue #1515, opened 2023-01-26
[Processed dataset] Cannot download processed dataset


jcomedouteau:
As dicom -> This is quite normal, but we need to display a correct warning / error

As nifti, the rights check fails:

> java.lang.NullPointerException: null
        at org.shanoir.ng.dataset.security.DatasetSecurityService.hasRightOnEveryDataset(DatasetSecurityService.java:382)

This is due to the fact that the dataset acquisition is null  for processed dataset.
-> check on parent dataset

michaelkain:
Perfect, it would be great to solve that for the V1 of the GME FLI and to complete this functionality on processed datasets

PR #1516, opened 2023-01-27
#1515 download processing download

jcomedouteau:
- Do not download dicom for processed datasets
- Download nifti correctly

How to test:
- Import a processed dataset
- On solr search, select the processed dataset and a MR dataset
- Download as nifti -> Both are downloaded
- Download as dicom -> Only the MR is downloaded

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
@@ -555,7 +555,13 @@ public class DatasetApiController implements DatasetApi {
 				}
 				String studyName = studyRepo.findById(dataset.getStudyId()).orElse(null).getName();
 
-				Examination exam = dataset.getDatasetAcquisition().getExamination();
+				Examination exam;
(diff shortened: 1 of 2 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -657,7 +667,6 @@ public void massiveDownload(String format, List<Dataset> datasets, HttpServletRe
 	}
 
 	/**
-<<<<<<< HEAD
```
michaelkain, line 660: Hi Jean-Come, could you please remove the head comment? Thank you and with kind regards, Michael

michaelkain (approved):
Hi Jean-Come, very good! I approve by code review only and I will test it during our next test session on our qualif. The code looks completely logic. Perfect for now. I am wondering for the future, if it would be an option to arrange the processed datasets in a separate folder, as BIDS does with /derivatives, because a processed dataset from e.g. two datasets from two exams (M0 + M1) to which exam does it belong? to both, but we would duplicate the information, so maybe provide it in a separate folder. But perfect for the moment! With kind regards and have a good start into your week, Michael

PR #1518, opened 2023-01-30
vip processed dataset role management

alaeessaki:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/controller/CarminDatasetProcessingApiController.java
```diff
@@ -97,13 +105,32 @@ public ResponseEntity<Void> updateCarminDatasetProcessing(
 
     @Override
     public ResponseEntity<List<CarminDatasetProcessing>> findCarminDatasetProcessings() {
-        final List<CarminDatasetProcessing> carminDatasetProcessings = carminDatasetProcessingService.findAll();
+        final List<CarminDatasetProcessing> carminDatasetProcessings = carminDatasetProcessingSecurityService.filterCarminDatasetList(carminDatasetProcessingService.findAll(), RIGHT_STR);
```
michaelkain, line 108: Dear Alae, thank you for your PR. We normally try to put all SecurityServices into annotations, either PreAuthorize or PostAuthorize. In this case I would propose PostAuthorize. Normally the controller implementation should not need a reference to the SecService. As example:
@PostAuthorize("hasRole('ADMIN') or @datasetSecurityService.**filterDatasetDTOList**(returnObject.getBody(), 'CAN_SEE_ALL')")

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/controller/CarminDatasetProcessingApiController.java
```diff
@@ -37,6 +40,11 @@ public class CarminDatasetProcessingApiController implements CarminDatasetProces
     @Autowired
     private ExecutionStatusMonitorService executionStatusMonitorService;
 
+    @Autowired
```
michaelkain, line 43: Hi Alae, lines 43-47 could be completely removed if moved to annotation in CDProcApi.java

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/carmin/security/CarminDatasetProcessingSecurityService.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+@Service
+public class CarminDatasetProcessingSecurityService {
+
+    private static final String ROLE_ADMIN = "ROLE_ADMIN";
```
michaelkain, line 17: Hi Alae, if you declare the constant, it would be great using it below, or not declaring it. Thank you and with kind regards, Michael

michaelkain (changes requested):
Hi Alae, please see my 3 comments. Thank you and with kind regards, Michael

alaeessaki:
Hello Michael,
Thank you for your review.
I fixed those propositions. 
But for the security service, i use it to filter a list depending of if it has the right of the study, so i dont think a postAuthorize can help me in this case. the name is security service but the functionality is to filter data using the studysecurity service. 
You are right, it better to not use the security service inside the controller, i changed it to be called in the service layer. 
Please check my updates. 
With kind regards, Alae

PR #1521, opened 2023-01-31
Kubernetes [do not merge]

michaelkain:
The idea of this PR is to prepare and analyse, what is necessary to deploy a full Shanoir instance on using Kubernetes. First step: deploy shanoir-ng using minikube.
- kompose used on docker-compose.yml
- apiVersion replaced to apps/v1 for all deployments
- modifications on users-deployment.yaml, that correct errors


michaelkain:
I close, because dc.yml has been updated + use version with github packages references

PR #1527, opened 2023-02-01
Fix for problem of PH: by default go to public catalog

michaelkain:
Hi, my local tests work:
- https://shanoir-ng-nginx goes to /shanoir-ng/welcome
- when I click access, I go to the login page, when I cancel I come back


michaelkain:
to see how it behaves on qualif, for sure

jcomedouteau (approved):
It works for me on qualif !

PR #1529, opened 2023-02-02
Spring Boot 3.1.2

michaelkain:
- Spring Boot 3 + Java 17 migration, to prepare Spring Native usage
- Spring Data 3 has separated PagingAndSortingRepo from CrudRepo, what requires additional adaptations in our repositories code as e.g. methods like findById where missing
- updates to Junit 5, as new standard for testing in Spring Boot 3
- removes all old usage of JUnit 4, as legacy since some time
- changes all packages related to javax to jakarta
- updates (as forced) from swagger2 (not supporting Java 17) to swagger3 SpringDoc
- ms studies: removes Gson, as not needed because of Jackson and not compatible with Java 17
- ms studies: changes strange hierarchy of CenterServiceImpl because of duplicate security annotation exception with new version of SpringSecurity
- ms datasets: removal of SolrCrudRepository, spring-data-solr not deprecated and not compatible with Java 17
- ShanoirUploader version 8.0.0 (new Lombok version): change on study card compatibility check
- shanoir-ng-anonymization: update to poi version 4.1.2
- ms preclinical: changes column Reference.value to reference.refvalue, because "value" is keyword in H2Database and inserts with test-data-h2.sql did not work anymore
- removes keycloak-spring-boot-security, as deprecated, and migrates all to Spring Security 6 OAuth2 resource server usage
- removes usage of SolrTemplate and SolrCrudRepository (and most components of spring-data-solr) from shanoir-ng, except 5-10 classes used for facet search object construction, that do not conflict with Java 17 (I prefer this hybrid solution, instead of copying all these classes into shanoir-ng code base)
- updates keycloak client as well: 20.0.0 to 22.0.0, as in 20.0.0 resteasy dependency based on javax (not supported anymore)
- update to SpringDoc 2.2.0 and SecConfig to latest state of Spring Security in Spring Boot 3.1.2
- ShanoirUploader: complete refactor to remove dcm4che2 dependencies (all) and use dcm4che3 v5.30.0, as all mss
- ShanoirUploader: use all code from ms-import now and remove own code for importing data


michaelkain:
Just to keep trace of that:
- Spring Data Solr has been removed/is discontinued within Spring Data, see https://spring.io/blog/2020/04/07/spring-data-for-apache-solr-discontinued
- The official recommendation is to migrate to ElasticSearch (For current users of Spring Data Solr, we recommend looking into alternative search-engine-related modules like [Spring Data Elasticsearch](https://github.com/spring-projects/spring-data-elasticsearch).
- The problem: in MS Datasets, we use a version of Spring Data Solr, that is slightly older than the latest version. The new version (last update in 2021) requires an old version of spring-data-commons, that is not compatible anymore with the new version of spring-data-commons, that is in Spring Data of Spring Boot 3. What results in a ClassNotFoundException of SolrTemplate in Spring (Caused by: java.lang.ClassNotFoundException: org.springframework.data.convert.EntityInstantiators).
- I tried: I downgraded spring-data-commons to the version required by latest spring-data-solr, but then the new class ListCrudRepository of latest Spring Data is missing.
- Option A) Migrate to ElasticSearch (no idea, what that implies for the frontend and the backend)
- Option B) Move Solr classes into separate microservice and use old dependencies
- Option C) Manually add missing classes into MS Datasets to complete spring-data-solr with missing code

michaelkain:
Good progress today: now MS Preclinical compiles + JUnit tests run!

michaelkain:
Yes, 41 files merged with develop!

michaelkain:
Yes, all 5 microservices compile now and the tests run fully, after the merge. Only need to fix and reset my migration of dcm4che3 in ShUp, that I should not have started on that PR.

michaelkain:
Documentation:
On this PR I had to break and remove the schema in MS Studies: the usage of the BasicEntityService + -Impl., used by AcqEquipment, Coils, ManufacturerModel and Manufacturer.
For three reasons:

1) **The unit tests for the 4 entities on the ServiceTest, that are based on Mockito only did not work anymore**:
ManufacturerModelServiceTest.findAllTest:71 » NullPointer Cannot invoke "org.springframework.data.repository.CrudRepository.findAll()" because **"this.repository" is null**
It took me a lot of time to figure out why. Mockito is working fine and injects the Mock repository into the Service, BUT the second repository in the BasicEntityServiceImpl remained always NULL. Furthermore it looks weird, that in this hierarchy, two repositories are used and both annotated with @Autowired, what is ignored by Mockito for sure. I am really wondering how it worked before with Mockito.

2) Spring Boot 3 is based on Spring Security 6.1 and this has changed its way to process SecurityAnnotations like @PreAuthorize and now it throws the following exception, I had to fix:

AcquisitionEquipmentSecurityTest.testAsAdmin:97 This should return an AccessDeniedException but got a AnnotationConfigurationException at AuthorizationAnnotationUtils.java:65 - details : 
org.springframework.core.annotation.AnnotationConfigurationException: **Found more than one annotation of type interface org.springframework.security.access.prepost.PreAuthorize** attributed to public org.shanoir.ng.acquisitionequipment.model.AcquisitionEquipment org.shanoir.ng.acquisitionequipment.service.AcquisitionEquipmentServiceImpl.create(org.shanoir.ng.acquisitionequipment.model.AcquisitionEquipment) Please remove the duplicate annotations and publish a bean to handle your authorization logic.

The problem is the interface inheritance hierarchy, that lets Spring Security count one annotation with @PreAuthorize twice: 1. in the interface AcqEquipmentService + 2. in the interface BasicEntityService, see the following post on that:
https://stackoverflow.com/questions/75102645/annotationconfigurationexception-since-upgrade-to-spring-boot-3-jdk-17

3) This construct is okay to avoid some code duplication, but introduces, see above unnecessary code complexion, that leads to problems. I had already to work around during my last Spring Boot migration to 2.6.6.

michaelkain:
Entity graph traversal changes

To generate a query, Hibernate needs to traverse the entity graph and determine which entities should be included in the statement. During this traversal, it also tries to detect circular references. As it said in the [documentation](https://github.com/hibernate/hibernate-orm/blob/6.0/migration-guide.adoc#fetch-circularity-determination): “As back-ground, Hibernate does understand whether a fetch is actually, truly circular. It simply understands that while walking a fetch-graph, it encounters the same table/column(s) making up a particular foreign key. In this case, it simply stops walking the graph any deeper”.

We can traverse through entities using depth-first or width-first approaches. Hibernate 6 uses width-first in contrast to previous versions that used depth-first. This may affect query generation, which can be surprising for experienced Hibernate developers. Let’s consider the following entity:

public class Pet {
    @Id
    @Column(name = "id", nullable = false)
    private Long id;

    @Column(name = "name")
    private String name;

    @ManyToOne
    @JoinColumn(name = "parent_1_id")
    private Pet parent1;

    @ManyToOne
    @JoinColumn(name = "parent_2_id")
    private Pet parent2;
}

For this entity, let’s execute a simple query:

Pet pet = petRepository.findById(1L);

Now it is time to look at the generated SQL, especially at join clauses. For Hibernate 5, it will be the following:

select
…
    from
        pet pet0_ 
    left outer join
        pet pet1_ 
            on pet0_.parent_1_id=pet1_.id 
    left outer join
        pet pet2_ 
            on pet1_.parent_2_id=pet2_.id 
    where
        pet0_.id=?

Looks OK. It is a straightforward query with no surprises. Now let’s have a look at the SQL generated by Hibernate 6:

select
…
    from
        pet p1_0 
    left join
        pet p2_0 
            on p2_0.id=p1_0.parent_1_id 
    left join
        pet p3_0 
            on p3_0.id=p2_0.parent_2_id 
    left join
        pet p4_0 
            on p4_0.id=p1_0.parent_2_id 
    left join
        pet p5_0 
            on p5_0.id=p4_0.parent_1_id 
    where
        p1_0.id=?

We have four join’s here! For three recursive associations, there will be 15 join’s. This is how Hibernate developers explain this: “as you can see, this leads to a lot of joins very quickly, but the behavior of 5.x simply was not intuitive. To avoid creating so many joins, and also in general, we recommend that you use lazy fetching i.e. @ManyToOne(fetch = FetchType.LAZY) or @OneToOne(fetch = FetchType.LAZY) for most associations, but this is especially important if you have multiple self-referencing associations as you can see in the example”.

In the sixth version, Hibernate combines all possible joins and foreign keys in the query and stops only when it sees that the reference is circular. It becomes clear is we look at the example with three recursive associations:

FROM Node
JOIN Node.node1
JOIN Node.node1.node2
JOIN Node.node1.node2.node3
JOIN Node.node1.node3
JOIN Node.node1.node3.node2
JOIN Node.node2
JOIN Node.node2.node1
JOIN Node.node2.node1.node3
JOIN Node.node2.node3
JOIN Node.node2.node3.node1
JOIN Node.node3
JOIN Node.node3.node1
JOIN Node.node3.node1.node2
JOIN Node.node3.node2
JOIN Node.node3.node2.node1

So, in Hibernate 6, prefer the LAZY fetch type to avoid combination explosion for the recursive associations during queries.

michaelkain:
Hi @youennmerel, hi @pierrehenri-dauvergne, I think it is time to start the code review of this PR.
Please ignore shanoir-uploader, as tested with Java 17 with a CHU already.
@youennmerel: please review maybe MS Studies + Preclinical
@pierrehenri-dauvergne: please review maybe MS Users + Datasets

There are already some known issues, I will try to fix in the mean time, but we should bring this home asap. I prefer fixing on develop then producing an even bigger PR every day.

Known issues:
- The id generation is not perfect yet, it works but does not count up properly
- I have an issue with the list of exams (SQL with preclinical or something like that)
- Locally, the full indexation of Solr works, but I have an error when indexing datasets directly after the import (no exception, but the datasets do not appear in the search). Problem with sync of database transactions, I guess.

michaelkain:
We need to test just all. So, no worries. Imports work: I tested with MR and CT already.

michaelkain:
Hi, good news! Since I fixed and cleaned the identity generation issue, and now I am back using MySQL auto_increment, as in prod today, my issues are gone:
- the datasets are indexed, directly after the import
- the list of exams is displayed locally

michaelkain:
Bug solved with examinations list, related to HQL parsing evolution, see commit
Caused by: org.hibernate.query.sqm.ParsingException: line 1:57 no viable alternative at input 'selectexfromExaminationasexwhereex.preclinicalis?'
datasets  | 	at org.hibernate.query.hql.internal.StandardHqlTranslator$1.syntaxError(StandardHqlTranslator.java:46)
datasets  | 	at org.antlr.v4.runtime.ProxyErrorListener.syntaxError(ProxyErrorListener.java:41)

michaelkain:
Hi, I did some load testing and imported one phantom serie a 100 times with the creation of 100 subjects + exams + started an import in parallel. It worked fine locally and took only 55 seconds to import. And my local machine remained stable.

michaelkain:
Tested locally:
- OK, for mass imports
- OK, import webGUI MR+CT+Bruker
- OK, create and edit study
- OK, latest version of ShUp
- OK, mass indexation for Solr search
TO TEST:
- ShUp v7.0.1 in production for Ofsep: test with server 3.1.2

michaelkain (commented):
I have reviewed again ms-import, for me it is fine. Found nothing strange, at all. ShanoirUploader is and will have to be tested separately. With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -219,66 +214,60 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
...
-							modelName = serie.getMriInformation().getManufacturersModelName();
-							foundPatient = true;
-							currentDate = studyDate;
-							selectedStudyDate = studyDate;
-						}
-					}
-				}
-			}
-		}
+//		for (Patient patient : patients) {
```
youennmerel, line 217: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -333,15 +322,15 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
...
-			final String dicomBirthDate = pat.getDescriptionMap().get("birthDate");
-			if (!StringUtils.isEmpty(dicomBirthDate)) {
-				Date dicomBirthDateAsDate = Util.convertStringDicomDateToDate(dicomBirthDate);
-				Calendar cal = Calendar.getInstance();
-				cal.setTime(dicomBirthDateAsDate);
-				cal.set(Calendar.MONTH, Calendar.JANUARY);
-				cal.set(Calendar.DAY_OF_MONTH, 1);
-				dicomData.setBirthDate(cal.getTime());
-			}
+//			final String dicomBirthDate = pat.getDescriptionMap().get("birthDate");
```
youennmerel, line 325: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -363,9 +352,9 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
 		 */
 		logger.info("5 Fill MRI info");
 
-		for (Iterator<Serie> iterator = selectedSeries.iterator(); iterator.hasNext();) {
-			Serie serie = iterator.next();
-			Util.processSerieMriInfo(uploadFolder, serie);
+		for (Iterator<SerieTreeNode> iterator = selectedSeries.iterator(); iterator.hasNext();) {
+			SerieTreeNode serie = iterator.next();
+// @todo			Util.processSerieMriInfo(uploadFolder, serie);
```
youennmerel, line 357: Is this still "to do" ?

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -405,7 +394,7 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
 		Subject subjectFound = null;
 		String subjectStudyIdentifier = null;
 		try {
-			subjectFound = shanoirUploaderServiceClientNG.findSubjectBySubjectIdentifier(subjectIdentifier);
+//			subjectFound = shanoirUploaderServiceClientNG.findSubjectBySubjectIdentifier(subjectIdentifier);
```
youennmerel, line 397: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -417,29 +406,29 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
 		if (subjectFound != null) {
 			logger.info("8 Subject exists, just use it");
 			subject = subjectFound;
-			ImportUtils.addSubjectStudy(study2, subject, SubjectType.PATIENT, true, subjectStudyIdentifier);
-			subject = shanoirUploaderServiceClientNG.createSubjectStudy(subject);
+//			ImportUtils.addSubjectStudy(study2, subject, SubjectType.PATIENT, true, subjectStudyIdentifier);
```
youennmerel, line 409: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -417,29 +406,29 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
...
 	
 			subject = new Subject();
 			subject.setName(csvImport.getCommonName());
-			if (!StringUtils.isEmpty(pat.getDescriptionMap().get("sex"))) {
-				subject.setSex(Sex.valueOf(pat.getDescriptionMap().get("sex")));
-			} else {
-				// Force feminine (girl power ?)
-				subject.setSex(Sex.F);
-			}
+//			if (!StringUtils.isEmpty(pat.getDescriptionMap().get("sex"))) {
```
youennmerel, line 416: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -417,29 +406,29 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
...
 	
-			subject.setImagedObjectCategory(ImagedObjectCategory.LIVING_HUMAN_BEING);
-	
-			subject.setBirthDate(dicomData.getBirthDate().toInstant().atZone(ZoneId.systemDefault()).toLocalDate());
-	
-			ImportUtils.addSubjectStudy(study2, subject, SubjectType.PATIENT, true, subjectStudyIdentifier);
-	
-			// Get center ID from study card
-			subject = shanoirUploaderServiceClientNG.createSubject(subject, true, centerId);
+//			subject.setImagedObjectCategory(ImagedObjectCategory.LIVING_HUMAN_BEING);
```
youennmerel, line 424: If this part is not used anymore, better remove it from the code base

shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromCsvRunner.java
```diff
@@ -471,7 +460,7 @@ private boolean importData(CsvImport csvImport, List<StudyCard> studyCardsByStud
 		logger.info("10 Import.json");
 
 		ImportJob importJob = ImportUtils.prepareImportJob(uploadJob, subject.getName(), subject.getId(), createdExam.getId(), study2, sc);
-		importJob.setFromCsv(true);
+		importJob.setFromShanoirUploader(true); // @todo: set from csv here for upload
```
youennmerel, line 463: is this still "to do" ?

shanoir-uploader/src/main/java/org/shanoir/uploader/action/RSDocumentListener.java
```diff
@@ -14,33 +14,19 @@ public RSDocumentListener(final MainWindow mainWindow) {
...
-					& (mainWindow.mSexR.isSelected() || mainWindow.fSexR.isSelected())
-					& mainWindow.isDicomObjectSelected == true)
-				mainWindow.downloadOrCopyButton.setEnabled(true);
-		} else {
-			if (mainWindow.newPatientIDTF.getText().length() != 0 & mainWindow.birthDateTF.getText().length() != 0
-					& (mainWindow.mSexR.isSelected() || mainWindow.fSexR.isSelected())
-					& mainWindow.isDicomObjectSelected == true)
-				mainWindow.downloadOrCopyButton.setEnabled(true);
-		}
+		if (mainWindow.lastNameTF.getText().length() != 0 & mainWindow.firstNameTF.getText().length() != 0
```
youennmerel, line 17: I think this long condition should be stored in a well named variable (e.g. `isFormComplete`)

shanoir-uploader/src/main/java/org/shanoir/uploader/action/RSDocumentListener.java
```diff
@@ -14,33 +14,19 @@ public RSDocumentListener(final MainWindow mainWindow) {
...
+				& mainWindow.birthNameTF.getText().length() != 0 & mainWindow.birthDateTF.getText().length() != 0
+				& (mainWindow.mSexR.isSelected() || mainWindow.fSexR.isSelected())
+				& mainWindow.isDicomObjectSelected == true)
+			mainWindow.downloadOrCopyButton.setEnabled(true);
 	}
 
 	public void removeUpdate(DocumentEvent e) {
-		if (mainWindow.noAnonR.isSelected()) {
-			if (mainWindow.lastNameTF.getText().length() == 0 || mainWindow.firstNameTF.getText().length() == 0
```
youennmerel, line 33: I think this long condition should be stored in a well named variable (e.g. `isFormComplete`)

shanoir-uploader/src/main/java/org/shanoir/uploader/action/RSDocumentListener.java
```diff
@@ -14,33 +14,19 @@ public RSDocumentListener(final MainWindow mainWindow) {
...
-				|| mainWindow.birthNameTF.getText().length() == 0 || mainWindow.birthDateTF.getText().length() == 0
-				|| !(mainWindow.mSexR.isSelected() || mainWindow.fSexR.isSelected())
-				|| mainWindow.isDicomObjectSelected == false)
-			mainWindow.downloadOrCopyButton.setEnabled(false);
-		} else {
-			if (mainWindow.newPatientIDTF.getText().length() == 0 || mainWindow.birthDateTF.getText().length() == 0
```
youennmerel, line 39: I think this long condition should be stored in a well named variable (e.g. `isFormComplete`)

shanoir-uploader/src/main/java/org/shanoir/uploader/action/SelectionActionListener.java
```diff
@@ -56,72 +56,68 @@ public void valueChanged(TreeSelectionEvent event) {
...
 		// returns all selected paths, which can be patients, studies and/or series
 		TreePath[] paths = mainWindow.dicomTree.getSelectionModel().getSelectionPaths();
 		if (paths != null && paths.length > 0) {
 			for (int i = 0; i < paths.length; i++) {
 				TreePath tp = paths[i];
 				Object o = tp.getLastPathComponent();
-				Patient patient = null;
+				PatientTreeNode patient = null;
 				// handle if patient in paths has been found
-				if (o instanceof Patient) {
```
youennmerel, line 92: It seems in Java 17 you can now switch case with instanceof (and avoid explicit casting) : https://www.baeldung.com/java-switch-pattern-matching#1-type-pattern

shanoir-uploader/src/main/java/org/shanoir/uploader/dicom/DicomServerClient.java
```diff
@@ -39,69 +37,67 @@ public class DicomServerClient implements IDicomServerClient {
...
 	 * @see org.shanoir.uploader.dicom.IDicomServerClient#echoDicomServer()
 	 */
 	@Override
 	public boolean echoDicomServer() {
-		Echo echo = new Echo();
-		boolean success = echo.echo(config);
-		logger.info("Echoing of the DICOM server was successful? -> " + success);
-		return success;
+		// @todo add echo here
+		logger.info("Echoing of the DICOM server was successful? -> ");
```
youennmerel, line 61: Is this still "To do" ?

shanoir-ng-users/src/main/java/org/shanoir/ng/accessrequest/controller/AccessRequestApi.java
```diff
@@ -19,107 +19,108 @@
...
 
-	@ApiOperation(value = "", notes = "Saves a new access request", response = AccessRequest.class, tags = {})
+	@Operation(summary = "", description = "Saves a new access request")
 	@ApiResponses(value = {
-			@ApiResponse(code = 200, message = "created access request", response = AccessRequest.class),
-			@ApiResponse(code = 401, message = "unauthorized", response = AccessRequest.class),
-			@ApiResponse(code = 403, message = "forbidden", response = AccessRequest.class),
-			@ApiResponse(code = 422, message = "bad parameters", response = AccessRequest.class),
-			@ApiResponse(code = 500, message = "unexpected error", response = AccessRequest.class) })
+			@ApiResponse(responseCode = "200", description = "created access request"),
```
youennmerel, line 39: Shouldn't we still define the response type, so that the schema is displayed in swagger ?
Or is it now deducted in swagger v3 ?

youennmerel (commented):
Hi Michael,

Some comments. 
I got carried away and comments some of shanoir uploader, but maybe this part is not yet finished on your side.

Cheers.

shanoir-ng-users/src/main/java/org/shanoir/ng/accessrequest/controller/AccessRequestApi.java
michaelkain: swagger v3 removed the response type

michaelkain:
swagger3 works on qualif
image

pierrehenri-dauvergne:
ms-studies OK

michaelkain:
all microservices are now running on qualif, could validate acc requests
I could finish an import, that is now indexed to the search as well

youennmerel:
OK for ms-users

PR #1543, opened 2023-02-15
Fix mysql container hanging at startup

a-ba:
As of mysql/mysql-server:5.7.34 `/mysql-init-complete` was moved into the `/var/lib/mysql-files/` directory.


michaelkain (approved):
Thank you Anthony! Great job!

PR #1545, opened 2023-02-15
#1544 - access request - allow user deletion & access request

### LINKED ISSUE #1544: [prod] Access request: issues
Hi Jean-Come,
thank you, I can now validate or refuse extension requests as admin, what is fine. Great!
But I am having issues on the prod with some access requests. I think I can validate new
user access requests, e.g. made on one of the challenges. What does not seem to work:
- existing user, already in one study (e.g. id 209), ask for access to study id 208, failure!
/accessrequest/resolve/2 500, when I click on "accept" as study admin (mkain account)
and after the access request is gone, what is weird too
- I had the same issue for some reason, with a newly created account too, same 500
- as sys admin I can not delete anymore users, that are duplicated
- in the email sent to the study admin, after I added a new member to my study, I do not
have my account, with my email in the confirmation mail, but shanoir-neurinfo-adm@lists.gforge.inria.fr
I will send you the screens, as I do not want to share them publicly because of user infos.
With kind regards, Michael

jcomedouteau:
Closes #1544 
Hi @michaelkain, I don't really get how we arrived to this.
- Users impossible deletion was quite logic -> They are still linked to one ore more access request. I updated the logic so that, when a user is deleted, associated access requests are deleted.

For the second point, it's quite misterious, we have user with "account_request_on_demand" set to NULL, as they should be set to either "true" (non existing user) or "false" (existing user).
For the already existing user, I can imagine that the "on demand" could be set to null, and I added a check, but for non existing user, I cannot understand how we arrived in that state..
I'll check a little bit further..
Jean-Côme

### DIFF
--- a/shanoir-ng-front/src/app/studies/study/study.component.html
+++ b/shanoir-ng-front/src/app/studies/study/study.component.html
@@ -203,10 +203,10 @@ along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 				</ol>
 			</fieldset>
 			<fieldset [class.hidden]="activeTab != 'general'">
-				<legend i18n="Study detail|Name label@@studyDetailDefaultAccessLevel">Default access level</legend>
+				<legend i18n="Study detail|Name label@@studyDetailDefaultAccessLevel">Access level</legend>
 				<ol>
 					<li>
-						<label i18n="Study detail|Visible by default label@@studyDetailVisibleByDefault">Visible by default</label>
+						<label i18n="Study detail|Visible by default label@@studyDetailVisibleByDefault">Visible</label>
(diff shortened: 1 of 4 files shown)


jcomedouteau:
I'm not really sure I got this point:
in the email sent to the study admin, after I added a new member to my study, I do not
have my account, with my email in the confirmation mail, but [shanoir-neurinfo-adm@lists.gforge.inria.fr](mailto:shanoir-neurinfo-adm@lists.gforge.inria.fr)

The email was sent to neurinfo-adm@lists.gforge.inria.fr ?

michaelkain (approved):
Hi Jean-Come, thank you! I will have to test on qualif.

PR #1546, opened 2023-02-16
shanoir-issue#1502-4: remove unused dependencies

pierrehenri-dauvergne:


shanoir-ng-front/src/app/examinations/shared/examination.service.ts
```diff
@@ -25,6 +25,7 @@ import { ExaminationDTO, ExaminationDTOService } from './examination.dto';
 import { ServiceLocator } from '../../utils/locator.service';
 import { LoadingBarComponent } from '../../shared/components/loading-bar/loading-bar.component';
 import { Subscription } from 'rxjs'
+import {saveAs} from "file-saver";
```
julien-louis, line 28: What is it for ? I might be wrong but it seems that those 2 ts have been unintentionally commited.

julien-louis (changes requested):
If you commit updates, just do it in your local branch pierrehenri-dauvergne:shanoir-issue#1502-4  and push it.

julien-louis (approved):

PR #1550, opened 2023-02-23
oneshot: allow signalling the whole process group

a-ba:


julien-louis:
note for later : add --pgrp to oneshot call in bootstrap.sh

PR #1567, opened 2023-03-07
#1544 Delete user with access request + Search by username

jcomedouteau:
3 things to test:
1) Create a new user with an access request on a study
Refuse this access request -> the user is well deleted

2) Create a new user with an access request on a study
Directly delete this user -> The user is well deleted

3) From study detail, add a user from username instead of email


michaelkain (approved):
Dear Jean-Come, thank you for this fix! I approve by code review only and add it on my little list to test on qualif. Looking forward to bring this live! With kind regards, Michael

PR #1574, opened 2023-03-09
VIP URL into .env (to directly connect via CARMIN client)

michaelkain:
Hi all,
the PR, that will connect the CARMIN-API client inside the Angular frontend in Shanoir with VIP, without a reverse proxy. The colleagues from VIP will have to add the CORS + Keycloak configuration, but it should be fine like that.
@a-ba to inform you about that "connection"
@youennmerel please review and test on your machine
For testing: Solr dataset search, select one dataset and Run process and verify, that you get a CORS error from VIP.
With kind regards, Michael


jcomedouteau:
@michaelkain 
I think you also have to modify application.yml from datasets microservice, so that the backend also has access to the VIP rest API

michaelkain:
Dear Jean-Come (@jcomedouteau), very good comment! I added the backend configuration as well to MS Datasets to unify the configuration at one place, by the way, the big advantage of docker-compose to trace all these dependencies. With kind regards, Michael

youennmerel:
Hi @michaelkain,

I've tested your branch on my machine, and I get the CORS error as expected :
image

michaelkain:
Dear Youenn @youennmerel, thank you! It would be great to please approve my PR in that case. With kind regards, Michael

youennmerel (approved):

youennmerel:
@michaelkain build successful !

PR #1594, opened 2023-03-20
shanoir-issue#1588: fixed center field while creating a new study + …

pierrehenri-dauvergne:
…css in case of long string that goes over the edge of the dropdown menu


michaelkain:
@julien-louis Hi Julien, it would be great to review this PR. Thank you and with kind regards, Michael

michaelkain:
Hi Pierre-Henri (@pierrehenri-dauvergne), there seems to be a migration script error on the PR? 4 errors (new migrations must have a greater sequence number than all existing migrations in origin/master)
Error: Process completed with exit code 1. Let's discuss if this PR should be part of the new release? Thank you and with kind regards, Michael

pierrehenri-dauvergne:
Hi @michaelkain it should be better now :)

jcomedouteau:
Hey PHD, I don't think that it was in the issue initially but I'm taking advantage of that pull request:
- If the name is really long, could we display the full name when hovering (we see the three dots at the moment, but we can never see the full name)
- When the name is too long, I cannot click anymore on the "+" button (in multi-center)

image

(with a short name center, I can)
image

julien-louis:
@pierrehenri-dauvergne I think the issue with the select option length has been fixed elsewhere could you check on qualif ?

pierrehenri-dauvergne:
> @pierrehenri-dauvergne I think the issue with the select option length has been fixed elsewhere could you check on qualif ?

If it was fixed, it wasn't merged yet. This is on qualif :
image

jcomedouteau:
I think some conflict resolution has to be done.

pierrehenri-dauvergne:
> I think some conflict resolution has to be done.

Done

jcomedouteau (approved):

PR #1597, opened 2023-03-21
[Fli-iam#248] Dataset sizes

youennmerel:
Added total size of datasets in study details ("Storage volume")

image

- sizes are stored in table `dataset_expression.size`
- if no dataset is associated to the study, will display `0 Bytes`
- if at least one dataset size is not known (preexisting dataset), will display `Unknown`
- will adapt unit to size value (KB, MB, GB, etc)


youennmerel:
to correct : Even if not all dataset volume are known, display the partially calculated volume

shanoir-ng-front/src/app/studies/study/study.component.ts
```diff
@@ -217,6 +234,30 @@ export class StudyComponent extends EntityComponent<Study> {
...
+    return this.datasetService.getByStudyId(id).then(datasets => {
+
+      let promises: Promise<any>[] = [];
+
+      if(datasets == null || datasets.length == 0){
+        return 0;
+      }
+
+      let totalSize: number = 0;
+      for (let ds of datasets) {
```
jcomedouteau, line 248: Maybe it was how we wanted to design this, but I'm wondering if it's a good idea to make X rest calls for every concerned dataset.
We may make one single call on the study ID and then calculate in the back (?)

The other workaround would be that, if you add size to datasetExpression in the front classes, it will be retrieved automatically with the first call, and you can make the calculation in the front directly.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/DatasetExpression.java
```diff
@@ -57,9 +57,13 @@ public class DatasetExpression extends AbstractEntity {
 	@JoinColumn(name = "dataset_id")
 	private Dataset dataset;
 
+
 	/** Dataset expression format. */
 	private Integer datasetExpressionFormat;
 
+	/** size of the associated dataset files. */
+	private Long size;
```
jcomedouteau, line 65: If you add a field, an SQL script to add the column is necessary for the qualif / production server. (but not locally)
Please see what is done in ../shanoir-ng/docker-compose/database/db-changes/datasets

jcomedouteau:
@youennmerel A small review, we can discuss it if you want !
I trust you for the code refactoring, thanks for taking the time to do this :)

shanoir-ng-front/src/app/studies/study/study.component.ts
youennmerel: I'm not very satisfied with my design either. I'll discuss it with you this afternoon if you're available.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/DatasetExpression.java
youennmerel: All right, i'll add it !

youennmerel:
@jcomedouteau @michaelkain I've just pushed my corrections.

michaelkain:
Hi Jean-Come (@jcomedouteau), could you please approve again, if okay for you? I can then merge, e.g. for test sessions. With kind regards, Michael

jcomedouteau (approved):

Issue #1622, opened 2023-03-31
Import BIDS isn't working


pierrehenri-dauvergne:
Click on "import BIDS"
Select a study and a center
Click on "choose file"
Select file
An error appears on screen: 
![image](https://user-images.githubusercontent.com/117283961/229128352-041a543a-ec1c-4b15-b410-8304c47c114f.png)

PR #1629, opened 2023-03-31
shanoir-issue#1622: fix bug where bids import weren't working

pierrehenri-dauvergne:
Related issue : https://github.com/fli-iam/shanoir-ng/issues/1622


jcomedouteau (approved):
Subject id is not defined here, but it's not important.

PR #1634, opened 2023-04-04
shanoir-issue#1625: fix bug - allow only centers of the study for whi…

pierrehenri-dauvergne:
…ch the acquisition equipment is created
To test :
 - Create a mono-center study
 - Create a multi-center study
 - Create a new study card for the mono-center study (it doesn't have any acquisition equipment)
     - Create an acquisition equipment : available center is the one of the study
 - Do the same with the multi-center study: when creating an acquisition equipment, the centers available are those of the study
 
image
image


jcomedouteau:
Is this one reviewable @pierrehenri-dauvergne  ?

jcomedouteau (changes requested):
As discussed, we should use breadcrumds logic instead of context state in this case

julien-louis (approved):
tested

PR #1639, opened 2023-04-06
fix keycloak build-time config

a-ba:
8df648cc1610 removed code which was not actually dead.

A small list of keycloak variables have to be set at runtime *and* at buildtime. If not, then the quarkus image is dynamically rebuilt at each startup, which currently cause a 3-minute delay.


jcomedouteau (approved):
@julien-louis tu peux merge siouplait ?

PR #1640, opened 2023-04-07
shanoir-issue#1635: move schema.org DataCatalog to welcome page (used to be in login page)

pierrehenri-dauvergne:


michaelkain:
to be more clear: "SHANOIR_URL_SCHEME://SHANOIR_URL_HOST" will not be replaced after your move of code, from my point of view

michaelkain (changes requested):
See my comment

michaelkain (approved):

PR #1641, opened 2023-04-07
Arthur stat pr 1618

julien-louis:
Cherry picked arthur commits to merge them in master.

The PR was also merged in develop.


julien-louis (commented):
I checked that it's identical to the pr #1618 to develop

Issue #1617, opened 2023-03-31
[Quality Control] - review


jcomedouteau:
- Remove "re-apply study card" button from quality control detail.
- Edit study - select quality tab, no quality card existing, Click on Edit study, the screen is "empty": propose a button "create quality card"
- Rename quality report to "error report" and be clear, that valid results are not in report

jcomedouteau:
java.lang.NullPointerException: null
	at org.shanoir.ng.dataset.security.DatasetSecurityService.hasRightOnSubjectStudies(DatasetSecurityService.java:165)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.expression.spel.support.ReflectiveMethodExecutor.execute(ReflectiveMethodExecutor.java:139)
	at org.springframework.expression.spel.ast.MethodReference.getValueInternal(MethodReference.java:139)
	at org.springframework.expression.spel.ast.MethodReference.access$000(MethodReference.java:55)
	at org.springframework.expression.spel.ast.MethodReference$MethodValueRef.getValue(MethodReference.java:386)
	at org.springframework.expression.spel.ast.CompoundExpression.getValueInternal(CompoundExpression.java:92)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:188)
	at org.springframework.expression.spel.ast.OpOr.getBooleanValue(OpOr.java:56)
	at org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:47)
	at org.springframework.expression.spel.ast.OpOr.getValueInternal(OpOr.java:37)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:188)
	at org.springframework.expression.spel.ast.OpAnd.getBooleanValue(OpAnd.java:57)
	at org.springframework.expression.spel.ast.OpAnd.getValueInternal(OpAnd.java:52)
	at org.springframework.expression.spel.ast.SpelNodeImpl.getValue(SpelNodeImpl.java:188)
[...]

PR #1652, opened 2023-04-13
Quality round 2

julien-louis:
* fix the issue list : https://github.com/fli-iam/shanoir-ng/issues/1617
* provides quality card testing (applying with no database update)
* provides quality cards auto checking at import
* new import notifications (with Jobs)
* job details on click in table


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
```diff
@@ -26,18 +26,18 @@
...
 				// In case of Enhanced MR, we need to the pixel data in order to use Dcm4chee emf extract method.
-				datasetAttributes = dIS.readDataset(-1, -1);
+				datasetAttributes = dIS.readDataset(-1);
 			} else {
 				// Else we do not load the picture in Ram for faster performance.
-				datasetAttributes = dIS.readDataset(-1, Tag.PixelData);
+				datasetAttributes = dIS.readDataset(Tag.PixelData);
 			}
-			
+			dIS.close();
```
jcomedouteau, line 40: This is not necessary with "try with ressource" as done:
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
```diff
@@ -26,18 +26,18 @@
...
 			Attributes datasetAttributes;
-			if (isEnhancedMR) {
+			if (isEnhancedMR != null && isEnhancedMR) {
 				// In case of Enhanced MR, we need to the pixel data in order to use Dcm4chee emf extract method.
-				datasetAttributes = dIS.readDataset(-1, -1);
+				datasetAttributes = dIS.readDataset(-1);
 			} else {
 				// Else we do not load the picture in Ram for faster performance.
-				datasetAttributes = dIS.readDataset(-1, Tag.PixelData);
+				datasetAttributes = dIS.readDataset(Tag.PixelData);
```
jcomedouteau, line 38: I'm not convinced this will work the same -> The second argument is a "stop to this tag" argument.
I think you can use this method: readDatasetUntilPixelData
(if existing on our version)

shanoir-ng-front/src/app/import/pre-clinical-context/pre-clinical-context.component.ts
```diff
@@ -87,6 +87,7 @@ export class PreClinicalContextComponent extends AbstractClinicalContextComponen
         importJob.acquisitionEquipmentId = context.acquisitionEquipment.id;
         importJob.converterId = context.niftiConverter.id;
         importJob.archive = contextImportJob.archive;
+        importJob.timestamp = timestamp;
```
jcomedouteau, line 90: Maybe initialize timestamp here ? (same for basic-clinical-context.component.ts, maybe code refactoring is possible ?)

docker-compose/database/db-changes/users/0003_increase_task_msg.sql
```diff
@@ -0,0 +1,2 @@
+alter table events modify message text;
```
jcomedouteau, line 1: I still don't really understand what is the purpose of this timestamp ?

jcomedouteau:
Hey @julien-louis ! 
Just some small remarks, i'll test this afternoon! :)
Could you correct the conflicts too? And in a general matter, I think the indentation in java file has to be changed in your IDE I think (or maybe add save changes ?)
Thanks !
Jean-Côme

jcomedouteau:
I'm testing, it works quite well :)
It seems that there is no way to see the full message from the job panel:
image

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
julien-louis: from the source code : 

```
    /**
     * @deprecated Use one of the other {@link #readDataset()} methods instead. If you want to
     * specify a length limit, you may supply a {@link LimitedInputStream} or use
     * {@link #createWithLimit} or {@link #createWithLimitFromFileLength}.
     */
    @Deprecated
    public Attributes readDataset(int len, int stopTag) throws IOException {
        return readDataset(len, tagEqualOrGreater(stopTag));
    }

    public Attributes readDataset(int stopTag) throws IOException {
        return readDataset(tagEqualOrGreater(stopTag));
    }

    public Attributes readDataset(Predicate<DicomInputStream> stopPredicate) throws IOException {
        return readDataset(UNDEFINED_LENGTH, stopPredicate);
    }

    ...

    private static final int UNDEFINED_LENGTH = -1;
```

docker-compose/database/db-changes/users/0003_increase_task_msg.sql
julien-louis: I could remove it. It was to link a job with an import after having launching it.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/dto/ImportJob.java
```diff
@@ -69,6 +71,14 @@ public class ImportJob implements Serializable {
 	
 	private Long userId;
 	
+	public long getTimestamp() {
+        return timestamp;
+    }
+
```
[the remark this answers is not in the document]
julien-louis: those are because the file is indented with tabs, I'm replacing them with spaces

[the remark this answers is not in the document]
julien-louis: It seems to be the case in files written by atoboul

docker-compose/database/db-changes/users/0003_increase_task_msg.sql
julien-louis: I will remove it

jcomedouteau:
Je me souviens plus trop de ce qu'on avait dit, est-ce que le fait de repasser une quality-card fait le "ménage" des autres tags ?

jcomedouteau:
Du coup en effet quand j'ai testé, un nouveau round de quality ne clean pas les anciens tags.

jcomedouteau:
Hey @julien-louis !

I think i finished the review, the only point I see for the moment, is the tag cleaning after a new quality round.
Otherwise, everything works well, thanks !

Jean-Côme

jcomedouteau (approved):

jcomedouteau:
Hi  @julien-louis, can you merge develop so that we can merge this one ?

julien-louis:
shit

PR #1653, opened 2023-04-13
shanoir-issue#1638: bugfix when importing DICOM, removing studycard would block the import. Also fixes center value in studycard details

pierrehenri-dauvergne:


jcomedouteau:
I'm testing it

jcomedouteau (approved):

Issue #1660, opened 2023-04-17
Download limit to 500 datasets


jcomedouteau:
Just to try on qualif.

michaelkain:
Wait for your test feedback all together on qualif.

michaelkain:
Implemented, now to deploy

PR #1661, opened 2023-04-17
#1660 - dataset download limit to 500

jcomedouteau:
Test for download with higher datasets limit


jcomedouteau:
Works on qualif, we need to monitor RAM / memory usage / disk space (for dicoms more importantly)

michaelkain:
Dear @jcomedouteau, as Anthony has not had any opposition to that extension, I propose to test it. First next week in a test session with the other developers (without me), to see if multiple downloads are started, that the server remains stable. And bring it live next week as well? With kind regards, Michael

PR #1674, opened 2023-04-19
#1668 - invitation as expert bug

### LINKED ISSUE #1667: [Access request] Remove Accept / Refuse access request from creation

### LINKED ISSUE #1668: [Study access] Add user to study using its login as expert
isMeSecurityService.isMe(userName)  -> Cannot work properly as an expert

jcomedouteau:
- Add specific method for invitation to get user by username

Closes #1668 

How to test:
- Connect as an expert
- Go to an administrable study detail
- Edit it to add a member using its username (and not mail)
--> The user is well added

### DIFF
--- a/shanoir-ng-users/src/main/java/org/shanoir/ng/accessrequest/controller/AccessRequestApiController.java
+++ b/shanoir-ng-users/src/main/java/org/shanoir/ng/accessrequest/controller/AccessRequestApiController.java
@@ -194,7 +194,7 @@ public class AccessRequestApiController implements AccessRequestApi {
 		Optional<User> user = this.userService.findByEmail(email);
 		
 		if (!user.isPresent()) {
-			user = this.userService.findByUsername(email);
+			user = this.userService.findByUsernameForInvitation(email);
 		}
 
 		// User exists => return an access request to be added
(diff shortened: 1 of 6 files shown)


youennmerel:
Hi @jcomedouteau,

Adding a user by his login is OK for me in dev.

When trying to add a non existing login or mail, the following error occurs : "An internal error occured while adding the user. Please try again later or contact an administrator. "

I think we should rather explicitly indicate that the chosen username / email does not exists in this case.

However, this is absolutely non blocking, and we can merge anyway if time is of the essence !

shanoir-ng-users/src/main/java/org/shanoir/ng/accessrequest/controller/AccessRequestApiController.java
```diff
@@ -194,7 +194,7 @@ public 	ResponseEntity<AccessRequest> inviteUserToStudy(
 		Optional<User> user = this.userService.findByEmail(email);
 		
 		if (!user.isPresent()) {
-			user = this.userService.findByUsername(email);
+			user = this.userService.findByUsernameForInvitation(email);
```
youennmerel, line 197: Maybe rename the "email" variable ?

jcomedouteau:
Actually the login should not contain any '@' char, so I was thinking of just checking it..

michaelkain:
Hi, one option could be to make it the other way round, check for a valid email and try otherwise, see
https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/EmailValidator.html Just an idea.

youennmerel (approved):
Hi @jcomedouteau,

Tested in qualif, it seems to work all right.

PR #1684, opened 2023-04-21
[fli-iam#1678] set expire to 30m

youennmerel:


youennmerel:
It seems to work on qualif (no white page after redeploy).

I suggest to merge it on develop, and see if we still encounter the white page in the coming days.

However, note that you will need a first cache refresh, as the files were cached for 7d before the change.

jcomedouteau (approved):

Issue #1689, opened 2023-04-25
Constraint error when deleting a study card is not properly caught


youennmerel:
When deleting a study card, an SQL constraint error occurs if linked to a dataset acquisition, but the displayed error is a generic one. 

The  API should return a 422 HTTP error, and a message of the type "...is linked to other entities, it was not deleted." should be displayed on front side, as already done for some others entities.

Issue #1688, opened 2023-04-25
Can delete an equipment even if attached to a study card


youennmerel:
Deleting an acquisition equipment is possible even if said equipment is linked to a study card.

As a result, the study card are still linked to the equipment, and fails to be displayed correctly.

This will also block import (all field disabled) if the default selected study is concerned.

 However, this does not seems to be a problem for equipment attached to centers.

PR #1690, opened 2023-04-25
[Fli iam#1474] Delete from tree nodes

youennmerel:
From tree displayed in study and subject details, the following elements can be deleted via the node dropdown menu :
- acquisition
  - examination
    - dataset
      - processing
        - processed dataset
- equipment
- study card

The "delete" option is displayed only if the authenticated user is allowed to do so (see https://github.com/fli-iam/shanoir-ng/wiki/Roles-and-rights-specification#synthesis-table)

Note that two bug related to equipment and study card deletion are currently open :
- https://github.com/fli-iam/shanoir-ng/issues/1688
- https://github.com/fli-iam/shanoir-ng/issues/1689


shanoir-ng-front/src/app/examinations/tree/examination-node.component.ts
```diff
@@ -161,25 +165,44 @@ export class ExaminationNodeComponent implements OnChanges {
...
+            processing.outputDatasets ? processing.outputDatasets.map(ds => this.mapDatasetNode(ds, true)) : [],
+            this.node.canDelete
         );
     }
+
+    deleteExamination() {
+        this.examinationService.get(this.node.id).then(entity => {
+            this.examinationService.deleteWithConfirmDialog(this.node.title, entity).then(deleted => {
+                if (deleted) {
+                    console.log("DELETED !")
```
jcomedouteau, line 197: Remove logs

jcomedouteau:
Makes me sad to duplicate the code as such, but it's working fine for me !
Just remove the logs and we are good !

shanoir-ng-front/src/app/examinations/tree/examination-node.component.ts
youennmerel: :hand_over_mouth:

youennmerel:
> Makes me sad to duplicate the code as such, but it's working fine for me ! Just remove the logs and we are good !

Yeah, it would be great to have more generic components and inherited methods for tree nodes...
Maybe I could have added a NodeComponent class with common methods inherited by all nodes ? 

That being said, I've removed the log.

PR #1691, opened 2023-04-25
fix getStatistics procedure by rewriting the query

arthursw:
Matches the original version: 
Studies are fetched from examination.study_id instead of dataset.study_id, and centers from examination.center_id.

For some datasets, dataset.study_id is null whereas the corresponding examination.study_id is not null!


youennmerel (changes requested):
Hi @arthursw,

You should create a new script in `docker-compose/database/db-changes/datasets` (e.g. `docker-compose/database/db-changes/datasets/0025_fix_stats_procedure.sql`), and revert your changes on this file.

youennmerel (approved):
OK on OFSEP & Neurinfo qualifs

PR #1696, opened 2023-04-27
[fli-iam#1695] Manage listing including related datasets

youennmerel:
Listing examinations on a study with related examinations from another study on which the user does not have rights produced an error.

Happened on examinations listing in dataset import and study tree.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
@@ -78,12 +78,6 @@ public class ExaminationApiController implements ExaminationApi {
 	@Autowired
 	private ShanoirEventService eventService;
 
-	@Autowired
-	private StudyRepository studyRepository;
-
-	@Autowired
-	private StudyRightsService rightsService;
-
 	private final HttpServletRequest request;
 
 	@org.springframework.beans.factory.annotation.Autowired
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
@@ -14,24 +14,8 @@
 
 package org.shanoir.ng.examination.controler;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.time.LocalDate;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import javax.validation.Valid;
-
+import io.swagger.annotations.ApiParam;
 import org.apache.commons.io.FileUtils;
-import org.shanoir.ng.dataset.model.Dataset;
 import org.shanoir.ng.datasetacquisition.model.DatasetAcquisition;
 import org.shanoir.ng.examination.dto.ExaminationDTO;
 import org.shanoir.ng.examination.dto.SubjectExaminationDTO;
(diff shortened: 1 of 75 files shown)


youennmerel:
Some test are failing, checking it

pierrehenri-dauvergne (approved):

michaelkain (changes requested):
Dear Youenn, thank you for this PR and the solution proposed. I am not convinced, that the frontend should send a variable that influences the list of examinations, that comes back. The backend should always deliver the same list depending on the rights and studies. Furthermore the code, that searches study IDs from the related_datasets and groups it is complicated. I propose: for the moment, adding the user to the other study, to fix it, in the idea, that we will work on the feature Derived study, where we will have to clean this entire situation, see roadmap. Your fix should work, but I have fear we will get lost in the future, with the current implementation proposal, as this remains something we have to clean up globally. Thank you for your understanding and with kind regards, Michael

youennmerel:
Hi @michaelkain,

Thanks for reviewing this.

Just to clarify, I modified 2 things : 
1. added an API parameter to choose to include related datasets or no
2. in case related dataset are included, check user rights to exclude those for which the user does not have the rights on the study

Not implementing this 2 points means  :
1. examinations from related studies could be proposed when importing data in a study
2. study tree will fail to build if it contains related dataset to study on which the user have no rights (see https://github.com/fli-iam/shanoir-ng/pull/1696/files#diff-f2a5f4de3bd5ae1929701b5fe93b1adc2f8e39a717c351d0d9fa308ffa224800R206)

IMHO, I think we should at least implement the second point while waiting for related studies.

youennmerel:
Hi @michaelkain,

A priori, the solr view does not list the related datasets.

The findExaminationsBySubjectAndStudy API is used only for listing examination in import and in tree.

Currently, related examination / acquisition / datasets are displayed in tree, but without 'Details' links. 

image

youennmerel:
As discussed with Michael, we won't list the related examination for all users.

michaelkain:
Hi @youennmerel, I tested on qualif with Paolos account: when I click on the subject, I can see nothing today, except the name of the subject. So, when we remove it entirely, the user will not loose any data, but we will clear the implementation, before specifying and implementing a clean version of Derived study. With kind regards, Michael

michaelkain:
some tests remain in error Tests in error: 
  ExaminationApiControllerTest.deleteExaminationTest » IllegalState Failed

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
```diff
@@ -85,10 +76,13 @@ public class ExaminationApiController implements ExaminationApi {
...
 	@Autowired
-	ShanoirEventService eventService;
+	private ShanoirEventService eventService;
+
+	@Autowired
+	private StudyRepository studyRepository;
 
 	@Autowired
-	StudyRepository studyRepository;
+	private StudyRightsService rightsService;
```
julien-louis, line 85: It seems unused.

If not, you will have to mock it in the failing unit test (like the other @Autowired dependencies) : 

```
2023-05-04T15:36:21.3596446Z Tests in error: 
2023-05-04T15:36:21.3600774Z   ExaminationApiControllerTest.deleteExaminationTest » IllegalState Failed to lo...
2023-05-04T15:36:21.3601628Z   ExaminationApiControllerTest.findExaminationByIdTest » IllegalState Failed to ...
2023-05-04T15:36:21.3602328Z   ExaminationApiControllerTest.findExaminationsTest » IllegalState Failed to loa...
2023-05-04T15:36:21.3603700Z   ExaminationApiControllerTest.saveNewExaminationTest » IllegalState Failed to l...
2023-05-04T15:36:21.3604428Z   ExaminationApiControllerTest.testAddExtraData » IllegalState Failed to load Ap...
2023-05-04T15:36:21.3605533Z   ExaminationApiControllerTest.testDeleteExaminationWithExtraData » IllegalState
2023-05-04T15:36:21.3606563Z   ExaminationApiControllerTest.testDownloadExtraData » IllegalState Failed to lo...
2023-05-04T15:36:21.3607347Z   ExaminationApiControllerTest.testDownloadExtraDataNotExisting » IllegalState F...
2023-05-04T15:36:21.3608034Z   ExaminationApiControllerTest.updateExaminationTest » IllegalState Failed to lo...
```
```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'examinationApiController': Unsatisfied dependency expressed through field 'rightsService'; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.shanoir.ng.study.rights.StudyRightsService' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
```

julien-louis (changes requested):

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/controler/ExaminationApiController.java
youennmerel: @michaelkain @julien-louis tests are all right now, thx !

PR #1699, opened 2023-04-28
[fli-iam#1671] Optimize preclinical subjects loading

youennmerel:


shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/controler/SubjectApiController.java
```diff
@@ -92,11 +93,22 @@ public ResponseEntity<SubjectDTO> findSubjectById(
 	}
 
 	@Override
-	public ResponseEntity<List<SubjectDTO>> findSubjects() {
-		final List<Subject> subjects = subjectService.findAll();
+	public ResponseEntity<List<SubjectDTO>> findSubjects(boolean preclinical, boolean clinical) {
```
julien-louis, line 96: Is this necessary to have two booleans when a subject is either clinical or preclinical ?

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/controler/SubjectApi.java
```diff
@@ -68,7 +70,12 @@ ResponseEntity<Void> deleteSubject(
 	@GetMapping(value = "", produces = { "application/json" })
 	@PreAuthorize("hasAnyRole('ADMIN', 'EXPERT', 'USER')")
 	@PostAuthorize("hasRole('ADMIN') or @studySecurityService.filterSubjectDTOsHasRightInOneStudy(returnObject.getBody(), 'CAN_SEE_ALL')")
-	ResponseEntity<List<SubjectDTO>> findSubjects();
+	ResponseEntity<List<SubjectDTO>> findSubjects(
```
julien-louis, line 73: same : Is this necessary to have two booleans when a subject is either clinical or preclinical ?

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/controler/SubjectApiController.java
youennmerel: Hi @julien-louis,

I thought of 3 possible uses cases :
- we want all subject (default values : clinical = true & preclinical = true)
- we want only clinical (clinical = true & preclinical = false)
- we want only preclinical (clinical = false & preclinical = true)

However, for now, API is only used for getting all subjects, or just preclinical one.

Do you think of a more elegant solution ?

youennmerel:
Loading time is now ~ 1-2 seconds in qualif.

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/controler/SubjectApiController.java
julien-louis: Ok I see now, seems fine, thank you

julien-louis (approved):

Issue #1619, opened 2023-03-31
New GUI: access request, study list


michaelkain:
- column "public" is empty for studies, for which we have access
- maybe it would be better to display an open lock
- and a closed lock for public studies, that we have not yet access too?

PR #1714, opened 2023-05-12
Locks in studies

julien-louis:
https://github.com/fli-iam/shanoir-ng/issues/1619


youennmerel (approved):
Hi @julien-louis,

Tested it in qualif, everything seems all right !

PR #1737, opened 2023-06-06
shanoir-issue#1728-2: when building studycenter object from studycenterDTO, also keep prefixes

### LINKED ISSUE #1728: subject-prefix not working
Dear @pierrehenri-dauvergne, I hope you are doing fine.
Solène was asking me for an interesting feature for her (as member of the consortium shanoir), that we had already implemented.
Normally, when I edit a study - tab centers, I can enter a subject-prefix par center, to be used for the same study as prefix during the import, that all subjects start with the same prefix per center (can not be changed). When I try to configure such a prefix with edit study (I used the account of Solène, who is study admin), I can not configure the prefix, BUt it tells me, that the update was successful. It would be great to bring back that feature, as we especially developed it. Thank you and with kind regards, Michael

pierrehenri-dauvergne:
Closes #1728

### DIFF
--- a/shanoir-ng-front/src/app/centers/shared/center.dto.ts
+++ b/shanoir-ng-front/src/app/centers/shared/center.dto.ts
@@ -71,6 +71,7 @@ export class CenterDTOService {
             //studyCenter.center.id = dto.id;
             studyCenter.study = new Study();
             studyCenter.study.id = scDto.study.id;
+            studyCenter.subjectNamePrefix = scDto.subjectNamePrefix;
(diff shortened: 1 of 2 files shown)


jcomedouteau (commented):
@pierrehenri-dauvergne 
To keep it consistent, could you add it to center.dto.ts file too ?

Everything works fine otherwise !

jcomedouteau (approved):
Working file:
- Prefix is saved and well displayed
- Prefix appears when creating a new subject in the study

PR #1739, opened 2023-06-07
#1738 - multi-exam-import bug

### LINKED ISSUE #1738: Multi-exam import fails
Currently, multi examination does not work.
This is due a bug on our side when creating the new subject

jcomedouteau:
- Avoid unavoidable NPE

How to test:
- Ask me for some multi-examination datasets
- Realize an import with such data (Import -> Dicom -> check multi-examination)

-> The file loads correctly and then import goes on and does not fail.

Closes #1738


michaelkain (approved):
Hi Jean-Come, I approve this by code review only. It is only in the method for multi-exam import (at the moment probably only used by your colleague). With kind regards, Michael

PR #1753, opened 2023-06-19
shanoir-issue#1750: bugfix, access-request popup is now shown when us…

pierrehenri-dauvergne:
…er opens study details (if no access)


shanoir-ng-front/src/app/shared/components/entity/entity-list.component.abstract.ts
```diff
@@ -194,7 +194,14 @@ export abstract class EntityListComponent<T extends Entity> implements OnDestroy
         this.router.navigate([this.entityRoutes.getRouteToList()]);
     }
 
-    rowClick(item): string {
+    getRowRoute(item): string {
+        if (item.visibleByDefault && item.locked && !this.keycloakService.isUserAdmin()) {
+            if (item.accessRequestedByCurrentUser) {
```
jcomedouteau, line 199: Just return null ?

jcomedouteau:
For the rest, it works well :) !

jcomedouteau (approved):

PR #1765, opened 2023-06-30
Fix overflow

julien-louis:
* put back overflow hidden in tables 
* deal with the hidden select boxes menus in tables


pierrehenri-dauvergne:
Works well until I add a new subject in the list

julien-louis:
@pierrehenri-dauvergne 

This bug already exists in develop. (adding a subject to a study and giving him a tag at the same time)

I propose to merge this pr and fix the bug later

pierrehenri-dauvergne (approved):

Issue #1774, opened 2023-07-06
ZIP Import: issue with multi-frame DICOM


michaelkain:
- see mails exchanged
- many examinations created

michaelkain:
first analysis: probably: multi-exam import checkbox has been used

michaelkain:
My further findings:
The MultiframeExtractor throws a NullPointerException on the last line of the following lines of code:
String frameType3 = sf.getString(Tag.ImageType, 2);
if ("ANGIO".equals(frameType3))
    sf.setString(Tag.AngioFlag, VR.CS, "Y");
if (frameType3.startsWith("CARD")) // NPE thrown here, because frameType3 is empty

The reason is the below attribute, has the below content for 2 series out of all series:
(0008,0008) CS [ORIGINAL\PRIMARY] ImageType
frameType3 gets the third part of the string separated by backslash from ImageType, what is null in the two error series

For all other series, they have the following values:
(0008,0008) CS [ORIGINAL\PRIMARY\**T1**\NONE] ImageType
frameType3 becomes T1 here, and no error occurs or
(0008,0008) CS [ORIGINAL\PRIMARY\DIFFUSION\NONE] ImageType
frameType3 becomes DIFFUSION here, and no error occurs

michaelkain:
from the DICOM Enhance MR Image specification:

C.8.16.1 Image Type and Frame Type

**Image Type (0008,0008)** and associated Image Type related Attributes provide a high level description of a multi-frame SOP Instance. These Attributes describe properties that provide key summary information to users of the SOP Instance. Image Type (0008,0008) contains the highest level summary of what is in the SOP Instance.

 Frame Type (0008,9007) mirrors the corresponding Image Type Attribute and applies to the frame level rather than to the image level.

 **Image Type (0008,0008) and Frame Type (0008,9007) shall consist of four values.**

2 series of the above error provide one two values instead of four values!!!


**C.8.16.1.3 Image Flavor**

**Value 3** is an overall representation of the image type. This value may be a summary of several other Attributes or a duplication of one of the other Attributes to indicate the most important aspect of this image. Value 3 Image Flavor is to be used with Value 4 Derived Pixel Contrast to indicate the nature of the image set.
Note

For example Value 3 = DIFFUSION together with Value 4 = NONE indicates that the image set was originally collected for DIFFUSION.

[...]

PR #1788, opened 2023-07-11
MS Import: better error/ignore management + update to dcm4che3 v5.30.0

michaelkain:
This PR is the answer to the GitHub issue https://github.com/fli-iam/shanoir-ng/issues/1774

It introduces new attributes on the DICOM serie (used in select series): erroneous + ignored, that we bring back the information to the user, if the exam/DICOM study contains a serie, that causes problems, that the user directly sees, that there is a problem and on which serie, what is hidden today.

Furthermore this PR updates dcm4che3 to version 5.30.0 on all shanoir-ng.

PR contains fix for OFSEP import issue with one center, that had NPE during isSpectro check, now fixed.


michaelkain:
Tested successfully locally with
- multi-frame issue data (NoDicomDir), MR Enhanced
- sample.zip, MR standard
- CT.zip, CT standard
- ShanoirUploader, MR sample1.zip
All imports worked.

michaelkain:
Hi @julien-louis, it would be great, if the web import GUI of Shanoir in select-series would express the new attributes on serie:
- erroneous (maybe in "red" and serie disabled, to avoid selection; and maybe display the errorMessage in some nice way)
- ignored (maybe in "grey" and serie disabled, to avoid selection)

michaelkain (commented):
All reviewed again.

michaelkain:
Sorry, I require this PR for further evolution on the import, and in ShanoirUploader. So I merge it. I tested it successfully locally and it will be part of our big testing on Spring Boot 3.1.2.

PR #1799, opened 2023-07-19
hide tags in clinical context if none + fix select box bug

julien-louis:
* hide tags in clinical context if none
* fix select box bugs with the compact mode (typing or reseting search)


pierrehenri-dauvergne:
When opening the available tag list, it is empty but a "reset search" link appears. When clicked, an error occurs in console.

image
image

pierrehenri-dauvergne (approved):

PR #1810, opened 2023-08-03
RenameSuggestion

AnonymousAccount4SE:
### Renaming Suggestion of Method Names to Make Them More Descriptive
---
**Description**
---

We have developed a tool (**NameSpotter**) for identifying non-descriptive method names, and using this tool we find several non-descriptive method names in your repository:
```java
/* Non-descriptive Method Name in shanoir-ng-studies/src/main/java/org/shanoir/ng/shared/common/CommonIdsDTO.java */
	public Long getEquipementId() {
		return equipementId;
	}

	public void setEquipementId(Long equipementId) {
		this.equipementId = equipementId;
	}

```
We considered "get(set)EquipementId" as non-descriptive because it contains a typo, i.e., "Equipement" -> "Equipment", which could be misleading (there are also "Equipment" used in the project). We help to correct all the "Equipement" to "Equipment".

```java
/* Non-descriptive Method Name in shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/service/DICOMWebService.java */
	private String extractSeriesUIDUID(String url, String seriesUID) {
		return extractUidPattern(url, "seriesUID", "&objectUID", seriesUID);
	}
```
"extractSeriesUIDUID" contains two "UID", which is redundant.  We have corrected them (including the comments and the same name occurring in other file) and submitted a pull request.

Do you agree with the judgment? 

- If not, could you please leave your valuable opinion?

- If you do agree, the relevant modification has been submitted as a PR, and thanks for your precious time to consider it.

### DIFF
--- a/docs/Shanoir-NG_Import/Testing_Import.txt
+++ b/docs/Shanoir-NG_Import/Testing_Import.txt
@@ -4,7 +4,7 @@ Testing of new import
 - dicom zip upload and import testing
 ----  sampleArchiveHealthy1.zip
 -------- ExamCard appear empty in the tree, remove their series (fixed)
--------- DicomEquipement is empty (fixed, now DicomEquipment is displayed)
+-------- DicomEquipment is empty (fixed, now DicomEquipment is displayed)
 -------- But with only one serie: bug?, maybe selected == null? (fixed, yes, related, check added), still bug in ms datasets now, to test tomorrow (fixed in ms datasets)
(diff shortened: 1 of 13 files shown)


AnonymousAccount4SE:
Hi, do you agree with our proposals? Even if you can not merge it for various reasons, please tell us if this is fine, which means a lot to me. Thanks!

PR #1824, opened 2023-08-22
shanoir-issue#1780-2: add links to preclinical examination table

pierrehenri-dauvergne:


youennmerel:
Tested on qualif, links in preclinical examination table are okay for Subject, Study, Center

Issue #1829, opened 2023-08-22
[Import] EEG import fails


jcomedouteau:
EEG import fails when starting import job -> 401 [no body]

Linked to the fact that we expect a ResponseEntity<Void> and actually have a ResponseEntity<String>

PR #1833, opened 2023-08-23
#1829-EEG-import-bug - Use rabbitMQ

jcomedouteau:
How to test:
- Get some EEG data (ask me if necessary)
- Import EEG data, the "Import now" step should now be working well.


michaelkain (approved):
Hi Jean-Come, thank you! I approve your PR and will merge it. It is very straight forward. Great, that you replaced the former REST-call with a rabbit-mq message, what is even better and fits better to the design of the other imports. My local tests worked and I could import the EEG dataset. It was not indexed into Solr, so I shall probably create an issue for that. With kind regards, Michael

PR #1848, opened 2023-09-04
[fli-iam#1846] Correct order by numeric value in lists

youennmerel:
In study list :
- Sorting by "storage volume" is now correct
- Sorting on "storage volume" is disabled while fetching
- ~Storage volumes are fetched only for studies loaded in the current page~ (not possible due to sorting)

Correct https://github.com/fli-iam/shanoir-ng/issues/1868

How to test :
- Display study list as admin (so all the studies will be loaded, it will take some time)
- Order by "storage volume", asc & desc (must be disabled while fetching)
- Display study details

### DIFF
--- a/shanoir-ng-front/src/app/studies/study-list/study-list.component.ts
+++ b/shanoir-ng-front/src/app/studies/study-list/study-list.component.ts
@@ -171,7 +171,7 @@ export class StudyListComponent extends BrowserPaginEntityListComponent<Study> {
             {
                 headerName: "Storage volume", field: "totalSize", disableSearch: true, disableSorting: this.isStudyVolumesFetching, type: "number", orderBy: ["totalSize"],
                 cellRenderer: (params: any) => {
-                    if (params.data.totalSize != null && this.isStudyVolumesFetching) {
+                    if (params.data.totalSize == null && this.isStudyVolumesFetching) {
                         return "Fetching..."
                     }
                     return this.studyService.storageVolumePrettyPrint(params.data.totalSize);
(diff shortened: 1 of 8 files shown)


shanoir-ng-front/src/app/studies/study/study.component.ts
```diff
@@ -118,6 +118,9 @@ export class StudyComponent extends EntityComponent<Study> {
     }
 
     initView(): Promise<void> {
+
+        console.log(this.breadcrumbsService.currentStep.entity);
```
jcomedouteau, line 122: Remove debug logs ?

shanoir-ng-front/src/app/studies/study-list/study-list.component.ts
```diff
@@ -96,33 +99,41 @@ export class StudyListComponent extends BrowserPaginEntityListComponent<Study> {
...
-                                sizesByLabel.set(DatasetExpressionFormat.getLabel(sizeByFormat.format), sizeByFormat.size);
+            this.fetchStorageVolumes(studies);
+        });
+        return earlyResult;
+    }
+
+    private fetchStorageVolumes(studies: Study[] | AccessRequest[]) {
+        let pageSize = Number(this.table.maxResults);
+        let promises = [];
+        for (let i = 0; i < studies.length; i += pageSize) {
```
jcomedouteau, line 110: Why do we do it page by page ? Is it too long otherwise ?

jcomedouteau: It is quite weird that we can start to sort by size wereas all study size are not loaded yet -> The display can be then misleading

michaelkain: Hi! I agree with Jean-Come. Users will find our tool strange, when we propose a filter on size, but only consider the current page.

jcomedouteau (approved):

PR #1851, opened 2023-09-07
Quality control fixes and enhancements

julien-louis:


ShanoirChallengeMigrator/.project
```diff
@@ -20,4 +20,15 @@
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
```
michaelkain, line 23: Hi @julien-louis, same here: I am not sure, if this file should be part of the PR? It might be good to remove it?

shanoir-ng-users/.project
```diff
@@ -45,12 +45,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1600787284252</id>
```
michaelkain, line 48: same here, not sure, if necessary

shanoir-ng-study-rights/.project
```diff
@@ -22,12 +22,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1600787284249</id>
+			<id>1697013852272</id>
```
michaelkain, line 25: same here, please remove

shanoir-ng-parent/.project
```diff
@@ -20,4 +20,15 @@
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
```
michaelkain, line 23: same here, please remove

shanoir-ng-front/src/app/studies/study-list/study-list.component.ts
```diff
@@ -261,7 +259,7 @@ export class StudyListComponent extends BrowserPaginEntityListComponent<Study> {
                     const text: string = 'You are a member of at least one study that needs you to accept its data user agreement. '
                         + 'Until you have agreed those terms you cannot access to any data from these studies. '
                         + 'Would you like to review those terms now?';
-                    const buttons = {ok: 'Yes, proceed to the signing page', cancel: 'Later'};
+                    const buttons = {yes: 'Yes, proceed to the signing page', cancel: 'Later'};
```
michaelkain, line 262: Hi @julien-louis, I think this belongs to DUA. Is this something missing in today's develop? and you use this PR to bring it into prod? In case yes, it would be good to mention it in the PR "notes".

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
```diff
@@ -75,7 +75,7 @@ public abstract class Dataset extends AbstractEntity {
 
 	/** Dataset expression list. */
 	@JsonIgnore
-	@OneToMany(fetch = FetchType.LAZY, mappedBy = "dataset", cascade = CascadeType.ALL)
+	@OneToMany(fetch = FetchType.EAGER, mappedBy = "dataset", cascade = CascadeType.ALL)
```
michaelkain, line 78: Dear @julien-louis, I am really wondering, why this change here is to "eager" and same below for the dataset_files. The performance impact is really, really bad here, when each time, we get a dataset, we grab expression + files with eager, what results in huge query results.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/DatasetExpression.java
```diff
@@ -58,7 +58,7 @@ public class DatasetExpression extends AbstractEntity {
 	private Long size;
 
 	/** Set of files. */
-	@OneToMany(fetch = FetchType.LAZY, mappedBy = "datasetExpression", cascade = { CascadeType.MERGE,
+	@OneToMany(fetch = FetchType.EAGER, mappedBy = "datasetExpression", cascade = { CascadeType.MERGE,
```
michaelkain, line 61: same here, not sure, if this is really necessary

shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/model/DatasetAcquisition.java
```diff
@@ -67,7 +67,7 @@ public abstract class DatasetAcquisition extends AbstractEntity {
 	
 	/** Datasets. */
 	@JsonIgnore
-	@OneToMany(fetch = FetchType.LAZY, mappedBy = "datasetAcquisition", cascade = CascadeType.ALL)
+	@OneToMany(fetch = FetchType.EAGER, mappedBy = "datasetAcquisition", cascade = CascadeType.ALL)
```
michaelkain, line 70: Hi @julien-louis, thank you for our exchange! Here is another place, where it might be interesting to keep lazy please. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/dto/QualityCardResultEntry.java
```diff
@@ -35,6 +35,8 @@ public class QualityCardResultEntry {
 	
 	private QualityTag tagSet;
 
+	private boolean faildValid = false; // if tag VALID was to put but conditions failed
```
michaelkain, line 38: Hi @julien-louis, very minor typo: "failed"Valid would be great please. Thank you!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/DatasetAttributes.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package org.shanoir.ng.download;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.dcm4che3.data.Attributes;
+
+public class DatasetAttributes {
```
michaelkain, line 22: Hi, as discussed, please verify if it is still in use this class.

michaelkain:
Hi @julien-louis, this one is somehow minor as well, but might be confusing for the users. Please see my screenshot attached. I think it is "legacy" from the study cards, but for the quality cards, we should display "**Name of dataset**" instead of "New name for the dataset" please. Thank you and with kind regards, Michael
Screenshot from 2023-11-20 11-08-40

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/repository/ExaminationRepository.java
```diff
@@ -69,13 +70,19 @@ public interface ExaminationRepository extends PagingAndSortingRepository<Examin
...
 	 */
 	List<Examination> findByStudy_Id(Long studyId);
+
+	/**
+	 * Get a list of examinations for a study.
+	 * 
+	 * @param studyId
+	 * @return list of examinations.
+	 */
+	@Query("select e.id from Examination e where e.study.id = :studyId")
```
michaelkain, line 84: very good!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
```diff
@@ -41,4 +47,40 @@ public Attributes getDicomObjectAttributes(DatasetFile image, Boolean isEnhanced
 		}
 	}
 
+    public ExaminationAttributes getDicomExaminationAttributes(Study study, Boolean isEnhanced) throws PacsException {
+		ExaminationAttributes attributes = new ExaminationAttributes();
+		if (study != null) {
+			for (Serie serie : study.getSeries()) {
+				attributes.addAcquisitionAttributes(serie.hashCode(), getDicomAcquisitionAttributes(serie, isEnhanced));
```
michaelkain, line 54: Hi @julien-louis, I am wondering, you are using hashCode here, instead of seriesInstanceUIDs, here but later you search with long ids for datasets, what is quite confusing. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
```diff
@@ -41,4 +47,40 @@ public Attributes getDicomObjectAttributes(DatasetFile image, Boolean isEnhanced
...
+			}
+		}
+		return attributes;
+    }
+
+	public AcquisitionAttributes getDicomAcquisitionAttributes(Serie serie, Boolean isEnhanced) throws PacsException {
+		AcquisitionAttributes attributes = new AcquisitionAttributes();
+		for (Dataset dataset : serie.getDatasets()) {
+			try {
+				attributes.addDatasetAttributes(dataset.hashCode(), getDicomObjectAttributes(serie.getFirstDatasetFileForCurrentSerie(), isEnhanced));
```
michaelkain, line 74: same here, you add it by hashCode, and later I think you search it by long id, maybe I am wrong

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
```diff
@@ -41,4 +47,40 @@ public Attributes getDicomObjectAttributes(DatasetFile image, Boolean isEnhanced
 		}
 	}
 
+    public ExaminationAttributes getDicomExaminationAttributes(Study study, Boolean isEnhanced) throws PacsException {
```
michaelkain, line 50: Hi @julien-louis, the PacsException in this class are confusing, as all calls get into DicomObjectAttributes, that is read from the disk, if I understood it well. With kind regards, Michael

michaelkain:
Hi @julien-louis, there is a bug introduced now, when I do an import. Too many datasets are created for my test example with only 1 t1 serie. As I showed you, on qualif, my example creates only one dataset. And now, on the PR it creates 27 datasets, see my screenshot. Maybe related to the "migration" of SerieToDatasetsSeparator? With kind regards, Michael

michaelkain:
Very good point: I tested successfully an imported refused, because of missing serie content. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/controler/QualityCardApiController.java
```diff
@@ -169,4 +176,18 @@ public ResponseEntity<QualityCardResult> testQualityCardOnStudy(
         return new ResponseEntity<>(results, HttpStatus.OK);
     }
 
+	@Override
+    public ResponseEntity<QualityCardResult> testQualityCardOnStudy(
+        @Parameter(name = "id of the quality card", required = true) @PathVariable("qualityCardId") Long qualityCardId,
+		@Parameter(name = "dataset number start ", required = true) @PathVariable("start") int start,
```
michaelkain, line 182: Hi, please change swagger here from "dataset number start" to "examination number start". Thank you!

michaelkain:
Hi @julien-louis, I hope you are doing good. There might be a big optimization potential here, please see below. For the ExaminationAttributes, all is in the memory and only read once from the disk in DicomProcessing, very good! But for the DICOM information from the PACS you iterate over all conditions and maybe call the pacs for all acquisitions, for each condition. Maybe better get all before and re-use, when you check all conditions? With kind regards, Michael
       for (StudyCardCondition condition : getConditions()) {
            StringBuffer msg = new StringBuffer();
            boolean fulfilled = true;
            if (condition instanceof StudyCardDICOMConditionOnDatasets) {
                if (dicomAttributes != null) {
                    fulfilled = ((StudyCardDICOMConditionOnDatasets) condition).fulfilled(dicomAttributes, msg);
                } else {
                    fulfilled = ((StudyCardDICOMConditionOnDatasets) condition).fulfilled(**examination.getDatasetAcquisitions()**, msg);
                }

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/StudyCardDICOMConditionOnDatasets.java
```diff
@@ -0,0 +1,300 @@
...
+@Component
+@DiscriminatorValue("StudyCardDICOMConditionOnDatasets")
+@JsonTypeName("StudyCardDICOMConditionOnDatasets")
+public class StudyCardDICOMConditionOnDatasets extends StudyCardCondition implements ApplicationContextAware {
+	
+    private static final Logger LOG = LoggerFactory.getLogger(StudyCardDICOMConditionOnDatasets.class);
+    
+	private int dicomTag;
+
+    private static ApplicationContext context;
```
michaelkain, line 57: Hi @julien-louis, I do not think the AppContext is necessary here, to discuss, what the best solution might be, but this might be to change: WADODownloaderService downloader = (WADODownloaderService) context.getBean("WADODownloaderService");. The reference to the static context is complicated and the bean look up, takes time as well (even if not much probably). We should find a way please to use @Autowired

michaelkain: and move some logic to a separate service, that deals with the entity

michaelkain:
Hi @julien-louis, it might be interesting to analyze the method downloadMetadataFromPacs in WADODownloaderService, that uses today RestTemplate.exchange, if the http-connection to the pacs remains open during each call and if the http-requests could be sent in parallel. Keeping the connection open, might create huge performance gains, sending in parallel with a thread-pool, see apache-http-client, the request, can bring nice perf gains as well. With kind regards, Michael

MS_ChallengeScores/.project
```diff
@@ -33,12 +33,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1600787284235</id>
+			<id>1697013852269</id>
```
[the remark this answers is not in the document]
julien-louis: I don't understand why this is here, I'll remove it

shanoir-ng-front/src/app/studies/study-list/study-list.component.ts
julien-louis: I just changed the variable name from 'ok' to 'yes' because I added the option 'no' to use the confirmation component for the "do you want to test qc on a sample ? yes / no,all exams / cancel" choice.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
julien-louis: done, seems to work with lazy

MS_ChallengeScores/.project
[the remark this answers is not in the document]
julien-louis: I will put emojis to know what comment I've assert yet

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/DatasetAttributes.java
julien-louis: removed

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
julien-louis: I think that in the case of an import I don't have the dataset ids yet, so I went with this. Do you think it's too unreadable ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
julien-louis: I've changed it to ShanoirException

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
julien-louis: Actually there may be a big problem here. I'm addressing it.

michaelkain:
Hi @julien-louis, I hope you are doing fine. It would be great to solve the merge conflict please. Thank you, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DicomProcessing.java
julien-louis: Ok I used seriesInstanceUIDs and SOPInstanceUID. Those are strings so I had to parametrize the two classes ExaminationAttributes and AcquisitionAttributes which is better because it reduces the risk of mixing up the two cases (db ids or dicom uids).

julien-louis:
@michaelkain I've fixed all the issues, except for the http request optimization. I will look at it now.

julien-louis:
@michaelkain I'm currently testing on ICAN, 5% in 2minutes, so the estimation is about 40min. Way better.

michaelkain:
@julien-louis Perfect! Happy to read about the 40min, what looks far better. I am reviewing and testing. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -352,45 +362,48 @@ private String getExaminationFirstDatasetUrl(Examination examination) {
...
-		long ts = new Date().getTime();
-		Dataset ds = getFirstIfExist(acquisition.getDatasets());
-		if (ds == null) return null;
-		Attributes result = getDicomAttributesForDataset(ds);
-		LOG.debug("get DICOM attributes for dataset " + ds.getId() + " : " + (new Date().getTime() - ts) + " ms");
-		return result;
+		return dAcquisitionAttributes;
 	}
 
 	public String getDicomJson(Examination examination) throws IOException {
```
michaelkain, line 409: Hi @julien-louis, could you please replace the stackTrace with a logger here in this method? Thank you!

michaelkain: please remove the unused method getDicomJson, not referenced anywhere, if my search is not wrong

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -315,9 +317,17 @@ public List<File> downloadDicomFilesForURLs(final List<URL> urls, final File wor
 
 	public String downloadDicomMetadataForURL(final URL url) throws IOException, MessagingException, RestClientException {
```
michaelkain, line 318: Hi @julien-louis, I think both methods downloadDicomMetadataForURL should be fusioned please. With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/StudyCardDICOMConditionOnDatasets.java
```diff
@@ -0,0 +1,317 @@
...
+@Configurable
+@DiscriminatorValue("StudyCardDICOMConditionOnDatasets")
+@JsonTypeName("StudyCardDICOMConditionOnDatasets")
+public class StudyCardDICOMConditionOnDatasets extends StudyCardCondition {
+	
+    private static final Logger LOG = LoggerFactory.getLogger(StudyCardDICOMConditionOnDatasets.class);
+    
+	private int dicomTag;
+
+    @Autowired
```
michaelkain, line 55: please clean up here

julien-louis:
@michaelkain it seems a much more slower this morning

julien-louis:
anyway I've made the corrections

julien-louis:
@michaelkain actually the testing seems fine but the "apply" is slow, I should investilgate

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/service/CardsProcessingService.java
```diff
@@ -77,84 +80,132 @@ public void applyStudyCard(StudyCard studyCard, List<DatasetAcquisition> acquisi
...
+		eventService.publishEvent(event);
+        Study study = studyService.findById(qualityCard.getStudyId());
+		if (study == null ) throw new IllegalArgumentException("study can't be null");
+		if (qualityCard.getStudyId() != study.getId()) throw new IllegalStateException("study and studycard ids don't match");
+		if (CollectionUtils.isNotEmpty(qualityCard.getRules())) {	    
+		    QualityCardResult result = new QualityCardResult();
+            Float i = 0f;
+            List<Examination> examinations;
+            if (start != null && stop != null) {
+                examinations = study.getExaminations().subList(start, stop < study.getExaminations().size() ? stop : study.getExaminations().size());
```
michaelkain, line 156: Hi @julien-louis, when you select a range of 10 exams, you will always get all from the database. A dedicated call to the database might be much more efficient? With kind regards, Michael

michaelkain: so, I mean you call all, and the you select a subList from all, why not only getting the range from the db

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/StudyCardDICOMConditionOnDatasets.java
```diff
@@ -0,0 +1,317 @@
...
+     * @param examinationAttributesCache to be used as a cache
+     * @param errorMsg
+     * @return
+     */
+    public boolean fulfilled(List<DatasetAcquisition> acquisitions, ExaminationAttributes<Long> examinationAttributesCache, WADODownloaderService downloader, StringBuffer errorMsg) {
+        if (acquisitions == null) throw new IllegalArgumentException("acquisitions can not be null");
+        int nbOk = 0; int total = 0; int nbUnknown = 0;
+        for (DatasetAcquisition acquisition : acquisitions) {
+            if (!examinationAttributesCache.has(acquisition.getId())) {
+                examinationAttributesCache.addAcquisitionAttributes(acquisition.getId(), new AcquisitionAttributes<Long>());
```
michaelkain, line 113: Hi @julien-louis, I hope you are doing fine. I am not sure, your cache will work here, as you always only add a new and empty object to it?

michaelkain: for the DatasetAttributes, it is different, it looks logic

julien-louis:
I have a bug. With the large study I***, the quality check stops silently at examination "CRANE", no logs

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/ExaminationAttributes.java
```diff
@@ -0,0 +1,129 @@
...
+        return sb.toString();
+    }
+
+    public Set<T> getAcquisitionIds() {
+        return acquisitionMap.keySet();
+    }
+
+    public void addAcquisitionAttributes(T acquisitionId, AcquisitionAttributes<T> dicomAcquisitionAttributes) {
+        if (acquisitionMap.containsKey(acquisitionId) && acquisitionMap.get(acquisitionId).isPresent()) {
+            acquisitionMap.get(acquisitionId).get().merge(dicomAcquisitionAttributes);
```
michaelkain, line 119: Hi @julien-louis, this code looks somehow strange, why do you merge acquisitionAttributes?

julien-louis: Here I want to add acquisition dicom attributes to the index acquisitionId

But if that index already contains data (dicom Attributes indexed on dataset ids) I don't want to overwrite it, I want to merge it, adding more (datasetId => Attributes) to the given acquisitionId.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/StudyCardDICOMConditionOnDatasets.java
michaelkain: you are right, but as you have multiple acquisitions, you can not only add once:addAcquisitionAttributes(acquisition.getId(), new AcquisitionAttributes<Long>()); so you always add a new acq attributes to the cache?

michaelkain (commented):
Hi Julien, my code review is finished now :-), so I should not find something new. Thank you already for your corrections! I will have to test now locally.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
michaelkain: THX

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/service/CardsProcessingService.java
michaelkain: hm, please check with the debugger and the sql command trace, what is really send to the database, as subList is a collection command, that has nothing to do with hibernate, I would be really surprised, if it ends up in a ranged sql query... but, it would be great if :-)

julien-louis:
Applying qc on large study (**AN), on qualif : 45 min

michaelkain (commented):
THX, fine, code reviewed again

michaelkain:
Hi Julien, please see my screenshot attached.
Screenshot from 2023-12-08 17-10-15
Do you agree, that my rule means all datasets of the exam must be t1, and if not, throw an error?
It does not work. I could import the sample.zip, that contains only not-t1 series and I had no error.

julien-louis:
I couldn't import sample.zip in my new empty study test julien 2 : 
image

qc : 
image

michaelkain:
I tested with an existing exam, that had already only t1 inside

julien-louis:
@michaelkain actually the empty study card message was about empty quality cards, I changed the message. I don't know what to do with an empty quality card. Return an empty report ?

julien-louis:
@michaelkain Ok I checked and you are right, I think at import the quality control is on the whole exam, not just the imported data. I will assert that immediately.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/ImporterService.java
```diff
@@ -249,26 +298,37 @@ private Set<DatasetAcquisition> generateAcquisitions(Examination examination, Im
...
-                qualityResult.merge(qualityCard.apply(examination, dicomAttributes));                       
+                qualityResult.merge(qualityCard.apply(examination, dicomAttributes, downloader));                       
             }
         }
         return qualityResult;
     }
 
+    private QualityCardResult checkQuality(Examination examination, Set<DatasetAcquisition> limitToTheseAcquisitions, ImportJob importJob) throws ShanoirException {
+        // save the exam acquisitions
+        List<DatasetAcquisition> saveList = new ArrayList<>();
```
michaelkain, line 319: Hi Julien, thank you for your corrections. I am wondering about the below code. The ImporterService is only responsible for new-imported-data (either on a new and empty exam or an existing exam). The below code is really quite a work-around (sorry for the word), but putting existing acquisitions into memory, simulate a new exam and then putting back the old acquisitions, is really strange. I have fear in two months we will not understand the code ourselves. Please let's discuss this and how we continue from here. Thank you and with kind regards, Michael

michaelkain:
Hi, I have a new issue during my tests. The dates are corrupted somehow in the report table in the frontend and in the csv, please see my screenshot. With kind regards, Michael
Screenshot from 2023-12-12 11-38-57

michaelkain:
Thank you for the time added to the tasks table! Very useful!

michaelkain (approved):
Hi Julien,
my tests on qualif are looking fine now. We will plan the refactor for 2024, see my last comment. I approve. See below my test result, with kind regards, Michael

QC rule created, on study demonstration:
Quality rule, that requires at least one T1,t1 serie in the metadata of shanoir

TEST: test button used: correct results, apply button used: correct results: OK

--- IMPORT-TESTS:

TEST: import quality conform serie, into new subject, new exam: OK
t1 phantom serie used, new subject + exam created during import

TEST: use quality rule conform serie, into existing exam, that is not quality-conform: OK
1 serie selected, that is t1_fl2d_tra_3mm mtc 10/08/2009 MR
Use subject TestMK20210107-01, that has existing exam CT chest plan

TEST: use serie, not quality-conform, into existing exam, that is quality-conform: OK
1 serie selected tse_vfl_WIP607 04/03/2010 MR, not qc conform
existing exam selected with one serie, qc conform

Issue #1853, opened 2023-09-08
[Multi exam import] BUG - When importing lots of data, temporary folder is deleted too soon


jcomedouteau:
When importing multi exam data, we call data import separately -> temporary folder deletion is triggered too soon by the first examination imported -> The following are failing

PR #1855, opened 2023-09-12
#1853 multi-exam-import bug

jcomedouteau:
- Correct logging in case of error with specific pacs error
- Do not delete directly the temp folder when importing

How to test:
- Ask me for data
- Test multi exam import with the data
- -> If error, the import does not stop.
- -> If no error, no error.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DIMSEService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DIMSEService.java
@@ -97,7 +97,9 @@ public class DIMSEService {
         BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
         String line;
         while ((line = reader.readLine()) != null) {
-            LOG.info(line);
+        	if (line.contains(" ERROR ")) {
+        		LOG.error(line);
+        	}
         }
         int exitCode = process.waitFor();
 		if (exitCode != 0) {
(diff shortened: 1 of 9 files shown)


michaelkain:
Hi @jcomedouteau, thank you for your PR. I hope you are doing fine. I am wondering about something and maybe I am wrong. We create an ImportJob par exam, that has a specific folder. Why, when using that folder, each importJob can not clean his workFolder at the end of each import per exam? Why do we require a flag is multipleExam in an ImportJob, that is by default made for one exam only? Maybe I am wrong and with kind regards, Michael

michaelkain:
To be more precise: I mean setting importJob.WorkFolder to the current folder of the exam.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/DIMSEService.java
```diff
@@ -91,13 +91,14 @@ public void sendDicomFilesToPacs(File directoryWithDicomFiles) throws Exception
 	private void execute(List<String> args) throws Exception {
 		LOG.debug("Calling command: " + args.toString());
 		ProcessBuilder processBuilder = new ProcessBuilder();
+		processBuilder.redirectErrorStream(true);
```
michaelkain, line 94: Great!

jcomedouteau:
Hi @michaelkain, welcome back !
You're right, my boolean was a false lead as I thought we had a problem when deleting import folder too soon, but this is not the case. The problem was directly linked to the send to the pacs method execution that was stopping abruptly.
I'm removing this !

jcomedouteau:
@michaelkain as discussed, I added an new log file for DIMSE
image

michaelkain:
Hi, I tested and it works. We need to discuss with Anthony, if 500KB per import are an issue. Means 50MB logs per day for OFSEP, what could be a lot. And please fix the pseudo issue on this PR, before merge. With kind regards, Michael

michaelkain (approved):
Thank you very much! It works on my machine.

PR #1894, opened 2023-10-05
Shanoir uploader fixes trailing slash

michaelkain:
Our production ShUp did not work anymore, because of the below change in SB3.
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#spring-mvc-and-webflux-url-matching-changes


michaelkain:
Hi, I tested on my machine with trailing slahes on the URLs in ShUp, as before, and it worked.

jcomedouteau (approved):
worked well

Issue #1902, opened 2023-10-11
[Subject Tags] Can't set first subjectTag


pierrehenri-dauvergne:
Open a study
Go to 'subject' tab
Update
Create 3 subject tags
Try to set the first tag to a subject, it's disabled.

![image](https://github.com/fli-iam/shanoir-ng/assets/117283961/4226398e-09b5-48a5-b71f-eff98858a518)

You can add the 2nd tag, then add the first one, but a small issue remains here.

PR #1904, opened 2023-10-11
#1902-openjdk-removal

### LINKED ISSUE #1903: remove openjdk 11 from import MS

jcomedouteau:
How to test:
- Import data use mriconverter converter (CT is better)

closes #1903


michaelkain (approved):
Hi Jean-Come, thank you! I approve, as the removal of JDK-11 is a good thing in any case. I will test mri converter with CT during our next test session. With kind regards, Michael

PR #1908, opened 2023-10-12
shanoir-issue#1905-2: add condition to check if arrays are empty

pierrehenri-dauvergne:


pierrehenri-dauvergne:
How to test: ask me

julien-louis (approved):

PR #1913, opened 2023-10-20
add dicom_size_mo column in export statistics procedure

arthursw:
The DICOM size is the sum of all `dataset_expressions` (with a `dataset_expression_format` equals to 6) linked to the dataset.


jcomedouteau (approved):

jcomedouteau:
We need to check if the script is well applied.

Issue #1920, opened 2023-10-23
Mass download on firefox


pierrehenri-dauvergne:
Using firefox
Go to solr
Select any number of dataset and click "download"
Job ends-up in error with message "timeout":

![image](https://github.com/fli-iam/shanoir-ng/assets/117283961/6a319ab9-01a9-4975-a3af-46f0b83425cd)

jcomedouteau:
This is now resolved

PR #1925, opened 2023-10-23
#1920 download browser corrections

jcomedouteau:
How to test:
- Try to download from solr on a study you don't have the download right
- Download from firefox
- Dataset folder is now in a new folder with timestamp.


julien-louis (commented):
:+1:

michaelkain (approved):

PR #1928, opened 2023-10-25
#1920 - remove timeout logic for import and augment timeout

jcomedouteau:
- set timeout to 5mn
- Apply timeout only for download

### DIFF
--- a/shanoir-ng-front/src/app/shared/mass-download/mass-download.service.ts
+++ b/shanoir-ng-front/src/app/shared/mass-download/mass-download.service.ts
@@ -2,17 +2,17 @@
  * Shanoir NG - Import, manage and share neuroimaging data
  * Copyright (C) 2009-2019 Inria - https://www.inria.fr/
  * Contact us on https://project.inria.fr/shanoir/
- * 
+ *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


shanoir-ng-front/src/app/shared/mass-download/mass-download.service.ts
```diff
@@ -157,6 +159,8 @@ export class MassDownloadService {
         } else {
             directoryHandlePromise = this.getFolderHandle();
         }
+        let path :string = "Datasets-" + formatDate(new Date(), "yyyyMMddHHmmss", "en-US") + '/';
```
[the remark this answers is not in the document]
jcomedouteau: Ah yes, I did not think of this. Maybe you were right, this upper folder is not the greatest idea

shanoir-ng-front/src/app/shared/notifications/notifications.service.ts
```diff
@@ -42,7 +42,7 @@ export class NotificationsService {
     readonly readInterval: number = 1000;
     readonly persistenceTime: number = 1800000;
     private freshTimeouts: SuperTimeout[] = [];
-    private readonly TIMEOUT: number = 30000;
+    private readonly TIMEOUT: number = 300000;
```
julien-louis, line 45: why ?

jcomedouteau: We had problem with the timeout, sometimes it takes more than 30sc to be updated.
The other problem was that this was applied for others jobs, taht weren't download, and import always finished as "timeout" before the end of he import

jcomedouteau:
I'm closing this one, as this is integrated in Julien's PR.

PR #1949, opened 2023-11-17
[fli-iam#1948] Add release version tag for docker images

youennmerel:


youennmerel:
Hi @michaelkain,

Yeah, I think we can try that !

michaelkain (approved):
Hi @youennmerel, thank you! As it looks very promising, we will give it a try. With kind regards, Michael

Issue #1958, opened 2023-11-23
[Download]


jcomedouteau:
Dicom download seems to be corrupted with "multipart" text lines => remove them

PR #1959, opened 2023-11-23
#1958-multipart-download-bug

jcomedouteau:
- Extract correctly dicom files after PACS download

How to test:
- Download dicom and open them

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
@@ -14,11 +14,7 @@
 
 package org.shanoir.ng.download;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.StringReader;
+import java.io.*;
 import java.net.MalformedURLException;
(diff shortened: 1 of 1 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -520,6 +510,54 @@ private void extractDICOMFilesFromMHTMLFile(final byte[] responseBody, final Str
...
+	 * @throws IOException
+	 * @throws MessagingException
+	 */
+	private void extractDICOMZipFromMHTMLFile(final byte[] responseBody, final String instanceUID, String name, ZipOutputStream zipOutputStream)
+			throws IOException, MessagingException {
+		try(ByteArrayInputStream bIS = new ByteArrayInputStream(responseBody)) {
+			ByteArrayDataSource datasource = new ByteArrayDataSource(bIS, CONTENT_TYPE_MULTIPART);
+			MimeMultipart multipart = new MimeMultipart(datasource);
+			int count = multipart.getCount();
+			if (count == 1) {
```
michaelkain, line 533: Hi @jcomedouteau, maybe we could always use a for loop? in case count == 1, we loop only once?

michaelkain (approved):
Excellent!

PR #1966, opened 2023-11-24
[CI] Push docker image with version in tag

youennmerel:
Hi @michaelkain,

I've been able to successfully manually run this workflow on my fork (https://github.com/youennmerel/shanoir-ng) by putting mock value (`v1.0`) in the $TAG env variable.

This seems to (finally) work fine :
> image

> image

The only uncertainty now is that `${{ github.ref }}` will give us the right $TAG value on triggered run, but it seems to be clearly the good variable to use : https://github.com/orgs/community/discussions/26686#discussioncomment-3396593

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,7 +4,7 @@ name: Publish Docker images
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ghcr.io/${{ github.repository }}
-  TAG: ${{ github.ref }}
+  TAG: ${{ github.ref_name }}
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


youennmerel:
Hi @michaelkain,

I've been able to successfully manually run this workflow on my fork (https://github.com/youennmerel/shanoir-ng) by putting mock value (`v1.0`) in the $TAG env variable.

This seems to (finally) work fine :
> image

> image

The only uncertainty now is that `${{ github.ref }}` will give us the right $TAG value on triggered run, but it seems to be clearly the good variable to use : https://github.com/orgs/community/discussions/26686#discussioncomment-3396593

michaelkain (approved):
Hi @youennmerel, looks perfect! Great idea for your "test-setup". Looking forward to see it running :-)

PR #1969, opened 2023-11-27
[fli-iam#1805] Set up service-account email at MS users start

youennmerel:


shanoir-ng-users/src/main/java/org/shanoir/ng/ShanoirUsersManagement.java
```diff
@@ -207,4 +218,30 @@ private UserRepresentation getUserRepresentation(final User user) {
...
+	 * Set up the email ${service-account.user.email}
+	 * of the keycloak user ${service-account.user.name} ('service-account-service-account')
+	 * associated with the keycloak client 'service-account'
+	 *
+	 * See service-account.user.* application properties
+	 */
+	private void setVIPServiceAccountEmail(){
+
+		final List<UserRepresentation> userRepresentationList = keycloak.realm(keycloakRealm).users().searchByUsername(this.vipSrvUsername, true);
+		if (userRepresentationList == null || userRepresentationList.isEmpty()) {
```
jcomedouteau, line 231: Small doubt here. Shouldn't we create this user if it does not exists ?

youennmerel: Hi @jcomedouteau,
This user is automatically created with service-account client.
If it does not exists, that mean the service-account is not set.
So i think creating it wouldn't have a lot of meaning here.

PR #1971, opened 2023-11-29
[fli-iam#1867] Restart monitoring of unfinished VIP jobs on startup

youennmerel:
How to test :
- Launch a "big" VIP execution (one that will take several minutes)
- Check that an "Execution monitoring" job has been created
- Stop the Shanoir instance : `dk stop ofsep-qualif*`
- Restart the Shanoir instance : `dk start ofsep-qualif*`
- Reconnect to Shanoir and check that the corresponding job is visible and updated (note that it will probably in a fail state, due to Shanoir shutdown)


shanoir-ng-front/src/app/async-tasks/task.model.ts
```diff
@@ -95,15 +95,15 @@ export class Task extends Entity {
             } else if (this.message.indexOf('dataset [') != -1) {
                 let substring = this.message.match(/dataset \[\d+\]/g)[0];
                 return '/dataset/details/' + substring.slice(substring.lastIndexOf("[") + 1, substring.lastIndexOf("]"));
-            } else if (this.message.indexOf('VIP Execution') != -1) {
-               return '/dataset-processing/details/' + this.objectId
             }
+        } else if (this.eventType === 'executionMonitoring.event' && this.status != -1) {
```
jcomedouteau, line 99: === is not used in java, and I think you can create an event type in ShanoirEventType.java to have the reference. (event status is in a constant on ShanoirEvent I think too)

shanoir-ng-users/src/main/java/org/shanoir/ng/events/ShanoirEventRepository.java
```diff
@@ -24,5 +24,7 @@ public interface ShanoirEventRepository extends CrudRepository<ShanoirEvent, Lon
 	 */
     public void deleteByLastUpdateBefore(Date expiryDate);
 
+	List<ShanoirEvent> findByObjectIdAndEventType(String objectId, String eventType);
```
jcomedouteau, line 27: indentation

jcomedouteau (approved):

PR #1976, opened 2023-12-11
remove deprecated column

julien-louis:
This column is deprecated and only leads to confusion.
It was removed by this script : docker-compose/database/db-changes/datasets/0023_quality_control_modify_study_cards.sql

### DIFF
--- a/docker-compose/database/db-changes/datasets/0037_remove_deprecated_column.sql
+++ b/docker-compose/database/db-changes/datasets/0037_remove_deprecated_column.sql
@@ -1,0 +1,4 @@
+-- the rule_id column has been replaced by a join table and all associations has already been copied in it
+-- see docker-compose/database/db-changes/datasets/0023_quality_control_modify_study_cards.sql
+alter table study_card_condition drop foreign key FKpi8v33fmd0wn64e06q8it8pkv;
+alter table study_card_condition drop column rule_id;


julien-louis:
Also the tester might test this script bu manually executing it on qualif.

It is meant to fix some issues with the enums in existing studycards. Before applying the script, check studycards in the O***P instance. For exemple n°115 : 
image
or n°161 : 
image

(The script doesn't concern the data model, only the data, and will be executed only once, that's why it's not meant to versioned) 

script : 
```
-- in study_card_condition_join, delete references of rules that have no references in any assignment
-- (having no assignment make a study_card_rule useless)
-- (quality cards have another join table : quality_card_condition_join)
delete from study_card_condition_join where study_card_rule_id in (select id from study_card_rule where id not in (select rule_id from study_card_assignment));

-- delete rules that have no references in any assignment
delete from study_card_rule where id not in (select rule_id from study_card_assignment);

-- change modality type MR to MR_DATASET as it is the standard in Shanoir (see DatasetModalityType.java) (1 is defined in DatasetMetadataField.java)
update study_card_assignment set value = "MR_DATASET" where field = 1 and value like "MR";

-- fix the case for spin density
update study_card_assignment set value = "SPIN_DENSITY" where field = 7 and lower(value) like "spin_density" and value not like binary "SPIN_DENSITY";
```

michaelkain (approved):
Hi @julien-louis, the code review looks perfect, thank you! I approve and check on qualif during the test session, if it applies accordingly. With kind regards, Michael

PR #1983, opened 2023-12-13
refactoring in quality check on partial exam at import

julien-louis:
This is a fix of a bad code design from my last quality check pr. Here, instead of temporary replacing the acquisitions of an exam to apply it only on the newly imported data, I use a completely distinct class from Examination.


jcomedouteau:
To be reviewed by @LaurentPV

LaurentPV (approved):
Pas de regression constatée. Develop a merger sur la branche avant de merge.

PR #1998, opened 2023-12-15
if missing sop instance uid, generate one

julien-louis:


shanoir-ng-import/src/main/java/org/shanoir/ng/importer/dicom/ImagesCreatorAndDicomFileAnalyzerService.java
```diff
@@ -282,6 +283,10 @@ private void addImageSeparateDatasetsInfo(Image image, Attributes attributes) th
 			MultiframeExtractor emf = new MultiframeExtractor();
 			attributes = emf.extract(attributes, 0);
 		}
+		String sopInstanceUID = attributes.getString(Tag.SOPInstanceUID);
+		if (sopInstanceUID == null || sopInstanceUID.isEmpty()) {
+			sopInstanceUID = "GeneratedSopInstanceUID." + RandomStringUtils.randomAlphabetic(26);
```
michaelkain, line 288: Hi Julien, please use what is in shanoir-ng-anonymization to create a sopInstanceUID. I am wondering, if the above will even enter into the pacs. Thank you, Michael

michaelkain, line 288: Hi Julien, please use what is in shanoir-ng-anonymization to create a sopInstanceUID. I am wondering, if the above will even enter into the pacs. Thank you, Michael

jcomedouteau:
This should be closed as already resolved.

Issue #1943, opened 2023-11-09
Web GUI (Angular), CT Import: centers: find existing and add to study or creation: use DICOM information


michaelkain:
Hi @pierrehenri-dauvergne, I hope you are doing fine.
As you know, I am currently working on ShUp, imports and auto-creation of SCs, centers, equipments.
I made a test with CT data. We display in the clinical context the center information coming from the
DICOM, see my screenshot. 1) It would be great to pre-select a matching center if it exists, and add it
to the study, in the worst case (CHRU Lille already exists in db as center) 2) propose a new center creation, but use the DICOM information when I click on create center, where today everything is empty, see my next screenshot.
I will implement the same behavior on ShUp.
With kind regards, Michael

michaelkain:
by the way, for the creation of the acquisition equipment, the serial number is pre-filled from the dicoms

PR #2011, opened 2024-01-10
shanoir-issue#1943: send dicom data to pre-fill center and acq equipment forms during imports

pierrehenri-dauvergne:
Fix issue #1943

I haven't been able to test the import of a CT Dicom that contains a device serial number (I don't have the data).

How to test:
Import a DICOM
Select a zip file of CT dataset
Select the CT series
On step 3 of import, add a center
The 'create center' form should be pre-filled with the DICOM's data (InstitutionName and InstitutionAddress) if not empty
Once the center is selected, create an acquisition equipment, it should also be pre-fill with the serial number (Device Serial Number in DICOM metadata)


shanoir-ng-front/src/app/acquisition-equipments/acquisition-equipment/acquisition-equipment.component.ts
```diff
@@ -112,6 +115,10 @@ export class AcquisitionEquipmentComponent extends EntityComponent<AcquisitionEq
     }
 
     buildForm(): UntypedFormGroup {
+        if (this.fromImport) {
```
julien-louis, line 118: I think It doesn't belong here, buildForm is for building the form object.

Imo it would be better in initCreate() or even in prefill() which is called by initCreate()

shanoir-ng-front/src/app/centers/center/center.component.ts
```diff
@@ -66,6 +69,10 @@ export class CenterComponent extends EntityComponent<Center> {
     }
 
     buildForm(): UntypedFormGroup {
+        if (this.fromImport) {
```
julien-louis, line 72: better in initCreate()

julien-louis (changes requested):

shanoir-ng-front/src/app/import/clinical-context/clinical-context.abstract.component.ts
```diff
@@ -483,6 +483,10 @@ export abstract class AbstractClinicalContextComponent implements OnDestroy, OnI
         studyCenter.study = this.study;
         let newCenter = new Center();
         newCenter.studyCenterList = [studyCenter];
+        if (this.importedCenterDataStr != null) {
```
julien-louis, line 486: :+1:

julien-louis (approved):

PR #1970, opened 2023-11-27
#1968-Download-refactoring Refactor code to better manage dataset download

jcomedouteau:

PR #2013, opened 2024-01-10
Shanoir issue#1968 download refactoring

julien-louis:
this replaces https://github.com/fli-iam/shanoir-ng/pull/1970


shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -399,26 +351,25 @@ public AcquisitionAttributes<Long> getDicomAttributesForAcquisition(DatasetAcqui
 		return dAcquisitionAttributes;
 	}
 
-	static String extractInstanceUID(String url) {
```
julien-louis, line 402: @jcomedouteau I think StringUtils have better performances than regexp here. I think I even measured it.

Do you think the code was malfunctioning ?

jcomedouteau: No, I think it's a mistake from the merge (or from my PR I'm not sure), but we need to keep your code !

julien-louis:
Ok I have tested and I approve the pr

PR #2015, opened 2024-01-12
add check of zip 500 limit

julien-louis:


julien-louis:
@pierrehenri-dauvergne I think it's fixed now

pierrehenri-dauvergne:
How I test (on chrome and windows):
On solr, filter the datasets by study: ucan and by subject: ucan_16_025
I select the first 100 datasets and click "download"

I checked without checking "unzip" and still got 'file_path_too_long' issue on 4 datasets
I tried checking "unzip" and got 100 errors with file_path_too_long'
I tried checking "unzip" and checking "short path" and still got "file_path_too_long' error for 7 datasets (id 445886, 445911, 495584, 495604, 495620, 495622 and 495638).

pierrehenri-dauvergne:
Tested on qualif:
Filter by study and subject on solr, select 100 datasets and click 'download'

**On firefox** : "missing start boundary" and "Could not download dataset as dicom" errors. Job page says every dataset was downloaded but error.json file says otherwise

**On chrome** : 67 OK and 33 errors ("file path too long" or "missing boundary")
-- check 'unzip' : 0 OK and 100 errors ("file path too long" or "missing boundary")
-- check 'unzip' and 'short path' : 70 OK and 30 errors ("missing boundary" only)

julien-louis:
after testing with @pierrehenri-dauvergne :  

* "missing boundary" is an error from the back, currently on investigation in https://github.com/fli-iam/shanoir-ng/issues/2128 and in the pr https://github.com/fli-iam/shanoir-ng/pull/2129
* "file path too long" sometimes can not be avoided but there is maybe still something to do : a short name option for zip files (already available for unzip option)
* when dl with firefox, the presence of a ERROR.json file in the zip is not reported in jobs, will add that

julien-louis:
@pierrehenri-dauvergne I've made the two upgrades

pierrehenri-dauvergne (approved):

PR #2024, opened 2024-01-17
remove sequence number duplicates in db_changes

julien-louis:


julien-louis:
I think the check doesn't accept changes in old .sql

PR #2045, opened 2024-01-26
Less task requests

julien-louis:
instead of reloading all tasks at every task update, just update the one task

Please check the tasks behaviour with every task related actions :
* download
* quality check
* import
* copy datasets

check : 
* left menu current tasks
* tasks table is live updated
* task details is live updated

if you look at the network tab in debug tools, the number of request and the volume of data will be reduced by this pr


julien-louis:
I first tested it with dataset copy tasks, wich is not merged yet but therer was a bug with dates (dates missing), I will have to fix that in time

pierrehenri-dauvergne:
Tested this PR in my local env and there are less task request for all (copy datasets, imports, downloads, quality checks) but:

- Copy datasets: the 'creation date' and 'last update date' fields are empty
- Imports: same
- Downloads: progress shows "%NaN" for a while at the begining of download (on firefox) + some errors in console (ZipPacsFileException: Received 404 NOT_FOUND from PACS)
- Quality check: also missing creation/last update fields and I got an error at 96% progress saying "setting quality subject tags" and console says "data too long for column 'report'."

image

julien-louis:
@pierrehenri-dauvergne so the problem you had when the datasets weren't check is imo not linked to this pr, I have create an issue. https://github.com/fli-iam/shanoir-ng/issues/2152

Is there other problems ?

pierrehenri-dauvergne (approved):

PR #2053, opened 2024-01-30
Correct execution status typo

youennmerel:


youennmerel:
see https://github.com/fli-iam/shanoir-ng/pull/2054

PR #2061, opened 2024-02-01
Qualitycards/studycards dicom values validation

julien-louis:


julien-louis:
Now the failed import email sent to the study admins has a link to the studycard, please test it.

PR #2062, opened 2024-02-02
[Fli iam#2059] Use new backend execution & pipeline APIs in Shanoir front

youennmerel:
# Changes
- All VIP API call are now made through MS datasets backend
  - `.../datasets/vip/pipeline`
  - `../datasets/vip/execution` 
- Use Spring 5 Web Client (see VipClientService.java)
- Write VIP handler logs into shanoir-vip-handler.log

# How to test
- Deploy on OFSEP qualif
- Check that datasets Swagger doc is up to date
- Execute OFSEp Seq Id & ct-tiqua VIP pipelines on multiple datasets through 
   - Shanoir website
   - shanoir-downloader Python script
- Check that execution status is correctly updated
- Check that execution logs (stdout, stderr) can be fetched in execution monitoring view 
- Check that result are well processed
- Check that handler logs are written in /var/log/dk/ofsep/qualif/shanoir-ng/shanoir-vip-handler.log


michaelkain:
Hi @youennmerel, thank you for this PR, but please remove the RestTemplate usage. In comparison to HttpClient and WebClient, it is definitely the weakest implementation. RestTemplate will be removed in a short time: https://www.baeldung.com/spring-5-webclient

youennmerel:
Hi @jcomedouteau, @michaelkain,

Looking forward for your reviews !

Maybe J-C can check the VIP part, and Michael look at the Spring 5 Web Client implementation ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/controller/VipClientService.java
```diff
@@ -0,0 +1,223 @@
...
+    private String vipPipelineUrl;
+
+    @Autowired
+    private KeycloakServiceAccountUtils keycloakServiceAccountUtils;
+
+    private WebClient webClient;
+
+    @PostConstruct
+    public void init() {
+        this.webClient = WebClient.create();
```
michaelkain, line 43: Hi Youenn, I hope you are doing fine and thank you for your PR!
I would propose, if possible to add .baseUrl("https://vip"); after the
create and work with get(uri), either executions or pipelines.
In the idea, as you always call the same server, that the opened
http connection (url + port -> socket) is re-used again and again.
To assure this, it might be worth going by baseUrl, just an idea.
Maybe WebClient is already so smart internally. With kind regards,
Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/dto/ExecutionCandidateDTO.java
```diff
@@ -0,0 +1,116 @@
...
+    private String identifier;
+    private String name;
+    private String pipelineIdentifier;
+    private Map<String, List<String>> inputParameters;
+    private List<DatasetParameterDTO> datasetParameters;
+    private Long studyIdentifier;
+    private String outputProcessing;
+    private String processingType;
+    private String refreshToken;
+    private String client = "shanoir-ng-front";
```
michaelkain, line 34: Hi, does it play a role, to fix this here by default, if the same DTO is send by the Python script potentially? Is the Python script a "simulated" front-end in your current setup?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/controller/ExecutionApi.java
```diff
@@ -0,0 +1,103 @@
...
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "Execution successfully created response."),
+            @ApiResponse(responseCode = "403", description = "forbidden"),
+            @ApiResponse(responseCode = "500", description = "unexpected error"),
+            @ApiResponse(responseCode = "503", description = "error from VIP API")})
+    @RequestMapping(value = "/",
+            produces = {"application/json"},
+            consumes = {"application/json"},
+            method = RequestMethod.POST)
+    ResponseEntity<IdName> createExecution(
```
michaelkain, line 55: Hi, should we put at least a role protection on the API methods?

jcomedouteau: The roles are checked in the method, but we should actually put it here into a specific class

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/dto/ExecutionCandidateDTO.java
jcomedouteau: The python script is a simulated shanoir-uploader client if I'm not wrong. But if not precised, then we are coming from the front, so shanoir-ng-front is the default.

youennmerel: I agree with Michael that a default value is not really useful here. I removed it, and added @NotNull annotations on mandatory fields

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/controller/ExecutionApi.java
youennmerel: In fact, as documented [here](https://github.com/fli-iam/shanoir-ng/wiki/Roles-and-rights-specification), the required rights on datasets included in execution is `CAN_ADMINISTRATE`.

I've corrected this and extracted the check into a `checkRightsForExecution` method. If relevant, this will throw a proper unauthorized RestServiceException with an explicit message.

jcomedouteau:
@michaelkain the tests are Ok for me

PR #2066, opened 2024-02-06
shanoir-issue#966: add user feedback

pierrehenri-dauvergne:
How to test:

As an expert
Go to 'pathology models'
Create a new pathology model with an already existing name
The form should display a message saying 'Name should be unique'
Do the same with therapies

### DIFF
--- a/shanoir-uploader/pom.xml
+++ b/shanoir-uploader/pom.xml
@@ -22,6 +22,8 @@
 		<junit.jupiter.version>5.8.2</junit.jupiter.version>
 		<junit.platform.version>1.8.2</junit.platform.version>
 		<poi.version>4.1.2</poi.version>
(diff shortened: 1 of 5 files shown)


shanoir-uploader/pom.xml
```diff
@@ -22,6 +22,8 @@
 		<junit.jupiter.version>5.8.2</junit.jupiter.version>
 		<junit.platform.version>1.8.2</junit.platform.version>
 		<poi.version>4.1.2</poi.version>
+		<maven.build.timestamp.format>yyyy-MM-dd'T'HH-mm-ss'Z'</maven.build.timestamp.format>
```
jcomedouteau, line 25: Is it part of the correction ?

pierrehenri-dauvergne: No, it's so shanoiruploader builds on windows (can't have path with ":" in it). I'll take this off

jcomedouteau (approved):

PR #2107, opened 2024-02-26
Wado url handler

michaelkain:
Hi,
please test: deletion from pacs + downloads (mass) + quality cards


shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -351,17 +331,13 @@ public AcquisitionAttributes<Long> getDicomAttributesForAcquisition(DatasetAcqui
...
-		boolean condition2 = url != null && url.contains("instances/");
-		if (condition1) {
-			String[] split = StringUtils.splitByWholeSeparator(url, "objectUID=", 2);
-			return StringUtils.split(split[1], "&", 1)[0];
-		} else if (condition2) {
-			String[] split = StringUtils.splitByWholeSeparator(url, "instances/", 2);
-			return StringUtils.split(split[1], "/", 1)[0];
-		} else return null;
+
+	public WADOURLHandler getWadoURLHandler() {
```
jcomedouteau, line 335: Do you really need this if it is autowired ?

jcomedouteau (commented):
Thanks for the refactoring, I'll be testing it!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/WADOURLHandler.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+@Component
+public class WADOURLHandler {
+
+	private static final String WADO_URI = "requestType=WADO";
+	
+	private static final String CONTENT_TYPE = "&contentType";
+
+	public boolean isWADO_URI(String url) {
+		if (url.contains(WADO_URI)) {
```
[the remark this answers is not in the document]
michaelkain: done

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
michaelkain: Unfortunately, yes! as the ExaminationAttributes class of Julien is a none-spring component and it requires the extract code as well, I found it more elegant to init the wado-url-handler by Spring into downloader and get it from there instead passing it through the methods.

michaelkain:
Hi @jcomedouteau, thank you for your review! Great! Could you maybe test with WADO-URI links on qualif? I tested and fixed locally with dicom.web=true and WADO-RS links, and it seems to work. But I want to avoid the effort, to change the switch to false and import again, and download/delete again, if you might testing on qualif anyway? With kind regards, Michael

shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
jcomedouteau: Could it be a static class at this moment ?

michaelkain:
I had it yesterday, and fixed it this morning.

jcomedouteau (approved):
Import worked
Deletion worked (checked on qualif PACS)
Download worked

Fine for me.

michaelkain:
Thank you Jean-Come!

PR #2111, opened 2024-02-27
[fli-iam#2095] Delete linked properties before deleting dataset

youennmerel:
# Changes
- Delete all DatasetProperty linked to dataset
- Remove dataset from inputDatasets list of all linked DatasetProcessing
- If DatasetProcessing is linked to no other dataset, delete it
- Delete processed dataset attached to DatasetProcessing
- Delete child dataset processing

# How to test
Delete exams / acquisition / datasets for which exists is VIP execution results


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -135,15 +135,10 @@ private void initialize() {
...
 		try {
 			datasetService.deleteById(datasetId);
 			return new ResponseEntity<>(HttpStatus.NO_CONTENT);
-		} catch (EntityNotFoundException e) {
-			return new ResponseEntity<>(HttpStatus.NOT_FOUND);
-		} catch (RestServiceException e) {
-			return new ResponseEntity<>(HttpStatus.UNPROCESSABLE_ENTITY);
 		} catch (Exception e) {
 			LOG.error("Error while deleting dataset. Please check DICOM server configuration.", e);
 			return new ResponseEntity<>(HttpStatus.NO_CONTENT);
```
jcomedouteau, line 144: Not your code obviously, but I think it should be more logic to return "HttpStatus.INTERNAL_SERVER_ERROR" as in deleteDatasets and not the same HTTP response as if the deletion was a success

youennmerel: Hi @jcomedouteau,

Thx for your remarks.
I've harmonized exception management for both methods.

In fact, `EntityNotFoundException` | `RestServiceException` | `Exception` will be intercepted in class `GlobalExceptionHandler` of ms-common, and properly returned as a `ResponseEntity<>`.

So for both method :
- I rethrow `EntityNotFoundException` | `RestServiceException` 
- I encapsulate others exception in a `RestServiceException` and throw it.

All right for you ?

jcomedouteau:
I have a weird error on your PR when trying to delete a dataset. Seems to be linked to @Transactional annotations:
image

youennmerel:
Hi @jcomedouteau,

I've made deleteById transactional, it seems all right.

jcomedouteau (approved):
Now working correctly

PR #2114, opened 2024-02-28
Write VIP handler logs into shanoir-vip-handler.log

youennmerel:


youennmerel:
Will be done in https://github.com/fli-iam/shanoir-ng/pull/2062

PR #2116, opened 2024-02-28
shanoir-issue#2078: add a check if user isnt part on any study

### LINKED ISSUE #2078: MS Studies: storageVolume: Angular exception with user, that has no study yet
Hi Youenn,
I am testing and I wonder if this is a frontend bug, we have not yet detected:
user type expert, account created, but not yet in any study, so his study list is empty
Error: Uncaught (in promise): TypeError: t is undefined
fetchStorageVolumesByChunk
It is a minor bug, but it would be great to solve it please.

pierrehenri-dauvergne:
Fix #2078 

How to test:
Create a new user
Login with it
Open the study list

Before: an error was thrown
Now: no error


jcomedouteau (approved):
Works well on qualif, thanks!

PR #2122, opened 2024-02-29
Update ShanoirEvent.java

michaelkain:
Fixes:
Error executing DDL "create table events (progress float(23), status integer not null, creation_date datetime(6), id bigint not null, last_update datetime(6), study_id bigint, user_id bigint, message varchar(32600), report varchar(32600), event_type varchar(255), object_id varchar(255), primary key (id)) engine=InnoDB" via JDBC [Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs]


michaelkain:
Hi @jcomedouteau, I have only tested it locally. Could you please test if it works on qualif? Thank you and with kind regards, Michael

jcomedouteau (approved):
Working well on qualif

PR #2165, opened 2024-03-27
[fli-iam#2093] Make Swagger UI fully functional

youennmerel:


youennmerel:
Will be done in https://github.com/fli-iam/shanoir-ng/pull/2179

PR #2182, opened 2024-04-09
filter tasks by user

julien-louis:
I create ```public class UserSseEmitter extends SseEmitte``` and added ```userId``` in it


julien-louis:
currently it blocks every task

julien-louis:
as jdouteau said, cannot get je keycloak token from this context (also it wasn't making any sense !), use the event userId instead

jcomedouteau (approved):
Ok I tested the 3 events that are displayed:
- Import
- Download (even if it's not the same logic)
- dataset copy
Everything seems to work properly.

Issue #2103, opened 2024-02-26
Remove unused study fields


pierrehenri-dauvergne:
Fields "Is with examination", "Data accessibility" are not used ?

Rename "visible" to "display as public study" ?

Etc.

PR #2200, opened 2024-04-23
shanoir-issue#2199: add studycard policy to study details + save in database

pierrehenri-dauvergne:
How to test:
Create or update a study
Change the value of "study card policy"
Check in database that it was stored

This PR only adds a new field, it is not used yet.

This PR also fixes https://github.com/fli-iam/shanoir-ng/issues/2103

### DIFF
--- a/docker-compose/database/db-changes/studies/0020_set_default_studycardpolicy.sql
+++ b/docker-compose/database/db-changes/studies/0020_set_default_studycardpolicy.sql
@@ -1,0 +1,1 @@
+UPDATE study SET study_card_policy=0;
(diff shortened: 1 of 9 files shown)


docker-compose/database/db-changes/studies/0020_set_default_studycardpolicy.sql
```diff
@@ -0,0 +1 @@
+UPDATE study SET study_card_policy=0;
```
jcomedouteau, line 1: I think you have to add the column too as it does not exists on qualif/prod yet.
( ALTER TABLE study ADD COLUMN study_card_policy [...])

jcomedouteau (approved):

PR #2209, opened 2024-05-02
Dockerfile optimisations

a-ba:
Here is a series of optimisations to reduce the build time and image sizes thanks to various buildkit features. I documented them in `docker-compose/README.md`

With these optimisations, building the images with *docker buildx bake* on the github runners roughly takes: 
- 10 minutes for the initial build
   ```
  468.1s  [nifti-conversion]
  142.3s  [datasets]
   84.2s  [keycloak]
   81.3s  [preclinical]
   59.0s  [studies]
   51.6s  [import]
   48.2s  [users]
   38.3s  [keycloak-database]
   35.3s  [solr]
    8.9s  [nginx]
    2.0s  [database]
   ```
- 50 seconds for the next builds (minimum time when the dockerfile are unchanged)
   ```
   33.4s  [nifti-conversion]
   31.2s  [keycloak]
   30.2s  [preclinical]
   23.0s  [datasets]
   20.9s  [studies]
   18.3s  [users]
    5.9s  [import]
    3.6s  [solr]
    1.7s  [nginx]
    0.8s  [database]
    0.0s  [keycloak-database]
   ```

Now the new bottleneck is the maven build (around 2m30).

PR #2218, opened 2024-05-21
Switch to java 21 using temurin package

jcomedouteau:


docker-compose/datasets/Dockerfile
```diff
@@ -26,7 +27,11 @@ RUN apt-get update -qq \
     gzip \
     locales \
     locales-all \
-    wget
+    wget \
+    apt-transport-https \
+    gpg \
+    adoptium-ca-certificates \
+    temurin-21-jdk
```
a-ba, line 34: (note: these comments apply to all images)

you should install `temurin-21-jre` rather than `temurin-21-jdk` (we only need the JRE in the docker images)

`openssl` and `adoptium-ca-certificates` are already listed as dependencies of `temurin-21-(jre|jdk)`, there is no need to list them here (unless we exclicitly use them, but i am pretty sure we don't).

I would rather suggest to delay this PR until #2209 is merged because it will highly conflict with it (also #2209 makes a point in *not* installing build time dependencies (like wget, gpg or jdk) in the final images).

jcomedouteau: Okay, we'll wait for your PR to be merged and I'll rebase myself on it. Thanks !

a-ba:
should be closed in favor of #2227

jcomedouteau:
duplicate #2227

PR #2226, opened 2024-05-23
merge all dockerfiles into a single dockerfile

a-ba:
Using a single Dockerfile allows sharing intermediate images between multiple builds, thus removing duplicated code (installation of the jvm, entrypoint_common/oneshot scripts, common entrypoint definition, ...)


jcomedouteau:
@a-ba I have this when deploying on qualif:
ERROR: failed to solve: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount540963045/Dockerfile: no such file or directory

Do you have to adapt something ?

a-ba:
right, the `build-shanoir-ng` script needed an update.
This should be fixed now.

jcomedouteau (approved):

PR #2227, opened 2024-05-23
Java 21

a-ba:
migration to java 21, using the temurin package

Note: this PR depends on #2226, which should be reviewed and merged first.


a-ba:
I added the migration to java 21 in the bootstrap script and rebased onto develop

julien-louis (approved):
running well on qualif

PR #2229, opened 2024-05-24
shanoir-issue#2228: remove condition 'is admin' to make a study a challenge

pierrehenri-dauvergne:
How to test:
As an expert, create a study
Make sure to have the admin rights
Edit study
Check the "challenge" checkbox and save

Before this PR, we had to be an admin to make it a challenge, now an expert can do it.


michaelkain (approved):
Hi @pierrehenri-dauvergne, thank you for this fast PR!
I validate it by code review only, as it compiles and looks perfect.
With kind regards, Michael

Issue #2256, opened 2024-06-12
Missing MrSequenceVariant: OS


jcomedouteau:
Unexpected error during the import: class java.lang.IllegalArgumentException : No enum constant org.shanoir.ng.datasetacquisition.model.mr.MrSequenceVariant.OS, please contact an administrator.

jcomedouteau:
Closed by #2261

PR #2261, opened 2024-06-13
#2256-missing-mr-sequence-variant

jcomedouteau:
Closing #2256 

To avoid further Errors, return UNKNOWN when not found

How to test (not that simple)
- Open a dicom with pydicom and update mr-sequence-variant field to set something not in the list
- Import this data
--> Check that this metadata is set to UNKNOWN in shanoir


shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/model/mr/MrSequenceVariant.java
```diff
@@ -95,7 +95,11 @@ public static MrSequenceVariant getIdByType(final String type) {
 		if (TOF.equals(type)) { // see GitHub issue #561
 			return MrSequenceVariant.SS;
 		}
-		return MrSequenceVariant.valueOf(type);
+		try {
+			return MrSequenceVariant.valueOf(type);
+		} catch (IllegalArgumentException exception) {
+			return UNKNOWN;
```
youennmerel, line 101: You could use EnumUtils from apache-commons :
``` java
return EnumUtils.isValidEnum(MrSequenceVariant.class, type) : MrSequenceVariant.valueOf(type) ? UNKNOWN
```
The `isValidEnum` method use the same try-catch method as your solution, though.

youennmerel:
```python
$ pip install pydicom
$ python
>>> from pydicom import dcmread
>>> ds = dcmread('/home/ymerel/Documents/samples/KO_dcm_sample/IM000000.dcm')
>>> elem = ds['MRSequenceVariant']
>>> elem.value = ['KO']
```

youennmerel (approved):
Hi @jcomedouteau,

Tested on qualif, everything fine !

> image

PR #2263, opened 2024-06-14
dicom vr ds & di better vm management (slice thickness is float array bug)

julien-louis:
in study cards & quality cards conditions, Slice Thickness (and others) dicom tag should not be <FloatArray> but just <Float>.

The problem was that the dicom Value Representation (VR) named Decimal String (DS) can be arrays or single values so I changed the conversion rules.


LaurentPV (approved):
Parfait !

Issue #2121, opened 2024-02-29
[BUG] Loading examination tree loads too much elements


jcomedouteau:
When loading examination detail, we automatically load the examination tree.
When some datasets have processing, we carry too much informations about this processing. 

The problem is particularly present when the examination has a lot of datasets, and every dataset of the exam is part of a single processing.
It means that for every dataset, we load the processing with all the parents.
With an examination of 3K datasets, we load ~9M json objects (up to ~500Mo).

We have to simplify this (load only id, name in the tree ? Do not load parent datasets of a processing ?)

Issue #2275, opened 2024-06-24
[Download] Processed dataset as EEG


jcomedouteau:
- Process dataset download sometimes fails (EEG)
- Workfolder is set to undefined-undefined
- processing input are not correctly displayed in the processing view

jcomedouteau:
Closed by #2278

PR #2278, opened 2024-06-24
#2275-processed-dataset-bug

jcomedouteau:
- Correctly download EEG processed dataset
- Correctly display input dataset no processing detail
- Correctly set processed dataset folder when downloading

How to test:
Deploy on OFSEP qualif (please)
- Import a processed dataset (select a nifti) and select EEG as dataset type.

--> Dataset can be downloaded
--> Path in the donwload folder is not undefined-undefined anymore

Go on processing detail:
--> Input datasets are correctly loaded.

Go on examination 52384
--> Check that the loading time of the tree is not >9000 !


jcomedouteau:
Closes #2121 
Closes #2278

pierrehenri-dauvergne (commented):
Downloading the dataset works fine
I didn't see any 'undefined-undefined' path during my tests
The processing details were OK

One problem though, in solr, some fields are missing:
image

pierrehenri-dauvergne (approved):

Issue #2285, opened 2024-07-01
Study edition is very long 


jcomedouteau:
Study edition takes way too much time. (study tags edition ?)

PR #2286, opened 2024-07-01
[Test ]#2285-study-edition-bug

jcomedouteau:
- Logs


jcomedouteau:
Only for tests purpose, corrected by youenn.

PR #2291, opened 2024-07-03
shanoir-issue#2248: study's history

pierrehenri-dauvergne:
How to test:
Open a study and its "History" tab
Make an import into the study
With another user, make an access request to this study and grant it from a study admin

Should see something like this:
image

Some usernames are empty (92233740 in db), probably because some calls where done throught rabbitmq, which change authentication data. I searched the username from the userId and there are no username for this id.


shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/amqp/RabbitMQDatasetsService.java
```diff
@@ -500,6 +501,7 @@ public void copyDatasetsToStudy(final String data) {
 					countProcessed + " are processed datasets and cannot be copied.");
 			event.setStatus(ShanoirEvent.SUCCESS);
 			event.setProgress(1.0f);
+			event.setStudyId(studyId);
```
jcomedouteau, line 504: If the event is already published once, you don't need to put it again.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/EegImporterService.java
```diff
@@ -98,6 +98,7 @@ public void createEegDataset(final EegImportJob importJob) throws IOException {
                 progress += 1f / importJob.getDatasets().size();
                 event.setMessage("Dataset " + datasetDto.getName() + " for examination " + importJob.getExaminationId());
                 event.setProgress(progress);
+                event.setStudyId(importJob.getStudyId());
```
jcomedouteau, line 101: If the event is already published once, you don't need to put it again.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/EegImporterService.java
```diff
@@ -197,6 +198,7 @@ public void createEegDataset(final EegImportJob importJob) throws IOException {
             event.setMessage("[" + importJob.getStudyName() + " (n°" + importJob.getStudyId() + ")]"
                     +" Successfully created datasets for subject [" + importJob.getSubjectName()
                     + "] in examination [" + examination.getId() + "]");
+            event.setStudyId(importJob.getStudyId());
```
jcomedouteau, line 201: If the event is already published once, you don't need to put it again.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/EegImporterService.java
```diff
@@ -206,6 +208,7 @@ public void createEegDataset(final EegImportJob importJob) throws IOException {
             event.setStatus(ShanoirEvent.ERROR);
             event.setMessage("An unexpected error occured, please contact an administrator.");
             event.setProgress(-1f);
+            event.setStudyId(importJob.getStudyId());
```
jcomedouteau, line 211: If the event is already published once, you don't need to put it again.

shanoir-ng-front/src/app/studies/study-history/study-history.component.ts
```diff
@@ -0,0 +1,87 @@
...
+    ngOnInit() {
+        this.eventHistory.then( () => this.onFetchHistory());
+    }
+
+    public onFetchHistory() {
+        this.shanoirEventService.requestHistory(this.study.id).then(history => {
+
+            history.forEach(item => {
+                let studyUser : StudyUser;
+                studyUser = this.study.studyUserList.find(user => user.userId == item.userId);
```
jcomedouteau, line 73: Here you get the user for every event, maybe you could create a Map<userId,username> at the beginning of the method then  re-use it when iterating over the events.

shanoir-ng-users/src/main/java/org/shanoir/ng/events/EventsApi.java
```diff
@@ -0,0 +1,34 @@
...
+    @Operation(summary = "", description = "If exists, returns the events corresponding to the given study id")
+    @ApiResponses(value = { @ApiResponse(responseCode = "200", description = "found user"),
+            @ApiResponse(responseCode = "401", description = "unauthorized"),
+            @ApiResponse(responseCode = "403", description = "forbidden"),
+            @ApiResponse(responseCode = "404", description = "no user found"),
+            @ApiResponse(responseCode = "500", description = "unexpected error") })
+    @RequestMapping(value = "/{studyId}", produces = { "application/json" }, method = RequestMethod.GET)
+    @PreAuthorize("hasRole('ADMIN') or (hasAnyRole('USER', 'EXPERT'))")
+    ResponseEntity<List<ShanoirEvent>> findEventsByStudyId(
+            @Parameter(name = "id of the study", required = true) @PathVariable("studyId") Long studyId) throws RestServiceException;
```
jcomedouteau, line 30: I thought we said that we were going to use Pageing loading.
There is potentially more than 100k events for a single study in shanoir (not yet, but as you added study id everywhere it will change), we need to load page by page.
So the research will be done as in examination list, field by field.

jcomedouteau (changes requested):
Hello @pierrehenri-dauvergne,
Some small remarks and a big one, as I think we need to use pagination.
See examination list page to see how it"s done!

jcomedouteau:
Hey @pierrehenri-dauvergne !
Some fields seems to be empty and maybe should not ? (event id 1927403604474481477 for example)

image

docker-compose/database/db-changes/users/0005_create_study_user_tables.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE `study_user` (
```
jcomedouteau, line 1: I think you need to fill these tables:
image

Otherwise, the table will be empty for other users than admins

jcomedouteau (approved):

jcomedouteau:
Working fine on qualif after review, thanks !

PR #2328, opened 2024-07-25
#2327-download-processed-dataset-chrome

jcomedouteau:
How to test:

- Download a processed dataset on chrome


youennmerel (approved):
OK for all downloads :
- (multiple) datasets
- (multiple) processed datasets
- both

PR #2336, opened 2024-07-29
Shanoir issue#2284: solr filter importDate

pierrehenri-dauvergne:
How to test:
Go to solr page
Open "import date" filter and try it out
The solr result should be filtered accordingly

An endDate value set before the startDate should display an error:
image


LaurentPV (approved):
Tested and approved

Issue #2345, opened 2024-08-01
Error on download on OFSEP


youennmerel:
With examination 1153573.

Import is OK, but when exporting examination in DICOM, there's `Received 404 NOT_FOUND from PACS` errors on the following datasets :
1153616
1153618
1153593
1153625
1153597
1153567
1153569
1153635
1153570
1153573
1153574
1153576
1153608
1153613

Same when downloading said datasets individually.

PR #2353, opened 2024-08-07
Fix for GitHub issue #2345

michaelkain:
Problem: some series, e.g. QSM_3DMULTIGRE_B contain up to 960 DICOM files.
They are organized in multiple directory and packaged by 500 per directory.
MS Datasets searched until this fix the parent directory of the first DICOM file
as reference and sends all containing DICOM files to the server. In the above
case as the DICOM files were distributed over 3 directories only a part of the
files has been send to the PACS. The problem is vendor dependent, as normally
all is structured per serie and concerns only the web import via file, not ShUp,
as in ShUp all DICOM files are queried in only one serie specific folder.

Testing: use serie above and verify, that preview + download work.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomPersisterService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomPersisterService.java
@@ -58,47 +58,46 @@ public class DicomPersisterService {
 	private DICOMWebService dicomWebService;
 
 	/**
-	 * This method reads the datasets for each serie from the json (String), gets
-	 * the dicom expression format and sends the images to the PACS.
+	 * This method reads the datasets for each serie from the json (String),
+	 * gets the dicom expression format and sends the images to the PACS.
 	 * 
 	 * @param serie
 	 * @throws Exception
 	 */
 	public void persistAllForSerie(Serie serie) throws Exception {
-		if (serie != null) {
-			for (Dataset dataset : serie.getDatasets()) {
-				for (ExpressionFormat expressionFormat : dataset.getExpressionFormats()) {
-					if (expressionFormat.getType().equals("dcm")) {
-						List<DatasetFile> datasetFiles = expressionFormat.getDatasetFiles();
-						if (datasetFiles != null && !datasetFiles.isEmpty()) {
-							// Iterate over all dataset files (DICOM files), in case all files might be
-							// distributed over multiple parent directories and therefore require multiple
-							// calls to sendDicomFilesToPacs (advantage: do not create to big multiparts)
-							File lastParentFile = null;
-							for (DatasetFile datasetFile : datasetFiles) {
-								String path = datasetFile.getPath();
-								if (path != null && !path.isEmpty()) {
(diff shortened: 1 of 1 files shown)


youennmerel (commented):
Hi @michaelkain,

I think "ifs" should be inverted in the whole method here, to limit cyclomatic complexity and improve readability :
``` java
public void persistAllForSerie(Serie serie) throws Exception {

	if (serie == null) {
		return;
	}

	for (Dataset dataset : serie.getDatasets()) {
		for (ExpressionFormat expressionFormat : dataset.getExpressionFormats()) {

			if (!expressionFormat.getType().equals("dcm")) {
				continue;
			}

			List<DatasetFile> datasetFiles = expressionFormat.getDatasetFiles();
			if (datasetFiles == null || datasetFiles.isEmpty()) {
				throw new ShanoirException("Send Dicoms to Pacs: DatasetFiles null or empty.");
			}

			// Iterate over all dataset files (DICOM files), in case all files might be
			// distributed over multiple parent directories and therefore require multiple
			// calls to sendDicomFilesToPacs (advantage: do not create to big multiparts)
			File lastParentFile = null;
			for (DatasetFile datasetFile : datasetFiles) {

				String path = datasetFile.getPath();
				if (path == null || path.isEmpty()) {
					throw new ShanoirException("Send Dicoms to Pacs: DatasetFile with empty path found in db.");
				}

				File dicomFile = new File(path);
				if (!dicomFile.exists()) {
					throw new ShanoirException("Send Dicoms to Pacs: dicomFile not existing on disk.");
				}

				File parentFile = dicomFile.getParentFile();
				if (lastParentFile == null || !parentFile.equals(lastParentFile)) {
					if (dicomWeb) {
						dicomWebService.sendDicomFilesToPacs(parentFile);
					} else {
						dimseService.sendDicomFilesToPacs(parentFile);
					}
					lastParentFile = parentFile;
				}
			}
		}
	}
}
```

shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/DicomPersisterService.java
```diff
@@ -71,17 +71,30 @@ public void persistAllForSerie(Serie serie) throws Exception {
...
+							// distributed over multiple parent directories and therefore require multiple
+							// calls to sendDicomFilesToPacs (advantage: do not create to big multiparts)
+							File lastParentFile = null;
+							for (DatasetFile datasetFile : datasetFiles) {
+								String path = datasetFile.getPath();
+								if (path != null && !path.isEmpty()) {
+									File dicomFile = new File(path);
+									if (dicomFile.exists()) {
+										File parentFile = dicomFile.getParentFile();
+										if (lastParentFile == null || !parentFile.equals(lastParentFile)) {
```
youennmerel, line 84: No need to check for null here, i think
```java
if (lastParentFile == null || !parentFile.equals(lastParentFile)) {
...
```

michaelkain:
Hi @youennmerel, thank you. All done.

youennmerel (approved):

PR #2356, opened 2024-08-09
shanoir-issue#2355: do not re-route users to /home

pierrehenri-dauvergne:
How to test:
As a User, import CT data
On the clinical context, create a new center

Before: shanoir would redirect the user to /home with a message saying you don't have the rights to do this

After: no more redirection. The message still displays but the user is kept on the clinical context page

It is also the case for many other routes (coil creation, equipment edit, etc.)


shanoir-ng-front/src/app/shared/roles/auth-admin-or-expert-guard.ts
```diff
@@ -32,10 +32,9 @@ export class AuthAdminOrExpertGuard  {
         if (this.keycloakService.isUserAdminOrExpert()) {
             return true;
         } else {
-            this.router.navigate(['/home']);
             this.consoleService.log('warn', 'Sorry, you have no right to visit to this page.', ['route : ' + this.router.url]);
```
michaelkain, line 35: Hi @pierrehenri-dauvergne, thank you! Please modify: 'no right to visit this page.' THX!

michaelkain (approved):
Locally tested: WORKS. Thx for the PR!

PR #2362, opened 2024-08-14
ShUp: optimize download from pacs + upload to server

michaelkain:


jcomedouteau:
Hi @michaelkain 

Could you describe a bit what changed and what should I test ?
Thanks !
Jean-Côme

PR #2363, opened 2024-08-19
Ohif viewer fixes

michaelkain:
Hi @pierrehenri-dauvergne,
there were quite some bugs in the backend.
Good news: the viewer calls /studies, but only with the studyInstanceUID wanted.
Now it should work. The problem that remains on this PR is the order of the series,
but I think you solved that already with a param. On the other hand it is very easy
to fix on the json level in the backend and order by seriesNumber the json itself.
With kind regards and happy testing, Michael


michaelkain:
Hi @pierrehenri-dauvergne, I hope you are doing fine.
Could you please on testing this PR between prod (old) and qualif (new), comparison test, use e.g. one serie of VHD study on shanoir-ofsep and compare the list of series + the loading times on each environment? It would be great to test with 5-10 examinations, randomly chosen, and 5-10 exams (very much acquistions, very few, very big images) etc. to have a complete comparison test please. Thank you and with kind regards, Michael

pierrehenri-dauvergne (approved):

PR #2367, opened 2024-08-20
shanoir-issue#2366: special char in ohif

pierrehenri-dauvergne:
How to test:
Search for a serie with a special character in its name (write ' ° ' in the solr search bar)
Open it in OHIF viewer
The name should be displayed correctly


pierrehenri-dauvergne:
Tested in qual and it's ok

michaelkain:
Hi @pierrehenri-dauvergne,
thank you for this great fix.
Not sure, if it is possible, but maybe it would even be better to query dcm4che5 already with the "good" encoding?
maybe this might work as well? do you think you might test it?

// Setting the Accept-Charset header to request UTF-8 encoding
            **httpGet.setHeader("Accept-Charset", "UTF-8");**
            try (CloseableHttpResponse response = httpClient.execute(httpGet)) {
                HttpEntity entity = response.getEntity();
                if (entity != null) {
                    // Explicitly specifying UTF-8 encoding when converting the entity to string
                    return **EntityUtils.toString(entity, "UTF-8");**
                } else {
                    LOG.error("DICOMWeb: findSeriesOfStudy: empty response entity for studyInstanceUID: " + studyInstanceUID);
                }
            }

michaelkain (approved):

PR #2377, opened 2024-08-23
#2364-github-packages

### LINKED ISSUE #2364: docker-compose: switch to GitHub packages
Dear @jcomedouteau, I hope you are doing fine.
Today we build locally on using the global Dockerfile and copy the locally compiled fat-jars into each Docker image.
This is good for local development but unnecessary for anybody else.
It would be great to reference the Docker images, auto-created now on GitHub packages
and propose an easier way of installation for somebody more interested in deployment than on development.
An idea as discussion base:
1) docker-compose.yml (all points to GitHub, by default)
2) docker-compose-dev.yml (points to local fat-jars build by local Maven)
3) docker-compose-dev-front.yml (for today's front dev extension)
I think it might be interesting to have a chat with Anthony about it as well please.

jcomedouteau:
- Update docker-compose file to use generated github packages.

How to test:
- Restart shanoir from scratch and check that every container works.
- 
Closes #2364


docker-compose-dev-front.yml
```diff
@@ -0,0 +1,37 @@
...
+  nginx:
+    environment: 
+      - SHANOIR_DEV=TRUE
+    depends_on:
+      - "front-dev"
+
+  front-dev:
+    container_name: front-dev
+    build:
+      context: docker-compose-dev
```
a-ba, line 25: *context* here refers the the source directory for the docker build (i.e. the `docker-compose-dev/` directory in the source tree). You should keep `docker-compose` here, otherwise it will break the build.

This comment also applies to all images defined in docker-compose-dev.yml.

a-ba (changes requested):

jcomedouteau:
@a-ba Done, I also updated the version.

a-ba (approved):

a-ba:
fine for me

PR #2383, opened 2024-08-28
shanoir-issue#2368-2: checkboxes + open multiple exams in ohif viewer

pierrehenri-dauvergne:
How to test:

Open a study with multiple exams
Check the exams you want to open:
image

Click the "Open in  viewer" button
In the left panel, you should see multiple studies and in the URL you should see the selected exam ids
image


michaelkain:
Hi @pierrehenri-dauvergne, thank you for your PR!
Do you think you could check in the code all entry points to the viewer and harmonize all texts?
I think we should separate the concerns and display in Shanoir everywhere "View DICOM" for the
viewer and remove the OHIF everywhere as most of the users will not know what it means. And in
some years we might (maybe) switch to another view (or not), but it would be great to display everywhere
the same text, on all links that lead to the viewer. With kind regards, Michael

shanoir-ng-front/src/app/studies/study/study-tree.component.html
```diff
@@ -13,6 +13,7 @@
 -->
 <div class="commands">
     <button class="right-icon dl-button" type="button" (click)="downloadSelected()" [disabled]="!(selectedDatasetNodes?.length > 0)">Download selection<i class="fas fa-download"></i></button>
+    <button class="right-icon dl-button" type="button" (click)="openInViewer()" [disabled]="!(selectedExaminationNodes?.length > 0)" [title]="'Select examinations to open in viewer'">Open in viewer<i class="fas fa-eye"></i></button>
```
michaelkain, line 16: I propose, if okay: 'Select examinations to compare DICOM files in viewer' and 'View DICOM' and see my other comment, use it everywhere please.

michaelkain:
and it would be great to add a limit please, as the viewer is pretty demanding on the backend, 10 exams might be a good start

michaelkain (approved):
approved by code review, I test locally now

michaelkain:
my local test work and it makes a big difference already
excellent to be now able to view only one serie in the viewer within a lot of series

PR #2384, opened 2024-08-28
export table as csv

julien-louis:
add a button on every table to export content as csv


michaelkain:
Hi @julien-louis, thank you for this important and very practical PR.
As mentioned already, I will be important to limit the number please.

julien-louis:
from @pierrehenri-dauvergne : 

export de tableau m'a donné un fichier avec ce nom : tableExport_1 725 374 576 884.csv ou tableExport_1 725 373 982 056.csv

pierrehenri-dauvergne (approved):

PR #2390, opened 2024-09-02
deprecated keycloak option: --proxy

a-ba:
https://www.keycloak.org/docs/25.0.4/upgrading/#deprecated-proxy-option


michaelkain (approved):
tested locally, all works

PR #2391, opened 2024-09-03
shanoir-issue#2389: allow 'view dicom' with exams or acquisition from one exam

pierrehenri-dauvergne:
How to test:
Open a study
Select exams or acquisition and click the "View DICOM" button
This button should be enabled only if: 
- only exams are selected
- acquisitions from a single exam are selected

image
image
image

This is happening because when the viewer is opened, if there are multiple series (acquisition) and studies (exam) selected, the viewer can't know which series belong to which study. In the same way, if there are only series selected, the viewer won't be able to know from which study they are from.


michaelkain (approved):

michaelkain:
wrong display of viewer as well on develop, as it works, I approve

Issue #2392, opened 2024-09-03
Vertical slider to the new tree view + keep buttons visible


michaelkain:
Hi @julien-louis,
I tested with Ican and to reach the bottom of the new tree view took very long.
Do you think you might please add a slider?
Thank you and with kind regards, Michael

pierrehenri-dauvergne:
Sometimes studies have a lot of subject and you have to scroll a long time in the tree-view to find where you are. 
Could we make the tree-view auto-scroll to wherever you are in the tree?

PR #2418, opened 2024-09-30
Tree updates

julien-louis:
see https://github.com/fli-iam/shanoir-ng/issues/2392

List of demands : 
* slider because ican subjects too long [OK]
* buttons visible [OK]
* open by default [OK]
* text filter [OK]
* unselect all ? [OK]
* refresh tree when deleting from a list (i.e. acquisition) [OK]

### DIFF
--- a/shanoir-ng-front/src/app/app.component.css
+++ b/shanoir-ng-front/src/app/app.component.css
@@ -8,17 +8,16 @@
 side-menu, .side-menu { overflow-x: hidden; scrollbar-width: none; overflow: -moz-scrollbars-none; -ms-overflow-style: none; }
 .side-menu { width: 300px; position: fixed; top: 0; z-index: 0; }
 
-study-tree { scrollbar-width: none; overflow: -moz-scrollbars-none; -ms-overflow-style: none; z-index: 1; 
-    max-height: calc(100% - 10px); display: inline-block; overflow-y: scroll; position: relative; min-width: fit-content; margin-top: 10px; }
+study-tree { z-index: 1; 
+    height: 100%; display: inline-block; overflow: hidden; position: relative; min-width: fit-content; margin-top: 10px; }
 
 .tree-frame { border-right: 1px solid var(--grey); margin-right: 20px; }
 
 .tree, .content { vertical-align: top; }
-    /*  */
 .close-button, 
 .open-button { width: 14px; height: 27px; position: fixed; top: 0; left: 0; cursor: pointer; color: var(--very-light-grey); z-index: 2; }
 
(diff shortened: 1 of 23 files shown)


julien-louis:
Ready for review.

pierrehenri-dauvergne:
@julien-louis :
- We can't open acquisitions in OHIF viewer anymore, the button is enabled but a click on it doesn't do anything (+ maybe change the subtitle to "View DICOM"):
image

- I feel like the research option is not visible enough, could we add a filter option at the top that would filter out every subject/exam/acquisition/dataset that do not have the searched input, something like this?
image

- If I check a couple of checkbox in a study, then open solr, then open that study from solr, I get a ExpressionChangedAfterItHasBeenCheckedError in console
image

- Whenever I open one study, go back to study list, then open another study (study can be opened from examination list or anywhere I think) I have an error: TypeError: this.globalClickSubscription is undefined

- I made some measure in OHIF viewer, that create new dataset in the acquisition. The name of the acquisition node is concatenated with all datasets (here, "abcdef", "abc", "a", "aa", etc. are all datasets of modality type "measurement"):
image

pierrehenri-dauvergne:
- When selecting 3 acquisition from the same examination and clicking "View DICOM", the url looks like this:

https://viewer/viewer?StudyInstanceUIDs=1.4.9.12.34.1.8527.56,1.4.9.12.34.1.8527.56,1.4.9.12.34.1.8527.56&SeriesInstanceUIDs=1.4.9.12.34.1.8527.538,1.4.9.12.34.1.8527.517,1.4.9.12.34.1.8527.516
There should be only one StudyInstanceUID instead of 3 times "1.4.9.12.34.1.8527.56" (an error is shown in the viewer but shouldn't, it's still working)


- We shouldn't be able to click "View DICOM" when multiple acquisitions are selected across multiple examination. This is because the viewer doesn't know which SeriesInstanceUIDs belong to which StudyInstanceUIDs.
OK : multiple examinations selected
OK : multiple acquisitions from the same examination
KO : multiple acquisitions from multiple examinations


- I don't know if it is on purpose but the subject search includes the "HEALTHY VOLUNTEER":
image


- This error is still thrown:
If I check a couple of checkbox in a study, then open solr, then open that study from solr, I get a ExpressionChangedAfterItHasBeenCheckedError in console
image

julien-louis:
I propose that if you select multiple acq from multiple exams it behaves just like you selected those exams.

julien-louis:
@pierrehenri-dauvergne for me it's ok

pierrehenri-dauvergne (approved):

PR #2424, opened 2024-10-02
shanoir-issue#2417: convert dates as string to Date

pierrehenri-dauvergne:
How to test:

Open study's list page (or users list as admin)
"Start date" column may look like this:
image

Click on the "Start date" header, the column is now order by its start date, either in ascending order or descending order:
image

Same goes for users list:
image


julien-louis:
@pierrehenri-dauvergne did you spot problems elsewhere than in users list ?

julien-louis:
I think a better solution would be removing all those "dateRenderer" function and let the table component handle it all. I open the pr https://github.com/fli-iam/shanoir-ng/pull/2428 and close this one. It can be re-opened if I'm wrong.

PR #2434, opened 2024-10-09
shanoir-issue#2307: solr search detects keywords and switch to expertMode

pierrehenri-dauvergne:
How to test:
Open solr
In the search field at the top, write one of the keywords of expertMode: centerName datasetCreationDate studyName subjectName subjectType acquisitionEquipmentName datasetId datasetName datasetNature datasetType processed examinationComment examinationDate importDate tags magneticFieldStrength pixelBandwidth sliceThickness studyId

As soon as you write any of those keyword, the research automatically switch to expertMode:
image

This PR also changes the syntax error message to display something more accurate:
image


julien-louis (changes requested):
Also I think that this features and the toggle switch doesn't work well together. Maybe we should remove the toggle switch. But also I think the solr syntax detection system is too weak. I think there may be requests that doesn't include any field name. I have looked at possible stronger validations but I haven't find a good solution yet. The best would be to rely on the Solr server/api itself (without running a request at every text change). Maybe we should discuss it with Michael.

julien-louis:
@pierrehenri-dauvergne So for now let's keep the button and automatically switch to expert mode but not go back automatically to normal mode.

PR #2441, opened 2024-10-15
performances optimizations in clinical context

julien-louis:
* get subjects by study Id
    * use PostAuthorize instead of PostFilter -> 1 db request instead of n (n subjects)
    * remove useless n request to retrieve subject study tags (already present in the subject study object)
    * move request that get study tags outside the loop : n to 1 calls
* get studies with their centers
    * remove the slow List<Study> findAllWithCenters();
* removed some @EntityGraph as I think it slowed down the request
    * I need to discuss it with @michaelkain to see if I didn't break anything
* removed loops with n db requests to fill children
* had to used "manual" HQL queries instead


julien-louis:
ok I have to keep findNamesAndCenters

LaurentPV:
The endpoint findAllSubjectsOfStudyId performance is greatly improved : from 2 minutes 45seconds to 6seconds

The endpoint findStudiesNamesAndCenters returns a 500 error for now.

julien-louis:
@LaurentPV  Ok it's fixed. One warning : if the uploader would feature setting tags to the subject at import this fix would break it. But I analyzed that the uploader hasn't got this feature.

LaurentPV:
tested today, everything is ok and performances are improved.

LaurentPV (approved):
nothing to say about your commits. everything seems ok

Issue #2435, opened 2024-10-10
ShanoirUploader - Requesting the PACS with only Patient's ID does not function


LaurentPV:
People doing the imports often use Hospital's IPP to identify the patient without using personal data.

The objective would be to allow requesting PACS on Patient level with the Patient ID (IPP or NIP) used by the center.

michaelkain:
fixed by Laurent, I close

PR #2444, opened 2024-10-17
SHANOIRUploader - Allowing Patient ID only C-FIND request

LaurentPV:
The problem of issue #2435 was the condition of length over the last name.
We implemented the wildcard "*" by default if only a last name was provided but then we sent an error if the last name was less than 3 characters long if it contained the wildcard "*".

Without any null or empty condition, the error was always poping up if the Name field wasn't filled.

The request using only one of the other fields was then in error with always the same name length condition error pop up.

In cas of a short last name ("Li" for example), the user couldn't manage to search for the patient.

I also simplified the user interface by replacing the "Help" button with a tooltip over Patient's name textField.


michaelkain:
@LaurentPV will make some more tests and come back to me

LaurentPV:
Tested and Ok for me, both on patient and study root query levels.

michaelkain (approved):
Hi, I approve by code review and a local test.

Issue #2437, opened 2024-10-11
Are acquisitions names in tree bad ?


julien-louis:
Names are actually set by the backend :

https://github.com/fli-iam/shanoir-ng/blob/b3d798cbce60741fd49951015c3568a97e194f64/shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/dto/mapper/ExaminationDatasetAcquisitionDecorator.java#L82

When saving segmentation datasets from the OHIF viewer, names can get a bit long : 
 
> made some measure in OHIF viewer, that create new dataset in the acquisition. The name of the acquisition node is concatenated with all datasets (here, "abcdef", "abc", "a", "aa", etc. are all datasets of modality type "measurement"):
![image](https://github.com/user-attachments/assets/8c8eadce-f355-4c3b-96c5-5d1d566843dd)

@pierrehenri-dauvergne was asking if this is really what we want

michaelkain:
Hi, please make the names short again :-), thank you!

PR #2451, opened 2024-10-21
shorten acquisition names in tree

julien-louis:
fix https://github.com/fli-iam/shanoir-ng/issues/2437


pierrehenri-dauvergne:
The names are too short now:
image

Here, it should display only "MPRAGE_iso_gado"

pierrehenri-dauvergne (changes requested):
see comment

julien-louis:
note : when multiple datasets, use the one that is not a segmentation (as modality)

PR #2459, opened 2024-10-25
shanoir-issue#2447: improve copy

pierrehenri-dauvergne:
How to test:
**1. Deletion of subject**
Can't reproduce a problem for now
Everytime I delete a subject, everything underneath (exam, acq, ds) is deleted properly

**2. Change error when copying a dataset that already exists in a study:**
Create a new study
Copy datasets into this new study
Try to copy once again the same dataset in this study
Before: the job was in error and shows "[CopyDatasets] Error during the copy of dataset."
Now: The job doesn't show error but this:
image

**3. Show that an entity is a copy or that it has been copied**
Create a new study
Copy datasets into this new study
Open the copied or source examination/acquisition/datasets details to see this new field (left image is source, right is copy):
image


michaelkain:
waiting for Julien

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
```diff
@@ -171,6 +171,8 @@ public abstract class Dataset extends AbstractEntity {
 	@Transient
 	public String SOPInstanceUID;
 
+	private String copyMessage;
```
julien-louis, line 174: What kind of message can be set here ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -186,7 +186,18 @@ public void deleteByIdIn(List<Long> ids) throws ShanoirException, SolrServerExce
 
 	@Override
 	public Dataset findById(final Long id) {
-		return repository.findById(id).orElse(null);
+		Dataset ds = repository.findById(id).orElse(null);
```
julien-louis, line 189: Is this the only case for the message ?

shanoir-ng-front/src/app/dataset-acquisitions/dataset-acquisition/dataset-acquisition.component.ts
```diff
@@ -125,4 +128,18 @@ export class DatasetAcquisitionComponent extends EntityComponent<DatasetAcquisit
     downloadAll() {
         this.downloadService.downloadAllByAcquisitionId(this.datasetAcquisition?.id, this.downloadState);
     }
+
+    getCopiedEntity(copyMsg: string) {
```
julien-louis, line 132: If the message field is intended to retrieve ids by parsing it here, it's not the right way to do it. The copied entity should have a copiedFrom attribute/column that references the original entity. Note that if this reference should remain when the origin entity is deleted, then it cannot have a foreign key constraint. Then the message would be reconstructed by the front using this id.

julien-louis (changes requested):

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
pierrehenri-dauvergne: Those are the possible values of copyMessage:

"This dataset is the copy of :"
"This dataset has been copied :"
"This acquisition is the copy of :"
"This acquisition has been copied :"
"This examination is the copy of :"
"This examination has been copied :"

shanoir-ng-front/src/app/dataset-acquisitions/dataset-acquisition/dataset-acquisition.component.ts
pierrehenri-dauvergne: Every (examination, acquisition, datasets) entity has a "source" value. If it's not empty, it means it's a copy and the id stored in 'source' is the id of the original entity. Otherwise, it's null.
We can't delete an original entity if a copy of it exists

pierrehenri-dauvergne:
Tested on qualif neurinfo and works as intended

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
```diff
@@ -165,7 +149,14 @@ public abstract class Dataset extends AbstractEntity {
...
 
-	private Long sourceId;
+	@ManyToOne(fetch = FetchType.LAZY)
+	@JoinColumn(name = "source_id")
+	private Dataset source;
+
+	@OneToMany(fetch = FetchType.LAZY, mappedBy = "source", cascade = CascadeType.ALL)
+	private List<Dataset> copies;
+
+    private Integer origin;
```
julien-louis, line 159: Several problems with origin imo : 
* I'm not sure it's needed here as a Dataset property. This information can be deduced by the state of ```Dataset source``` and ```List<Dataset> copies```. If it has copies it's a sources, if it has a source it's a copy. If the getter ```getOrigin()``` is usefull, it can infer its return using source and copies.
* An entity could be both a copy and a source. So you could either use a 3rd state in the enum, or have two booleans instead : isCopy and isSource.
* If origin is removed from Dataset and EntityOrigin is not to be persisted in the database, it does not need to have an id, a getType method or a constructor.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/repository/ExaminationRepository.java
```diff
@@ -84,7 +85,12 @@ public interface ExaminationRepository extends PagingAndSortingRepository<Examin
 	 */
 	@Query("select e.id from Examination e where e.study.id = :studyId")
 	List<Long> findIdsByStudyId(Long studyId);
-	
+
+
+	@Query(value = "SELECT copies_id FROM examination_copies WHERE examination_id = :examinationId", nativeQuery = true)
+	List<Long> findCopyIdsByExaminationId(@Param("examinationId") Long examinationId);
```
julien-louis, line 91: is this used somewhere ? sorry I don't see it

shanoir-ng-front/src/app/dataset-acquisitions/dataset-acquisition/dataset-acquisition.component.html
```diff
@@ -149,6 +149,20 @@ <h2 class="header command-zone">Create dataset acquisition</h2>
 							</ng-template>
 						</span>
 					</li>
+                    <li *ngIf="mode != 'create' && datasetAcquisition.origin == 'SOURCE'">
```
julien-louis, line 152: here if your acquisition is both a source and a copy you will not have every link

shanoir-ng-front/src/app/dataset-acquisitions/dataset-acquisition/dataset-acquisition.component.html
```diff
@@ -149,6 +149,20 @@ <h2 class="header command-zone">Create dataset acquisition</h2>
...
+                        <label>This acquisition has been copied :</label>
+                        <span class="right-col">
+                            <div *ngFor="let entityId of copyEntityIds">
```
julien-louis, line 155: why not ```let entityId of datasetAcquisition.copies``` directly ? Then coupled with ```*ngIf="datasetAcquisition.copies?.length > 0"``` instead of ```datasetAcquisition.origin == 'SOURCE'"``` you may have no use of the enum anymore (if it's not used elsewhere)

shanoir-ng-front/src/app/dataset-acquisitions/dataset-acquisition/dataset-acquisition.component.ts
```diff
@@ -125,4 +126,15 @@ export class DatasetAcquisitionComponent extends EntityComponent<DatasetAcquisit
     downloadAll() {
         this.downloadService.downloadAllByAcquisitionId(this.datasetAcquisition?.id, this.downloadState);
     }
+
+    getCopiedEntity(dsAcq: DatasetAcquisition) {
```
julien-louis, line 130: If I'm right with my previous comment, this will not be usefull anymore.

julien-louis (changes requested):

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/model/Dataset.java
pierrehenri-dauvergne: - Indeed, EntityOrigin is not needed since we can check the content of copies and source. I'll remove this enum and work with copies/source only.

- A copy can't be copied.

julien-louis:
(note : code is ok, just need to test)

julien-louis (approved):

PR #2461, opened 2024-10-28
CI: use docker-compose-dev.yml for building the images

a-ba:
was broken by #2377


michaelkain (approved):
Perfect, thx Anthony!

PR #2464, opened 2024-10-29
CI: fix docker bake invocation

a-ba:
the fix in #2461 was incomplete

since our config is no longer in the default 'docker-compose.yml' we need to set our bake files explicitly
https://docs.docker.com/build/bake/reference/


michaelkain (approved):

michaelkain:
Hi Anthony, thx for your PR! Is it normal, that the referenced file docker-bake.json is not part of the PR?

PR #2480, opened 2024-10-31
Shanoir issue#2463 Delete subject does not appear in study's history 

pierrehenri-dauvergne:
**How to test:**
Create a new study
Copy one dataset into it
Go to new study details > subject tab > edit > remove subject > save
Open the history's tab:
image

**Other test:**
Create a new study
Create a new subject into this new study
Remove subject from study > save
It should appear on study's history
(I haven't tested this one yet)

~~**Issue https://github.com/fli-iam/shanoir-ng/issues/2482 is blocking**~~


shanoir-ng-studies/src/main/java/org/shanoir/ng/study/service/StudyServiceImpl.java
```diff
@@ -325,10 +330,21 @@ public Study update(Study study) throws ShanoirException {
 
 			for (SubjectStudy subjectStudyDb : studyDb.getSubjectStudyList()) {
 				if(!updatedIds.contains(subjectStudyDb.getId())) {
-					removed.add(subjectStudyDb.getSubject());
+					Subject sub = subjectStudyDb.getSubject();
```
julien-louis, line 333: ok so it works only when you delete a whole study

shouldn't this be in subject service as well ?

julien-louis (commented):
Ok it works but only if you remove the subject by updating the study.

Question : should it work when deleting a subject from the subject details or the subject list as well ?

julien-louis (approved):
I create an issue though : https://github.com/fli-iam/shanoir-ng/issues/2570

(I think this has nothing to do with the pr.)

PR #2487, opened 2024-11-06
#2364 - github packages - 2 bootstrap.sh

jcomedouteau:
- Update bootstrap.sh with adapted docker-compose file


DuckflipXYZ (approved):
Some warning while building, but the script ends properly

PR #2490, opened 2024-11-07
Frame of reference UID fix

michaelkain:
This fix is the follow up of an issue raised by GE.
With the fix respect the DICOM structure concerning the frameOfReferenceUID, as given in the original DICOM.


michaelkain:
Hi, I tested it successfully with my local ShUp. Might be worth testing on the server too. Cheers, Michael

shanoir-ng-anonymization/src/main/java/org/shanoir/anonymization/anonymization/AnonymizationServiceImpl.java
```diff
@@ -350,6 +354,25 @@ private void anonymizeSOPInstanceUID(int tagInt, Attributes attributes, String m
...
+			} catch (Exception e) {
+				LOG.error(e.getMessage());
+			}
+			value = newUID;
+			frameOfReferenceUIDs.put(attributes.getString(tagInt), value);
+		}
+		anonymizeTagAccordingToVR(attributes, tagInt, value);
+	}
+
 	private void anonymizeSeriesInstanceUID(int tagInt, Attributes attributes, Map<String, String> seriesInstanceUIDs) {
```
jcomedouteau, line 376: If I'm not wrong, the three methods could be factorized into a single one (and we just give a different Map<string, string> for each case), but I don't know if it's really relevant in terms of "code clarity" (?)

shanoir-ng-anonymization/src/main/java/org/shanoir/anonymization/anonymization/AnonymizationServiceImpl.java
```diff
@@ -217,6 +219,8 @@ public void performAnonymization(final File dicomFile, Map<String, String> anony
 						anonymizeSOPInstanceUID(tagInt, datasetAttributes, mediaStorageSOPInstanceUIDGenerated);
 					} else if (tagInt == Tag.SeriesInstanceUID) {
 						anonymizeSeriesInstanceUID(tagInt, datasetAttributes, seriesInstanceUIDs);
+					} else if (tagInt == Tag.FrameOfReferenceUID) {
```
LaurentPV, line 222: Maybe change "else if" to a switch case.

shanoir-ng-anonymization/src/main/java/org/shanoir/anonymization/anonymization/AnonymizationServiceImpl.java
```diff
@@ -523,4 +484,25 @@ else if (vr.equals(VR.AE) || vr.equals(VR.AS) || vr.equals(VR.CS) || vr.equals(V
 		// VR.OD = Other Double String
 	}
 
+	private void anonymizeUID(int tagInt, Attributes attributes, Map<String, String> UIDs) {
```
michaelkain, line 487: @LaurentPV the problem now is with the method factorization, that it produces too much logs
the info that the StudyInstanceUID has been generated, one by exam, is important, but one log
per serie is just too much, probably the reason, why I kept the 3 methods in the first place

LaurentPV: I added a condition to log only when studyInstanceUID is concerned. If you think that's really better with two different methods I'll rollback.

PR #2493, opened 2024-11-08
shanoir-issue#2482: call subject delete in preclinical and datasets at the same time

pierrehenri-dauvergne:
How to test:
Create new study
Create new subject and add it to the new study
Open study details > Edit > Subject tab > Remove subject > Save study
Subject should be removed from the study without error
If the subject do not belong to any other study, it should be deleted along with all of its data (examinations, acquisitions, datasets)

Also try with a **subject that has examinations** and a **preclinical subject**


shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/amqp/RabbitMQDatasetsService.java
```diff
@@ -313,7 +313,7 @@ public void createDatasetAcquisition(final String studyStr) {
 
 	/**
          * Receives a shanoirEvent as a json object, concerning a subject deletion
-         * @param eventAsString the task as a json string.
+         * @param subjectIdAsString a string of the subject's id
          */
 	@RabbitListener(bindings = @QueueBinding(
```
jcomedouteau, line 318: Please remove this and update it to be simpler (and not event depending) :
	@RabbitListener(queues = RabbitMQConfiguration.DELETE_SUBJECT_QUEUE, containerFactory = "singleConsumerFactory")

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/controler/SubjectApiController.java
```diff
@@ -63,13 +66,16 @@ public class SubjectApiController implements SubjectApi {
...
 
+	@Autowired
+	private RabbitTemplate rabbitTemplate;
+
 	@Override
 	public ResponseEntity<Void> deleteSubject(
 			@Parameter(description = "id of the subject", required = true) @PathVariable("subjectId") Long subjectId) {
 		try {
 			// Delete all associated bids folders
 			subjectService.deleteById(subjectId);
```
jcomedouteau, line 77: Alors pour autant on l'utilise plus pour communiquer entre microservices cet event, mais est-ce qu'il faut pas le conserver quand même ? (sauf s'il est publié ailleurs ?)

pierrehenri-dauvergne: En effet, les events sont toujours utilisés pour l'historique des études, c'est rajouté

julien-louis:
ok I will approve this but there is still a (moderate) problem : the subject is deleted from studies but from datasets

julien-louis (approved):

Issue #2387, opened 2024-08-30
Migration Angular 17 -> 18


michaelkain:
Hi @julien-louis,
I hope you are doing fine.
I think it would be interesting to migrate to the latest version of Angular.
With kind regards, Michael

PR #2495, opened 2024-11-08
Angular 17 to 19

julien-louis:
devs may need to update node  and/or npm 

https://github.com/fli-iam/shanoir-ng/issues/2387


pierrehenri-dauvergne:
I have this error when deploying front-dev:
An unhandled exception occurred: Cannot find module @rollup/rollup-linux-x64-musl. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.

I tried most of the proposed solution in the thread but none worked. It's opened for 2 years and it's still an issue. I'm not quite sure what to do.

julien-louis:
Pierre Henry, 

Je crois que tu dois ajouter la dependence version windows à la suite de celle-ci. 

Julien

julien-louis:
@pierrehenri-dauvergne  I upgraded to angular 19 and the annoying dependencies are no longer needed

michaelkain:
Hi! to merge before test session?

pierrehenri-dauvergne (approved):
To make it work, i had to use "FROM node:lts-alpine as front-dev" in Dockerfile, it uses node v23.3.0

PR #2500, opened 2024-11-14
remove apache poi dependency

julien-louis:
Removed 2 dependencies in import pom

If it builds its ok right @michaelkain ?


michaelkain (approved):

michaelkain:
all good, I tested an import and pseudo still working (as dependency probably related to pseudo-module)

Issue #2455, opened 2024-10-22
New tree GUI: Papaya preview does not update between datasets


michaelkain:
Hi @julien-louis,
there might be an error: when I use the tree view to switch between datasets
and I use Papaya to display the images of one dataset, when switching the dataset
it looks like it is not updated for the new, next dataset.
Thank you and with kind regards, Michael

PR #2504, opened 2024-11-15
fix papaya not refreshing

julien-louis:
fix https://github.com/fli-iam/shanoir-ng/issues/2455


pierrehenri-dauvergne (approved):
Tested locally and OK for me

Issue #2249, opened 2024-06-07
Better + icon in clinical context


julien-louis:
find a better way to display the possiblility of creating a new subject / examination in clinical context page

PR #2505, opened 2024-11-18
Easier add subject/exam buttons in clinical context

julien-louis:
image

see https://github.com/fli-iam/shanoir-ng/issues/2249


pierrehenri-dauvergne:
All fine for me but could we also do the same in subject create form:
image

Center equipment:
image

Manufacturer:
image

Coils:
image

Study cards:
image

And in the study's details:
image

image


Also, maybe it should be another issue, but should we stylise every button the same? Some are dark text and filled with white, others are white text and filled with color:
image
For example in import:
image
image
image

What do you think ?

pierrehenri-dauvergne (approved):

Issue #2511, opened 2024-11-22
Can't delete a study


pierrehenri-dauvergne:
Create a new study
Open it and click the "delete" button
An error is thrown: DataIntegrityViolationException: could not execute statement [Cannot delete or update a parent row: a foreign key constraint fails (`studies`.`study_center`, CONSTRAINT `FK2hmmh3c0w1tk8npi76hpvf09i` FOREIGN KEY (`study_id`) REFERENCES `study` (`id`))] [delete from study where id=?]; SQL [delete from study where id=?]; constraint [null]

julien-louis:
will be fixed by https://github.com/fli-iam/shanoir-ng/pull/2525

Issue #2520, opened 2024-11-28
Remove center from a study


pierrehenri-dauvergne:
Open a study with multiple centers
Edit study
Remove a center
Update study

In qualif and prod: green message is displayed saying the study was correctly updated but the removed center is still here

![image](https://github.com/user-attachments/assets/ea9feffe-f29a-40cf-a730-a9de8111b414)

![image](https://github.com/user-attachments/assets/fa3bc5cc-ddf2-47a4-962d-132cb8e33258)

![image](https://github.com/user-attachments/assets/830d13c0-05d1-4d06-aeb0-c10b52e4f19c)

pierrehenri-dauvergne:
This might be fixed with https://github.com/fli-iam/shanoir-ng/pull/2522

julien-louis:
will be fixed by https://github.com/fli-iam/shanoir-ng/pull/2525

PR #2525, opened 2024-11-29
fix with LAZY and Transactional on controlers

julien-louis:


pierrehenri-dauvergne:
This PR is a fix for https://github.com/fli-iam/shanoir-ng/issues/2520 right ?

pierrehenri-dauvergne:
Hello @julien-louis 
Tested on qualif:
Create a study > delete study : OK
Create a study > import in a new subject > delete study : Warning, another entity is linked
Create a study > import in a new subject > delete exam of that import > delete study : Warning, another entity is linked
Create a study > import in new subject > delete subject : issue ongoing in another PR
Create a study > Create a new subject and add it to study > delete study : OK

When we delete a study, it should remove the subject from that study and also delete all data that is in that study, right ?

julien-louis:
@pierrehenri-dauvergne  Yes it fixes https://github.com/fli-iam/shanoir-ng/issues/2511 and also https://github.com/fli-iam/shanoir-ng/issues/2520

Is there any regression with that code ? If not I think we could merge it.

pierrehenri-dauvergne (approved):
I didn't see any regression while testing. Merging this so we can move forward on other issues

Issue #2491, opened 2024-11-07
Import mail: series order of exam is wrong


michaelkain:
Hi Guéwen,
I hope you are doing great.
This is a minor issue, that would be great to solve and start with:
When I receive an import mail, sent by Shanoir server, after an import,
in the mail the series (list of dataset acquisitions) is not ordered by sorting_index (seriesNumber).
This his confusing for our users, please fix this. Thank you and with kind regards, Michael

DuckflipXYZ:
Merged, but needs to be test in qualif (asap, currently impossible due to connexion issue)

PR #2528, opened 2024-12-02
#2491 : Sorting acquisitions in mail per sorting index when imported from PACS

DuckflipXYZ:
All in title


michaelkain (approved):
Hi Guewen, this look very promising! I approve by code review only, we will check during the next test session, if the order of dataset acquisitions is correct now.
By the way this concerns all emails for all imports, not only import from PACS.
With kind regards, Michael

Issue #2415, opened 2024-09-24
Shared datasets on Shanoir: streamline the process to request access to a Shanoir open dataset


cmaumet:
The goal of this feature will be to streamline the process of accessing shared datasets on Shanoir. 

We will review jointly the steps involved to access a shared dataset for a new user and identify any gaps or ways to make these even easier. This will also entail adding the corresponding documentation on the Shanoir website.

--

For example, clicking on "Requesting access" at https://shanoir.irisa.fr/shanoir-ng/welcome currently brings us to an account creation page. It is not entirely clear to me what a user who already has an account should do. (Note: this is only an example but the goal of this feature will be to review the full set of steps involved in getting access to open data for different user profiles).

cmaumet:
Link to relevant documentation (via @alexpron, thanks!): https://project.inria.fr/shanoir/access-request-rework-in-shanoir-a-public-catalog-for-more-fairness/ 

This doc may be updated and publicized as part of this feature.

michaelkain:
Hi @julien-louis, I hope you are doing fine.
Could you please have a meeting with Camille and detail the above specification, to fix, what the expectations of this issue are? and propose a work-plan, what this issue might include. Thank you and with kind regards, Michael

julien-louis:
* public studies list > request an account
    * problem : if you have an account you are asked to create one 
        * ***maybe it should ask if you have an account or not***
            * ***if yes : go to login page then to the /shanoir-ng/access-request/study/xxx page***
            * ***if no : go to account request***
    * "You have been invited to join a study in Shanoir, you have to create an account before accessing the data" : ***not invited in all use cases***
        * *** remove "you have been invited"*** 
    * "Shanoir study manager" : useless ? confusing for a newcomer
        * *** remove it ***
    * "Expertise / Role in the study"
        * 2 different things
        * not relevant ?
        * *** ? ***
    * "Request an account" could be "Request access to study xxx"

julien-louis:
# Proposition

* click access request on welcome page
    * popup "do you have an account"
        * yes : login page
            * redirect to access request form
                * if already pending access request : info msg + redirect to study list
                * if already full access to study : info msg + redirect to study
        * no : account request page

![image](https://github.com/user-attachments/assets/533d09c5-f830-41f5-ab61-563e72ed46e5)

![image](https://github.com/user-attachments/assets/ee6b0538-d28d-402a-a868-0041c6c7c7c8)

[...]

PR #2530, opened 2024-12-04
Access request updgrade

julien-louis:
see https://github.com/fli-iam/shanoir-ng/issues/2415


julien-louis:
Status : wait for Elise approval

julien-louis:
(also this PR includes the angular one, its commits will not be displayed anymore when angular 19 is merged in develop)

PR #2533, opened 2024-12-05
Reformat conditionResult messages 

LaurentPV:
add seriesDescription and seriesNumber to identify dataset instead of SOPInstanceUID


julien-louis:
@LaurentPV great ! But what if the quality control is launched from the UI on already imported data ? Wouldn't the dataset id usefull in this case ?

julien-louis (approved):

Issue #2506, opened 2024-11-18
[Roadmap discussion] Add an API to add dataset tags


jcomedouteau:
Dataset tags are currently only managed by processing handlers and can't be created "externally" (by the interface or an API)

We should add the possibility to add/remove them using an API.

-> Create a DatasetTagApi with a method addDatasetTag(Long datasetId, String tagName) in which we add a tag to a dataset.
(To see how it currently works, see OFSEPSeqIdHandler#setDatasetTags method -> Do not consider properties that are not relevant for this issue.)
-> Add a method removeDatasetTag(Long datasetId, String tagName) in which we remove this specific tag for this dataset.

The tag should be existing in the study (we currently base ourself on study tags, god knows why), and the used should have the right to import on the study / center.

Can also be done by Guewen :)

PR #2541, opened 2024-12-10
#2506 : dataset tag API created (add/remove tags to a dataset only)

DuckflipXYZ:
WIP, not sure if UI is required or not.

You can only refer to of the study of a dataset. Can be tested with swagger, access to Id and import right needed.


shanoir-ng-datasets/src/main/java/org/shanoir/ng/tag/controller/StudyTagApi.java
```diff
@@ -0,0 +1,64 @@
...
+			@ApiResponse(responseCode = "401", description = "unauthorized"),
+			@ApiResponse(responseCode = "403", description = "forbidden"),
+			@ApiResponse(responseCode = "404", description = "dataset does not exists"),
+			@ApiResponse(responseCode = "422", description = "bad parameters"),
+			@ApiResponse(responseCode = "500", description = "unexpected error") })
+	@PutMapping(value = "/addStudyTagsToDataset")
+	@PreAuthorize("hasRole('ADMIN') or (hasRole('EXPERT') and @datasetSecurityService.hasRightOnDataset(#datasetId, 'CAN_IMPORT'))")
+	ResponseEntity<Void> addStudyTagsToDataset(
+			@Parameter(description = "id of the dataset", required = true) @RequestParam(value = "datasetId") Long datasetId,
+			@Parameter(description = "study tag ids", required = true) @RequestParam(value = "studyTagIds") List<Long> studyTagIds)
```
LaurentPV, line 48: Maybe it would be easier to use a List<String> instead of a List<Long> as the study tag ids are not available unless you request the database.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/tag/controller/StudyTagApi.java
```diff
@@ -0,0 +1,64 @@
...
+			@ApiResponse(responseCode = "401", description = "unauthorized"),
+			@ApiResponse(responseCode = "403", description = "forbidden"),
+			@ApiResponse(responseCode = "404", description = "dataset does not exists"),
+			@ApiResponse(responseCode = "422", description = "bad parameters"),
+			@ApiResponse(responseCode = "500", description = "unexpected error") })
+	@PutMapping(value = "/removeStudyTagsFromDataset")
+	@PreAuthorize("hasRole('ADMIN') or (hasRole('EXPERT') and @datasetSecurityService.hasRightOnDataset(#datasetId, 'CAN_IMPORT'))")
+	ResponseEntity<Void> removeStudyTagsFromDataset(
+			@Parameter(description = "id of the dataset", required = true) @RequestParam(value = "datasetId") Long datasetId,
+			@Parameter(description = "study tag ids", required = true) @RequestParam(value = "studyTagIds") List<Long> studyTagIds)
```
LaurentPV, line 62: Same comment as previous one.

LaurentPV (changes requested):
Tested and works fine but would recommend to use actual tags instead of ids and it still needs to be implemented on the study tree so that we could see the tags on the ui.

LaurentPV (approved):
PR tested and approved

PR #2543, opened 2024-12-11
ShanoirUploader: new release

michaelkain:
- fixes (on switch statements) to comply again with Java 17 (necessary for some CHUs)
- cleanTempFolder before downloadFromPacs (in case old temporary images block a new import with the same images)
- fix for Windows (only): use copy + delete instead of nio.move call to avoid error: "another process has access to the current file" (what is definitely a misleading error, as no other process accesses to the temp files of ShanoirUploader and the switch to copy + deleted fixed the problem)


michaelkain:
Hi @LaurentPV, I hope you are doing fine. As soon as you approve, I merge. Thank you and with kind regards, Michael

michaelkain:
Hi @LaurentPV, PERFECT! Excellent work.

PR #2553, opened 2024-12-17
shanoir-issue#2542: test ohif viewer v3.10.0-beta.22

pierrehenri-dauvergne:
How to test:
Open solr > click on 'view dicom' icon (far right column)
The dataset is visible in the viewer without issues

### DIFF
--- a/docker-compose/Dockerfile
+++ b/docker-compose/Dockerfile
@@ -311,7 +311,7 @@ CMD []
 
 ################ nginx #####################################################
 
-FROM ohif/app:v3.9.0-beta.76 AS nginx-viewer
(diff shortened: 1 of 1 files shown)


michaelkain:
Hi @pierrehenri-dauvergne, thank you for your PR.
Have there been some errors in our backend as well? some requests with wrong return codes?
something in the JavaScript console? something in nginx logs? did the backend return 2 DICOM studies instead of one?

pierrehenri-dauvergne:
Tested with the version that ohif is using for their demo (v3.9.3 : https://viewer.ohif.org/)

This is what I have when I try to open a single acquisition from an examination:
image

It opens the three examination of my study (but only the acquisition selected) and an error message is displayed:
image

PR #2559, opened 2024-12-18
load bar for large files

julien-louis:
problem : when dowloading large files and / or with a slow network, nothing seem to happen. Actually the file is fully downloaded silently, then the brwser dl menu is triggered.

* Here a progress bar in the left side menu his displayed if the dl takes more than 5s.
* If trying to leave the page during a download a confirmation is asked
* if quit / reload the page, the download is aborted and the notification is gone


michaelkain:
This is an interesting feature, especially for our ongoing challenge MS-Multi-Spine, as downloads of big zip files can be very long. It would be great to test and merge it. With kind regards, Michael

DuckflipXYZ (approved):
All functionalities linked to the feature work correctly. Tested by downloading a subject

PR #2564, opened 2025-01-06
Allowing autocompletion on JComboBoxes for study cards and subjects

LaurentPV:
Allowing auto completion on study cards and existing subjects JComboBoxes.
It is not possible to edit the JComboBox with a non existing value.

It simplifies the ImportDialog.


michaelkain:
Hi @LaurentPV, thank you for this great and simplifying PR!
Code review is fine. I have only one point during my tests, that is not optimal:
for the study-cards, normally the users filter by using the device serial number, that is at the end of the study card name. They read it from above and filter by the number. I do not think, that this works like before with your change on study cards filter?

michaelkain:
Hi @LaurentPV, thank you for your PR.
Thx, you have corrected a "bug" I introduced within Subject.toString().
I was only thinking of having a better log and not of the side effect on the ImportDialog combobox.

michaelkain:
Hi, as soon as the filtering on studycards works with device serial number, I will merge it.
With kind regards, Michael

michaelkain:
Hi Laurent! Much better user interface now, great.
I found one issue during my tests, when I clean the study card combo box, see screenshot, I can click on Start auto-import and nothing happens. Should we gray the button?
Screenshot from 2025-01-17 10-42-04

michaelkain:
and I have this in the logs java.lang.ClassCastException: class java.lang.String cannot be cast to class org.shanoir.uploader.model.rest.StudyCard (java.lang.String is in module java.base of loader 'bootstrap'; org.shanoir.uploader.model.rest.StudyCard is in unnamed module of loader 'app')
        at org.shanoir.uploader.action.ImportFinishActionListener.actionPerformed(ImportFinishActionListener.java:72)

LaurentPV:
I added a new optional field isModalityMR (to not cause any exception with previous upload job without that field) in upload-job.xml to condition the display of study card selection.

My tests are fine, with XA dataset, MR dataset and previous imports.

michaelkain:
Hi @LaurentPV, thank you for your modifications.
For me, for the study cards, it was much better before, with a separate filter.
Please see my screenshot. It is quite confusing how the component behaves,
I can edit inside and produce strange lines. For the users, it is not even clear,
that they can filter on the study cards, as it looks like a regular drop-down.
Please set it back to before for the study cards at least, thank you!
Screenshot from 2025-03-12 09-23-42

michaelkain:
for the existing subject component it works fine

LaurentPV:
Rollback done to previous behavior of studyCard filter.
Ready to be merged.

michaelkain (approved):

Issue #2223, opened 2024-05-22
Display dicom metadata in dataset view


pierrehenri-dauvergne:
It would be easier for some users to display some specific dicom metadata in its dataset's detail view

Maybe we could merge this with #2195 and have the dicom's metadata next to the dataset's detail, instead of clicking a button?
Something like this:
![image](https://github.com/fli-iam/shanoir-ng/assets/117283961/0becc1b9-500f-4836-beaf-1c6dc10617b9)

Or maybe we could allow the user to pick some dicom's metadata field and add those field in the dataset's view?
![image](https://github.com/fli-iam/shanoir-ng/assets/117283961/590501b2-557b-4bc4-ac5a-ff7577f597b4)

julien-louis:
Note : see with Michael about Catherine that may have 2 more interesting fields to be directly added to the whole Shanoir data model instead of having this customizable display.

julien-louis:
### SliceThickness

* already present in MrProtocolMetadata
* acquisition level

**since the metadata is part of the acquisition protocol, I suggest adding it in relevant protocol types :**
* CtProtocol
* PetProtocol
* XaProtocol
* **Question : what about Bids ???**

### NumberOfSlices

* already present in PetProtocol
* acquisition level

**again, since the metadata is part of the acquisition protocol, I suggest adding it in relevant protocol types :**
* CtProtocol
* MrProtocol
* XaProtocol
* **Question : what about Bids ???*

### AcquisitionTime

* already present in MrDataset
* dataset level

**since the metadata is at dataset level, I suggest adding it in relevant dataset types :**
* CtDataset
* PetDataset
* XaDataset
* **Question : what about the other types ???*
    	* CalibrationDataset
	* CtDataset **[OK]**
	* EegDataset **[no]**
	* MegDataset
	* MeshDataset
	* MrDataset  **[OK]**
	* GenericDataset  **[no]**
	* ParameterQuantificationDataset
	* PetDataset  **[OK]**
	* RegistrationDataset
	* SegmentationDataset
	* SpectDataset
	* StatisticalDataset
	* TemplateDataset
	* BidsDataset  **[?]**
	* MeasurementDataset
	* XaDataset  **[OK]**

PR #2565, opened 2025-01-07
Add slice thickness, number of slices, acquisition start time

julien-louis:
see https://github.com/fli-iam/shanoir-ng/issues/2223

add sliceThickness & numberOfSlices (when not already existing) in :

* CtProtocol
* MrProtocol
* PetProtocol
* XaProtocol

add acquisitionStartTime in datasetAcquisition


julien-louis:
@LaurentPV it's ready now !

julien-louis:
@LaurentPV  I've added the missing db field

michaelkain:
not in next release, wait for dev in ShUp to use new features in quality cards

LaurentPV:
Tested, ready to be merged. New fields present at import, acquisition_start_time not included in the study or quality cards rules as it needs a new comparator and it seems irrelevant to use a start time as a potential condition for a study or quality card.

LaurentPV (approved):
Tested and approved.

michaelkain:
Hi @julien-louis,
thank you for this PR. There is an understanding bug on this PR, that is why I propose to revert it to have time to correct it.
NumberOfSlices is not the correct attribute, we require and it does not exist in MR, CT, PT, only in a very specific case in NM. Furthermore there is probably a bug with the parsing of the acquisition start time. to discuss.
And probably on this PR, you reopen maybe 12.000 dicom files to read an instanceNumber, that has already been calculated before. to discuss.
With kind regards, Michael
Screenshot from 2025-03-05 09-50-43

PR #2572, opened 2025-01-09
shanoir-issue#2560: stats export fails on big request

pierrehenri-dauvergne:
With this parameters: 
```
Study to include : Cohorte Principale.*|CISCO.*|Cohorte VHD.*
Subject to exclude : ^TestDummyRun.*|^test.*
```
The statistics export gives an empty file (doesn't even contain the headers).

With others parameters such as:
```
Study to include : Cohorte Principale - CHU Fort-de-France
```

It is working well

This PR is used to deploy and test on ofsep-qualif

This error is thrown:
o.h.e.jdbc.spi.SqlExceptionHelper - Java heap space

I think it happens in DatasetRepositoryImpl.queryStatistics method, on this line:
List<Object[]> results = query.getResultList();
Filling the List with 1+ millions line is filling the java heap space. A possible solution could be to increase the java heap space at application startup with something like this: java -Xmx1024m -Xms512m -jar app.jar but that wouldn't solve the issue for other bigger requests.

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
@@ -116,6 +116,9 @@ public class CreateStatisticsService {
                         eventService.publishEvent(event);
 
                         startRow += blocSize;
+                        if (startRow > procedureSize) {
+                            break;
+                        }
                     }
 
                     writer.flush();


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
```diff
@@ -82,6 +144,24 @@ public void createStats(String studyNameInRegExp, String studyNameOutRegExp, Str
         }
     }
 
+    private int querySize(String studyNameInRegExp, String studyNameOutRegExp, String subjectNameInRegExp, String subjectNameOutRegExp) {
```
michaelkain, line 147: good small method

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/repository/DatasetRepositoryImpl.java
```diff
@@ -14,31 +22,53 @@ public class DatasetRepositoryImpl implements DatasetRepositoryCustom {
     @PersistenceContext
     private EntityManager entityManager;
 
+	private static final Logger LOG = LoggerFactory.getLogger(DatasetRepositoryImpl.class);
     @SuppressWarnings("unchecked")
 	@Override
     public List<Object[]> queryStatistics(String studyNameInRegExp, String studyNameOutRegExp,
```
michaelkain, line 28: Method not needed anymore, I guess

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -367,6 +367,7 @@ public List<Dataset> findByExaminationId(Long examinationId) {
 
 	@Override
 	public List<Object[]> queryStatistics(String studyNameInRegExp, String studyNameOutRegExp, String subjectNameInRegExp, String subjectNameOutRegExp) throws Exception {
+		LOG.error("query statistics : " + studyNameInRegExp);
```
michaelkain, line 370: not LOG.error please and please log all RegExp to have the full info of the query in the logs

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
```diff
@@ -42,46 +47,124 @@ private File recreateFile(final String fileName) throws IOException {
...
+                            writer.write(String.join("\t", strings));
+                            writer.write('\n');
+                        }
+                        startRow += blocSize;
+                    }
+
+                    writer.flush();
+
+                } finally {
+                    outputStreamWriter.flush();
```
michaelkain, line 123: Hi @pierrehenri-dauvergne, I wonder if you missed the close here, what might be bad for the memory?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
```diff
@@ -42,46 +47,124 @@ private File recreateFile(final String fileName) throws IOException {
...
+            // Get the data
+            try (FileOutputStream fos = new FileOutputStream(zipFile);
+                 ZipOutputStream zos = new ZipOutputStream(fos)) {
+
+                ZipEntry zipEntry = new ZipEntry("shanoirExportStatistics_" + event.getId() + ".tsv");
+
+                zos.putNextEntry(zipEntry);
+                OutputStreamWriter outputStreamWriter = new OutputStreamWriter(zos);
+                BufferedWriter writer = new BufferedWriter(outputStreamWriter);
+                try {
```
michaelkain, line 77: maybe put the writer as well into try() to have auto-close it?

michaelkain (commented):
very close to finish

michaelkain:
One more thing: please adapt only the 3_add_statistics_procedure.sql, that concerns the ofsep-stats. The studyStats does not have to be impacted by your PR. Thanks!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
michaelkain: Hi @pierrehenri-dauvergne, I hope you are doing fine. Sorry, I should have been more clear above.
I think it would be worth to use a try-with-resources statement here?

michaelkain:
thank you! one minor item remains then we are ready, always better to use instead of an explicit close the try-with-resources, thx!

michaelkain:
Perfect for the code now. I approve.

michaelkain (approved):

michaelkain:
Hi @pierrehenri-dauvergne, can you please confirm, that it still runs on qualif? Thanks, if yes, I merge. Cheers, Michael

Issue #2518, opened 2024-11-26
BIDS subject zip generated on MacOS -hidden files


courvose:
When preparing BIDS subject files in zip from MacOS some system related hidden files are included.
In Shanoir import-BIDS, it creates some import error.
It is possible for users to  delete those files in the .zip archive before uploading them with bash command like:
zip -d [sub-XXX.zip] "*/.DS_Store"
and
zip -d [sub-XXX.zip] "__MACOSX/*"
It would be great if Shanoir could handle these extra files in the folder structure on the server side.

michaelkain:
Hi @alexpron, in case you might want to have a look on that? With kind regards, Michael

alexpron:
Thanks @courvose for reporting this issue. Sure @michaelkain I will give it a look

alexpron:
Pb solved I close the issue (other issue about filter class opened)

PR #2576, opened 2025-01-13
Shanoir issue#2518

alexpron:
This Pull Request aims at dealing with MAC OS added files within the BIDS data import. This PR solves issue #2518 
 ### MAC OS Hidden Files
There are several types of MAC Hidden files that are introduced either by Finder (and Zip) or by AppleDouble mechanism
see : 
+ [DS_store](https://en.wikipedia.org/wiki/.DS_Store#:~:text=DS_Store%20is%20a%20file%20that,Services%20Store%2C%20reflecting%20its%20purpose).
+ [AppleDouble](https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats) 
### Incompatibility with BIDS import
    1. At the subject or session check level an file other than starting with sub- or sess- prefix was making the import crash
    2. At data level these files if present in the datatype folder could be imported as real data 
### Solution
+ Filtering of the MACOS Hidden file to exclude them at the subject, session and data level
+ Also added fix to deal with empty _sessions.tsv file that there making the process crash (dont know if its better now let me know)

### DIFF
--- a/shanoir-ng-import/src/main/java/org/shanoir/ng/importer/bids/BidsImporterApiController.java
+++ b/shanoir-ng-import/src/main/java/org/shanoir/ng/importer/bids/BidsImporterApiController.java
@@ -317,7 +317,6 @@ public class BidsImporterApiController implements BidsImporterApi {
 		File sessionFile = sessionFiles[0];
 
 
-
 		// session_id;acq_time;pathology
 		// analyze date for every session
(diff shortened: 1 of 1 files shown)


alexpron:
Note that I found a parsing error in the json metadata file when numbers are written into scientific mode (i.e. 2e^-12).
Online json parser declare the .json file used compliant

PR #2581, opened 2025-01-15
2547 : back to initial transaction state (not splitted transaction)

DuckflipXYZ:
After reviewing with Mkain, testing with not splitted transaction (probably not needed, cuz no transaction operations stacking when indexing with solr, so no need to split transaction)


DuckflipXYZ:
If #2582 is already merged, SolrServiceImpl.indexAllNoAuth have to be replaced by this : 

@Async
    @Transactional
public void indexAllNoAuth() {
	List<ShanoirMetadata> documents = new ArrayList<>();
	Map<Long, List<String>> tags = new HashMap<>();
	ShanoirEvent event;

	try {
		event = new ShanoirEvent(
				ShanoirEventType.SOLR_INDEX_ALL_EVENT,
				null,
				0L,
				"Cleaning Solr index...",
				ShanoirEvent.IN_PROGRESS,
				0f);
		eventService.publishEvent(event);
		cleanOldIndex(event);
		fetchDatasToIndex(event, documents, tags);
		indexDatas(event, documents, tags);
	} catch (SolrServerException | IOException ignored) {
	}
}

michaelkain:
Hi @DuckflipXYZ, thank you! Did you probably forget to delete private SolrServiceImpl solrServiceImpl;?

DuckflipXYZ:
Hi @michaelkain ! You're right, transaction doesn't need to be splitted, I think the PR is relevant now

michaelkain:
Hi @DuckflipXYZ, thank you for this correction!
Did you test it locally and on ofsep-qualif? how long does it take now on ofsep-qualif? how is the memory behaving of ofsep-qualif? Could you please add a test case to test full Solr indexation in the new test case table on GitHub projects?

DuckflipXYZ:
Hello @michaelkain ! 

Yes, I tested it locally and on qualif ! It's not longer than before, and  I didn't take a look to the memory state because it worked the same as before (meaning without any issue :'). Tell me if you want a real check of the memory during the whole index.

Here is the test case : https://github.com/orgs/fli-iam/projects/1/views/2?pane=issue&itemId=94376438

michaelkain (approved):
Hi, I approve by code review only. With kind regards, Michael

PR #2583, opened 2025-01-15
push buildkit cache images in the OCI format

a-ba:
should fix `ERROR: target nifti-conversion: failed to solve: not found` when building the main branch


michaelkain (approved):

michaelkain:
Hi @a-ba, thank you a lot!

PR #2585, opened 2025-01-16
2558 : API for pipeline execution analysis

DuckflipXYZ:
Tested in local/remote. 

Linked with : https://github.com/fli-iam/py_noir/pull/11

Test case requirements :
[PyNoir project](https://github.com/fli-iam/py_noir) (Optional, but simplier)

Test case :

The idea is to make a call with the PyNoir Data API. Read the DataAnalysis README.md for more explanation about the API call. DataAnalysis context.conf requires to be modified too, refer to the root README.md for more details about the config.

2 feature to check (for every filter in {study, subject, examination, acquisition, dataset}). Care that a download filtered on a study can be kinda heavy :

- File is correctly downloaded 
- Its content is correct (compare with the processings in the web interface)

### DIFF
--- a/shanoir-ng-datasets/pom.xml
+++ b/shanoir-ng-datasets/pom.xml
@@ -136,6 +136,11 @@
 			<version>3.4.0</version>
 		</dependency>
 
+		<dependency>
+			<groupId>com.opencsv</groupId>
+			<artifactId>opencsv</artifactId>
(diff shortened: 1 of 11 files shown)


DuckflipXYZ:
Not working anymore, and this feature seems deprecated today.

PR #2587, opened 2025-01-17
auto clean of stale events

julien-louis:
* For events that has a "in progress" status, if no updates after 5 minutes the event is considered "error" and saved as so. The check runs when requesting the list of events so there is no extra db query.

* Also the filtering of the < 7 days is now done by the db query instead of the controller layer


DuckflipXYZ (approved):
1st feature works well, excepting that the web page needs to be refreshed for updating the job. Other thing, it does nto interrupt job, it only displayas them as "dead" jobs. I tested by adding 10 mins sleep to solr indexation, and the job ended correctly, even after the interruption display. I'm not sure if this is the expected behaviour or not. 

@julien-louis I let you merge that one if you think the refresh and the no job interruption are not issues

PR #2593, opened 2025-01-24
Tree scroll

julien-louis:
* auto scroll in tree
* opening body & equipment nodes wasn't displaying the tree : fixed


DuckflipXYZ:
Hi @julien-louis ! 

The feature works perfectly at first sight. One problem is that it only works once, the page needs to be refreshed for auto scrolling again

DuckflipXYZ (approved):
Hi @julien-louis ! No need to refresh anymore :) 

Last thing is that the autoscroll occurs only when dataset is loaded, which can take close to 1 min on ofsep qualif (within big study). Can you make an autoscroll step at the subject level while dataset is loading ? At least, we should be able to see the gearwheel rotating, as well as having an approximative location of the desired dataset. 

If you think that's useless, up to you to merge the PR :) 

++

DuckflipXYZ:
Nice update +1

PR #2603, opened 2025-01-28
Remove notif

julien-louis:
* adds a button to remove any notification from the left menu
* adds a study dl button, only for chrome and compatible browsers, with warning and size info

image

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApi.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApi.java
@@ -14,18 +14,13 @@
 
 package org.shanoir.ng.dataset.controler;
 
-import io.swagger.v3.oas.annotations.Operation;
-import io.swagger.v3.oas.annotations.Parameter;
-import io.swagger.v3.oas.annotations.responses.ApiResponse;
-import io.swagger.v3.oas.annotations.responses.ApiResponses;
-import io.swagger.v3.oas.annotations.tags.Tag;
-import jakarta.mail.MessagingException;
-import jakarta.servlet.http.HttpServletResponse;
-import jakarta.validation.Valid;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.util.List;
 
-import org.apache.solr.client.solrj.SolrServerException;
-import org.shanoir.ng.dataset.dto.DatasetWithDependenciesDTOInterface;
 import org.shanoir.ng.dataset.dto.DatasetDTO;
+import org.shanoir.ng.dataset.dto.DatasetLight;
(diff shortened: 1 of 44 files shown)


pierrehenri-dauvergne:
Tested locally and works well. To test on big studies maybe, make sure it's all good

pierrehenri-dauvergne:
Tested on a big study on ofsep-qualif (study Cohorte Principale - CHU Rennes):
When I click the "download study" button, the pop-up keeps loading. I thought it was measuring the total size the study would take on the disk but when it stopped loading, I got an error in console:
```
Error: datasets can't be null here
    at t._downloadDatasets (mass-download.service.ts:211:37)
    at mass-download.service.ts:121:29
```

Also, maybe we don't need to request the total size of the study since it's already displayed in the study's details:
image

I tried on another smaller study and it kept loading dispite the size being already displayed:
image
Why can't I start the download here?

After some time, it stops with this error in console, and the download fails:
```
GET https://shanoir-ofsep-qualif.irisa.fr/shanoir-ng/datasets/datasets/study/2 net::ERR_EMPTY_RESPONSE
getByStudyId | @ | dataset.service.ts:95
ngOnInit | @ | download-setup.component.ts:84
```

pierrehenri-dauvergne:
Also, we can remove the notif even before it's finished. Maybe wait until it's finished before allowing to remove it? I don't know what's best, I guess users could remove a notif by accident.

pierrehenri-dauvergne:
Hello @julien-louis 
On another PR, as some dataset can take a long time to delete, we made it async and visible with a task in the front. The issue is that when we delete lots of dataset in a row, it can create a big list of tasks like this:
image

Do you think we could add a button near "Jobs" to remove all tasks from the list?

julien-louis:
* There is definitely a bug coming from timeout requests
    * shanoir-ng/datasets/qualitycards/byStudy/17
    * shanoir-ng/datasets/datasets/study/17
* Remove a notif before it's finished raises difficulties and is not possible yet
* I will add a clear button

julien-louis:
@pierrehenri-dauvergne ready now

pierrehenri-dauvergne:
Hi @julien-louis. I just tested on ofsep-qualif:
- Study download worked fine the first time, but I haven't been able to download another study after. I click the "download" button in the download pop-up but it does nothing, no request is being sent. I'm not sure what's going on.
- We can still remove a notif of an ongoing task
- Clicking anywhere in this red rectangle doesn't load the "Jobs" page, even if the "clean all tasks" isn't displayed.
image
The other lines of the side-menu can be opened by a click anywhere on the line, not just on the text, it makes me feel like the link is not working properly when it's just that I'm not clicking on the text. Could we revert this?

julien-louis:
@pierrehenri-dauvergne 
1. yes I have bring back the total progression bar
2. you must be on Chrome or compatible browser

pierrehenri-dauvergne (approved):

Issue #2606, opened 2025-01-30
OHIF doesn't load images


pierrehenri-dauvergne:
When opening the OHIF viewer, the images don't load and this error is thrown in the logs:

> 2025-01-30_16:36:36.657 ERROR [phdauvergne(1120)] o.s.n.d.web.service.DICOMWebService - Content-Length must be a non-negative number
java.lang.IllegalArgumentException: Content-Length must be a non-negative number
        at org.springframework.http.HttpHeaders.setContentLength(HttpHeaders.java:1005)
        at org.shanoir.ng.dicom.web.service.DICOMWebService.findFrameOfStudyOfSerieOfInstance(DICOMWebService.java:190)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
        at java.base/java.lang.reflect.Method.invoke(Unknown Source)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:359)
        at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:723)
        at org.shanoir.ng.dicom.web.service.DICOMWebService$$SpringCGLIB$$0.findFrameOfStudyOfSerieOfInstance(<generated>)
        at org.shanoir.ng.dicom.web.DICOMWebApiController.findFrameOfStudyOfSerieOfInstance(DICOMWebApiController.java:221)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
        at java.base/java.lang.reflect.Method.invoke(Unknown Source)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:359)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:196)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
        at org.springframework.security.authorization.method.AuthorizationManagerBeforeMethodInterceptor.proceed(AuthorizationManagerBeforeMethodInterceptor.java:268)
        at org.springframework.security.authorization.method.AuthorizationManagerBeforeMethodInterceptor.attemptAuthorization(AuthorizationManagerBeforeMethodInterceptor.java:263)
[...]

PR #2612, opened 2025-02-03
Fix for GitHub issue #2606  related to Spring Boot 3.4.0 update

michaelkain:
Spring Boot 3.4.0 and Apache Http Client Release 5.4  are probably more restrictive on setContentLength(x smaller than 0), so we set it only in case of existing value to avoid the exception


michaelkain:
Successfully tested locally: at first: not working, then fix: working

Issue #2570, opened 2025-01-08
duplicated study-center (quite minor)


julien-louis:
* edit a multi-center study
* remove a center
* don't save
* put back this same center
* save study
===> this center is duplicated 

(backend issue)

PR #2613, opened 2025-02-03
Fix study center edit

julien-louis:
fix this issue : 
https://github.com/fli-iam/shanoir-ng/issues/2570

Also fix this bug : 
sometimes, trying to remove a study tag (as in a public study) fails with an error saying the tag is in use. It has no sense, actually the code mixes up the study public tags and the study tags that are available for linking them to subject-studies.

It's hard to reproduce as a study public tag has to have the same id as a subject-study tag.


pierrehenri-dauvergne (approved):

pierrehenri-dauvergne:
Tested on qualif, couldn't reproduce the tag issue but I didn't find any issue

PR #2624, opened 2025-02-05
shanoir-issue#2622: unfold exam takes too long

pierrehenri-dauvergne:
How to test:
On qualif, go to solr, find a subject with lots of datasets
Open the study
Unfold the node (in the tree view) corresponding to that subject

It keeps loading indefinitely, without any error thrown


pierrehenri-dauvergne:
Potential fix https://github.com/fli-iam/shanoir-ng/pull/2629

PR #2625, opened 2025-02-05
Hotfix shanoir issue#2518 (MAC OS hidden files handling  in BIDS archive

alexpron:
Hotfix to handle properly BIDS same code as closed PR #2576 answer issue #2518 (hotfix was created from master directly 
as develop might be under regression) 

+ mac OS hidden files filtered out at all level of the archive (sub-, ses- (if any), and datatype)
+ correct parsing of .json metadata file (if any) that previously crashed due to dcm2niix scientific notation

PR has been tested locally and I will test on qualification server

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/BidsImporterService.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/importer/service/BidsImporterService.java
@@ -35,6 +35,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.io.File;
+import java.io.FilenameFilter;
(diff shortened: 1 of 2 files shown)


alexpron:
I tested the PR on neurinfo qualif and everything seems to work fine, ready for review

michaelkain:
Hi @alexpron, thank you for your PR! Great.
Do you think you might create a separate class into ms-common or as probably only required by MS Import
to add the class there, that implements FilenameFilter and inject it as a component into your BIDS classes? it would be great to inject and use it as well in the ImporterApiController for the dicom zip file import upload as well please. Thank you!

michaelkain:
Hi, I close as already on develop. Cheers, Michael

PR #2628, opened 2025-02-06
Bids import fix

julien-louis:
* upgrade a log msg
* fix a bad rabbitMQ msg
* use Long instead of long so the undefined id is not serialized by "0", causing jpa errors


julien-louis:
Hi @alexpron could you review this please ?

alexpron (approved):
+ cleaner JAVA imports
+ better error message
+ rabbimq call fix (long --> Long) changed into Subject.java so that it is coherent

I tested on neurinfo qualif with:
+  with session (ses-) subdirectories bids zips 
+ without session (BIDS datatypes directly)
Both works perfectly fine
Thank you @julien-louis

Issue #2622, opened 2025-02-05
Broken pipe error / ms-datasets crashes when loading lots of datasets


pierrehenri-dauvergne:
Open solr, pick a subject with lots of datasets
Open this subject in the tree-view and unfold it

![Image](https://github.com/user-attachments/assets/37754bde-e6e6-4a4b-ad11-2243d30e0755)

The load never ends, at some point it seems like ms-datasets crashes, along with solr

We can sometimes see this error in datasets logs:
```
2025-02-04_12:16:36.976 ERROR [phdauvergne(203)] o.s.n.s.c.GlobalExceptionHandler - Unexpected error in the rest service.
org.apache.catalina.connector.ClientAbortException: java.io.IOException: Broken pipe
        at org.apache.catalina.connector.OutputBuffer.realWriteBytes(OutputBuffer.java:344)
        at org.apache.catalina.connector.OutputBuffer.flushByteBuffer(OutputBuffer.java:773)
        at org.apache.catalina.connector.OutputBuffer.append(OutputBuffer.java:676)
        at org.apache.catalina.connector.OutputBuffer.writeBytes(OutputBuffer.java:379)
        at org.apache.catalina.connector.OutputBuffer.write(OutputBuffer.java:357)
        at org.apache.catalina.connector.CoyoteOutputStream.write(CoyoteOutputStream.java:97)
        at org.springframework.security.web.util.OnCommittedResponseWrapper$SaveContextServletOutputStream.write(OnCommittedResponseWrapper.java:639)
        at org.springframework.util.StreamUtils$NonClosingOutputStream.write(StreamUtils.java:264)
        at com.fasterxml.jackson.core.json.UTF8JsonGenerator._flushBuffer(UTF8JsonGenerator.java:2203)
        at com.fasterxml.jackson.core.json.UTF8JsonGenerator.writeFieldName(UTF8JsonGenerator.java:271)
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:730)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:772)
        at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:119)
[...]

PR #2629, opened 2025-02-07
Hot fix to avoid huge sql related to copies (where source_id = ?)

michaelkain:
Potential solution for https://github.com/fli-iam/shanoir-ng/issues/2622
successfully tested on ofsep-qualif with subject, that has 7 exams and therefore the most exams


michaelkain:
Hi @julien-louis, thank you for your comment. Have you seen, that the checks are still on the service? With kind regards, Michael

Issue #2627, opened 2025-02-06
StudyUser(Rights) panel broken (prod + qualif)


michaelkain:
Hi @julien-louis,
I hope you are doing fine.
When I am on ofsep-prod and click on a member in my study, I get the following error:
09:52 - Error : TypeError x153
    e.pannelStudyUser.user is undefined
09:51 - Error : TypeError x51
    e.pannelStudyUser.user is undefined
and the screen looks weird.
With kind regards,
Michael

DuckflipXYZ:
![Image](https://github.com/user-attachments/assets/cd8780e7-1884-4978-9e52-72fc8009900f)

In the screen above, some points to talk about : 

the "general" part in the study user panel is not displayed when the current user role is not expert nor admin, so it's fine with the 1st fix, but the "general rights part" is displayed, and the role field is not accessible if the current user role is "user". This role is displayed too in the user list at the left of the study user panel.

Those role fields have to be disabled when the current user has "user" as role.

DuckflipXYZ:
Already fied, not relevant anymore

PR #2632, opened 2025-02-10
#2627 : studyUserPannel repaired when not admin rights

DuckflipXYZ:


michaelkain (approved):
Hi @DuckflipXYZ,
thank you a lot. Excellent work!
I tested it locally and it works. Perfect. With kind regards, Michael

PR #2633, opened 2025-02-10
#2627 : better web page management according to current user role

### LINKED ISSUE #2627: StudyUser(Rights) panel broken (prod + qualif)

DuckflipXYZ:
line 133 and 154 for .ts file


pierrehenri-dauvergne:
This is what I see as a user in a study where I'm a study admin:
image

Same but as an expert:
image

We talked a bit with @michaelkain about this and:
- as a user, there is no reason to hide the role of other members of the study
- as a study admin and expert, there is no reason to hide the email address 

Also, the "expert" role doesn't have a icon (it should have the tie, the user has it right now)

pierrehenri-dauvergne:
Hello @DuckflipXYZ 
It's all good for me except for this small icon 😄 
image

The expert icon should be the one with a tie, and for the user, the icon should not have a tie at all

michaelkain:
I think it would be great to finish this please.
Thank you! Michael

a-ba:
done

shanoir-ng-front/src/app/studies/study/study.component.ts
```diff
@@ -198,14 +198,7 @@ export class StudyComponent extends EntityComponent<Study> {
...
         }
-        if (this.keycloakService.isUserAdminOrExpert()) {
-            return this.fetchUsers().then(users => {
-                Study.completeMembers(this.study, users);
-            });
-        } else {
-            return Promise.resolve();
-        }
-
+        return this.fetchUsers().then(users => {Study.completeMembers(this.study, users);});
```
fragkiadamis, line 201: Here you try to fetchUsers for all the the user roles.

In the backend, the full user list is authorized **only** to admin and expert roles. As a result, if you try to load the study as a simple user, you get errors from the users container and nothing loads.

fragkiadamis (changes requested):

shanoir-ng-front/src/app/studies/study/study.component.ts
DuckflipXYZ: I guess it's normal behaviour, a user would have nothing to do with the user list. I'm wondering if the tab should be disabled with admin role, what do you think ?

DuckflipXYZ:
@michaelkain Do you think we can remove user tab access for "User" ? They don't need it from my pov

DuckflipXYZ:
@michaelkain up :)

DuckflipXYZ:
Not relevant anymore, closed

PR #2637, opened 2025-02-11
shanoir-issue#2635: fix get size procedure

pierrehenri-dauvergne:
How to test:
Make the same stat export before and after deploying this PR
Before: the job takes too long, at some point if you refresh the page you can see "inactivity timeout, there must has been" (message is cut)

After: the job works properly


pierrehenri-dauvergne:
Tested on ofsep-qualif successfully.

michaelkain (approved):

PR #2639, opened 2025-02-12
Shanoir issue#2518 (MACOS Hidden files)

alexpron:
Another PR to solve this problem (Hotfix in prod was cancelled we go through develop since MEP delayed)


alexpron:
Tested on qualif with several archives ! Works fine
Remark: BIDS import process still fragile : not robust to empty metadata files, date formatting etc.

michaelkain (approved):
Hi @alexpron, thank you for this PR!
I merge it to fix the immediate problem, perfect and thank you for your tests!
In a second tour it would be great to centralize it in a dedicated Filter class,
inject it with @AutoWired as component and use it as well in the ImporterApiController, where the zip file is uploaded to escape those file in a larger context.
Please create an issue and assign it to you to not forget this upgrade.

PR #2653, opened 2025-02-18
2646 : Vip executions tracking

DuckflipXYZ:

PR #2671, opened 2025-02-25
Maven build test correction

DuckflipXYZ:
Hello ! 

Core fixes are : 

- The creation of a RabbitTemplate configuration. It should be auto-configured by Spring, but for a mysterious reason it's not, so instead we configure it manually with a @Config @ComponentScan class.
- The addition of a @Lazy tag, that breaks a circular dependency which is (it starts with Bids in the cases that I experimented, but it can happens in other way I guess ...): 
        --> BidsAPIController --> BIDSServiceImpl --> ExaminationServiceImpl --> StudyInstanceUIDHandler -> ExaminationServiceImpl --> ...
        
Rest of modified code was not used anymore, so removing them save resources. Ah, and my IDE says that @MockBean is deprecated and from what I see, @MockitoBean is the standard replacement.


michaelkain:
Hi @DuckflipXYZ, thank you for this PR.
To be honest, I do not know, what the background and the goal is?
Please explain in the weekly. Thank you, Michael

DuckflipXYZ:
The point is to resolve issues which can occur at maven build step. For example, merge this to the PR #2653 should correct the maven build check made on github

pierrehenri-dauvergne:
As this is unclear what this PR is really fixing, this should be put on stand by

PR #2673, opened 2025-02-26
update open-api version with one compatible with Spring Boot 3.4

LaurentPV:
Fixing the following issue :

Capture d’écran 2025-02-26 à 15 01 00


LaurentPV:
Due to an incompatibility between spring boot v3.4 and open-api v2.2

pierrehenri-dauvergne (approved):

PR #2676, opened 2025-02-27
fix performance issues with lots of subject nodes

julien-louis:
the solution here is to fake the nodes until any interraction


pierrehenri-dauvergne:
Hi @julien-louis Tested on ofsep-qualif like this:
Open solr
Filter by study "strasbourg", check the one with lots of datasets
Go to page 3000 of solr results
Open a dataset

It takes a long time to open the details with the tree but once the details are loaded, the node of the subject keeps loading:
image

There are "Broken pipe" errors in the logs but I don't see what causes them

julien-louis:
change "hidden"

julien-louis:
@pierrehenri-dauvergne @michaelkain I think it's ok now, for subjects + members

The bug mentioned by ph is a backend issue where shanoir-ng/datasets/examinations/subject/xxx/study/xxx takes so much time it ends up in a client timeout, I will investigate.

julien-louis:
It seems that the problem has been fixed elsewhere.

michaelkain:
Hi @pierrehenri-dauvergne, hi @julien-louis,
I hope you are doing fine. Do you have any update on this PR?
It is tagged next-release and it would be great to merge it, if functional.
Cheers, Michael

pierrehenri-dauvergne (approved):
Tested on both qualif and it worked well

PR #2685, opened 2025-03-05
fix quality cards crash

julien-louis:
remove parallel quality


julien-louis:
* remove parallel quality run
* prevent the front to wait for the response
    * @Async service
    * the front sets the task id then listen to tasks stream to catch the one ending
* replaceAll has bad perf : use replace instead
* don't loop over all file path urls when only the first one is needed
* close JsonParser with try() as JsonParser is Closeable
* in the ExaminationAttributes object, store only in use DICOM tags
* on another branch, tried to use a stream
    * https://github.com/julien-louis/shanoir-ng/tree/quality-crash-stream
    * from request to PACS
    * to the dcm4chee JsonParser
    * fails : the dcm4chee JsonReader couldn't read the data

* couldn't run JFR (Java Flight Recorder)
* used ManagementFactory.getMemoryMXBean().getHeapMemoryUsage();
    * heap space increases when parsing (and fetching ?) the dicom json data
    * the garbage collector seem to free space regulary, but not all and it accumulates to crash

julien-louis:
After analyzing the heap dump, it's Hibernate that uses all this memory.

* org.hibernate.engine.internal.StatefulPersistenceContext takes half the memory of the threads (one thread per user quality request)

Screenshot from 2025-04-22 09-17-29

* the rest seems to be used by a lot of org.hibernate.collection.spi.PersistentBag

Screenshot from 2025-04-22 09-19-20

PR #2692, opened 2025-03-11
Re add metadata (slice thickness, ...) + fix heavy file read + try catch date parsing

julien-louis:
see https://github.com/fli-iam/shanoir-ng/issues/2223

add sliceThickness & numberOfSlices (when not already existing) in :

* CtProtocol
* MrProtocol
* PetProtocol
* XaProtocol

add acquisitionStartTime in datasetAcquisition

it's a "retry" of https://github.com/fli-iam/shanoir-ng/pull/2565


LaurentPV:
I rolled back to the previous method used to count unique instances in DicomProcessing (from the previous PR 2565) because the new way of counting was always returning 1 as it took only attributes from first instance.

LaurentPV (approved):
Tested and approved

PR #2582, opened 2025-01-15
1741 : solr auto exec complete

DuckflipXYZ:
Not tested in remote

If #2581 is already merged, SorlServiceImpl.indexAllNoAuth() have to be replaced by this (done) : 

	@Async
        @Transactional
	public void indexAllNoAuth() {
		List<ShanoirMetadata> documents = new ArrayList<>();
		Map<Long, List<String>> tags = new HashMap<>();
		ShanoirEvent event;

		try {
			event = new ShanoirEvent(
					ShanoirEventType.SOLR_INDEX_ALL_EVENT,
					null,
					0L,
					"Cleaning Solr index...",
					ShanoirEvent.IN_PROGRESS,
					0f);
			eventService.publishEvent(event);
			cleanOldIndex(event);
			fetchDatasToIndex(event, documents, tags);
			indexDatas(event, documents, tags);
		} catch (SolrServerException | IOException ignored) {
		}
	}

PR #2707, opened 2025-03-13
1741(bis): solr auto exec

a-ba:
This is a fix for #2582.

The ConnectException had to be caught inside the while loop. Also i simplified the overal design of the function (just repeat the solr query until we have no connection error).


DuckflipXYZ (approved):
Clearly a more concise code.. seems good to me !

PR #2708, opened 2025-03-13
possible memory leak ?

julien-louis:
ByteArrayInputStream implements Closeable so it should be closed. 

I'm not sure if it was closed automatically closed but if we have a doubt I propose to move it into the try () so we ensure that it will be closed.


pierrehenri-dauvergne (approved):
Tested locally, import worked

PR #2714, opened 2025-03-14
2698 (real 2712) : API for post processing VIP execution result

### LINKED ISSUE #2712: Allow post use/processing of processing dataset with API call
The idea is to allow processing of processed dataset whenever we want with an API call, not only directly after VIP execution. See processing of processed dataset in vip/handler.

DuckflipXYZ:
Tested in remote. Seems to work, ask me for testin, it's a kinda heavy procedure


DuckflipXYZ:
Not relevant anymore, closed

PR #2720, opened 2025-03-18
#2642 : OFSEP SIMS output processing correction

### LINKED ISSUE #2642: Bug : Wrong display of acquisition/dataset name
From a mail discussion with Céline Homo :

En regardant de plus près, en fait le champ serie.name est bien dans l’export mais pas avec la bonne valeur. C’est la valeur du champ volume.name qui est sorti. De plus, le champ serie.protocolValidityStatus est toujours vide dans l’export alors qu’il y a bien une valeur dans l’output.json. 

////////////////////////////////////////////////////////////////////

Pour les dataset acquisition qui sont composés d’un volume DP et d’un volume T2, dans le front Shanoir, les noms affichés du dataset acquisition et des 2 datasets qui le composent est « T2DAXDP ». Le MR sequence name du dataset_acquisition = AX T2 DP. Dans l’output json il y a bien un volume T2DAXT2 et un volume T2DAXDP et le serie-type = AX T2 DP. Il semblerait donc qu’il y ai un problème de reprise des données et donc ce qui est affiché sur le front n’est pas bon. Comme exemple, il y a le dataset_acquisition 116694  composé des datasets 156562 et 156563. Serait-il possible de corriger cela pour que les bonnes informations soient reprises ?

Please care that's not an incompatibility with neurinfo dev (I have no idea about that, but it can be).

DuckflipXYZ:
Not tested yet

INCLUDES : https://github.com/fli-iam/shanoir-ng/pull/2905

### DIFF
--- a/docker-compose/database-migrations/db-changes/datasets/0067_add_statistics_procedure.sql
+++ b/docker-compose/database-migrations/db-changes/datasets/0067_add_statistics_procedure.sql
@@ -1,0 +1,211 @@
+-- Shanoir NG - Import, manage and share neuroimaging data
+-- Copyright (C) 2009-2019 Inria - https://www.inria.fr/
+-- Contact us on https://project.inria.fr/shanoir/
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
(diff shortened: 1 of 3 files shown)


DuckflipXYZ:
Last commit removes a test, because a DB link is used in modified method, but this link does not exist with the test data ... I tried to fake the modified method using that link with mockito library but I didn't achieve ...

PR #2723, opened 2025-03-19
ShUp: add .csv stats for mass imports with table

michaelkain:
- Creates 2 .csv files in the workFolder, with a time stamp == one run of mass import from table (Excel)
OK contains, what worked well, KO, where there was a problem or the exam existed already with the
same date in Shanoir (no contained series comparison)
- Changes as well the logic slightly in selectPatientStudyAndSeries, maybe better now for local debug
in case two studies are in the pacs, with the same exam date. Now based on StudyInstanceUID should
make this more stable


LaurentPV (approved):
Tested and approved.

PR #2725, opened 2025-03-19
update anonymization table

julien-louis:
this table is read during import to chose the anonymization for the 600+ specified dicom fields, before copying files into the pacs


michaelkain:
Hi @julien-louis, thank you a lot for this PR. Great, that we have it.
This PR requires a special collaboration with all consortium members,
that is why we can only merge it after close coordination with them.

julien-louis:
After the meeting with representatives of both instances I made slight modifications (bolus start and stop time) and now we are ready to go.

PR #2739, opened 2025-03-24
shanoir-issue#2738: add JsonIgnore on profile field

pierrehenri-dauvergne:
How to test:
Open the study-list, no error
Make a multiple examination import (see [#2728](https://github.com/fli-iam/shanoir-ng/pull/2728) )


pierrehenri-dauvergne:
Tested locally and on neurinfo-qualif and it's working (PR 2728 too).

michaelkain (approved):

Issue #2683, opened 2025-03-04
Chrome: download dialog, add seriesNumber checkbox


michaelkain:
Hi @julien-louis,
I hope you are doing fine.
I exchanged with a researcher using Shanoir today.
It would be great to add a level acquisition (== DICOM serie) to the download options
at the bottom, that the downloads have one folder ordered by seriesNumber (sorting index in acquisition).
If this is not much development, it would be great to integrate it in the next weeks.
With kind regards, Michael

PR #2744, opened 2025-03-25
add acquisition folder option for download

julien-louis:
fixes https://github.com/fli-iam/shanoir-ng/issues/2683

* added acquisition folder as an option in download
* false by default
* name = sorting index (seriesNumber)
    * \+ protocol.metadata.name (if mr)
    * \+ type (else)
    * \+ id

image


DuckflipXYZ (approved):
Feature works well, and code seems good.

PR #2754, opened 2025-03-26
bugfix failure to retrieve subjectStudy for quality control

LaurentPV:
Correcting bug in quality control execution at Import from Shanoir Uploader


michaelkain:
not next release, bundle with fixes for QC mass processing

Issue #2767, opened 2025-03-31
Add existing subject to study crashes


pierrehenri-dauvergne:
While making an import, create a new subject
Finish the import
Go to the newly created subject's details
Edit
Add the subject to another study
Update

ms-studies throws this error:
```
2025-03-31_11:09:37.508 ERROR [phdauvergne(1120)] o.s.n.s.c.GlobalExceptionHandler - Unexpected error in the rest service.
org.springframework.dao.DataIntegrityViolationException: could not execute statement [Duplicate entry '84-19271' for key 'study_subject_idx'] [insert into subject_study (physically_involved,quality_tag,study_id,subject_id,subject_study_identifier,subject_type) values (?,?,?,?,?,?)]; SQL [insert into subject_study (physically_involved,quality_tag,study_id,subject_id,subject_study_identifier,subject_type) values (?,?,?,?,?,?)]; constraint [study_subject_idx]
        at org.springframework.orm.jpa.vendor.HibernateJpaDialect.convertHibernateAccessException(HibernateJpaDialect.java:290)
        at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:241)
        at org.springframework.orm.jpa.AbstractEntityManagerFactoryBean.translateExceptionIfPossible(AbstractEntityManagerFactoryBean.java:560)
        at org.springframework.dao.support.ChainedPersistenceExceptionTranslator.translateExceptionIfPossible(ChainedPersistenceExceptionTranslator.java:61)
        at org.springframework.dao.support.DataAccessUtils.translateIfNecessary(DataAccessUtils.java:343)
        at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:160)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
        at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.invoke(CrudMethodMetadataPostProcessor.java:165)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
        at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:223)
[...]

PR #2768, opened 2025-03-31
shanoir-issue#2758: add task for deletion of subject + make only one call

### LINKED ISSUE #2758: Delete subjects should create a task
When you delete a subject, it deletes all of its examinations, acquisition and datasets, but there is no task showing its progress or any error met.

pierrehenri-dauvergne:
How to test:
Make an import
Create a new subject for this import
Open the subject's details and click Delete
A new task should be visible
In each of the subject's studies, in the history tab, we should see that the subject was deleted (to test this, a subject should be in multiple studies but https://github.com/fli-iam/shanoir-ng/issues/2767 prevent this test)

FYI: a subject can be REMOVED from a study or DELETED from all of its studies. There are two different ShanoirEvents but both should appear in the study's history


DuckflipXYZ:
Not consistent anymore

PR #2770, opened 2025-04-01
shanoir-issue#2767: add transactional and return saved subject

pierrehenri-dauvergne:
How to test:

Create a subject, add it to a study
Edit that subject and add it to another study
Click update


DuckflipXYZ (commented):
Hello !

All seems good except one case, which I don't know if it has to be managed ..

When editing, if you remove the wrong study, and you try to added it again, the update/save button will raise an error.

If we don't care, it can be merged.

See you

michaelkain (approved):
Nice!

PR #2777, opened 2025-04-03
Fix download popup very slow

julien-louis:
Loads a light version of the dataset list.

To test : 

* in solr select several datasets including 899038 (that has a lot of input datasets)
* open the download dialog
* very slow (even crashing) without the pr / very fast with the pr

! some files in the VIP process where edited, must test VIP

I'm issuing another pr to set hibernate LAZY loading where it should be : [TO DO]


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApiController.java
```diff
@@ -414,9 +408,9 @@ public void massiveDownloadByDatasetIds(
 		}
 
 		// STEP 1: Retrieve all datasets all in one with only the one we can see
-		List<Dataset> datasets = datasetService.findByIdIn(datasetIds);
+		//List<Dataset> datasets = datasetService.findByIdIn(datasetIds);
```
DuckflipXYZ, line 411: I'm not sure about those comments, I think it breaks the download.

DuckflipXYZ (changes requested):
Pop up is instantly loaded, which is really nice, but download is not working anymore

julien-louis:
@DuckflipXYZ done !

DuckflipXYZ (approved):
Code seems good, feature works well

PR #2778, opened 2025-04-07
Lazy loading in processings

julien-louis:
Add Hibernate Lazy loading for : 
* Dataset.processing (optional input processing, aka processing where the dataset is an output)
* Dataset.processings (optional output processings, aka all proessings the dataset is an input of)
* Processing.inputDatasets

- Test processing view / create / edit / tree

Introduce a new Dataset DTO which is more suitable for lists of dataset : DatasetWithDependenciesForListsInterface
* affect only the API datasets/allById
    * used in mass downloading
    * used in vip : execution.component.ts l95

- Test mass downloading (chrome + firefox ?)
- Test VIP ?


julien-louis:
@MichaelKain Ok it's merge. But I did not realize that the new DTO for lists would be useless with the DatasetLight from the previous pr. So now this pr contains
* new LAZY configs
* a ready-to-use dataset dto for lists usage

now the question is : should we merge that or should we use this DTO where it could be (findDatasetsByExaminationId, findDatasetsByAcquisitionId, findDatasetsByStudycardId, findDatasetByStudyId, findDatasetsBySubjectIdStudyId) ?

julien-louis:
@DuckflipXYZ put this one aside, test https://github.com/fli-iam/shanoir-ng/pull/2782 instead

Issue #2763, opened 2025-03-27
ShUp (OFSEP): KO


michaelkain:
All imports with ShUp are broken, because of this exception

2025-03-27_17:40:19.846 ERROR [mkain1(255)] o.s.n.s.c.GlobalExceptionHandler - Unexpected error in the rest service. 
org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: org.shanoir.ng.subjectstudy.model.SubjectStudy.subjectStudyTags: could not initialize proxy - no Session
	at org.hibernate.collection.spi.AbstractPersistentCollection.throwLazyInitializationException(AbstractPersistentCollection.java:636)
	at org.hibernate.collection.spi.AbstractPersistentCollection.withTemporarySessionIfNeeded(AbstractPersistentCollection.java:219)
	at org.hibernate.collection.spi.AbstractPersistentCollection.initialize(AbstractPersistentCollection.java:615)
	at org.hibernate.collection.spi.PersistentBag.clear(PersistentBag.java:469)
	at org.shanoir.ng.subject.service.SubjectServiceImpl.lambda$**loadSubjectStudyTags**$2(SubjectServiceImpl.java:380)
	at java.base/java.util.Iterator.forEachRemaining(Unknown Source)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Unknown Source)
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(Unknown Source)
	at org.shanoir.ng.subject.service.SubjectServiceImpl.loadSubjectStudyTags(SubjectServiceImpl.java:379)
	at org.shanoir.ng.subject.service.SubjectServiceImpl.**findByIdentifierInStudiesWithRights**(SubjectServiceImpl.java:338)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:359)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:196)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
[...]

PR #2779, opened 2025-04-08
add transactional

julien-louis:
may fix https://github.com/fli-iam/shanoir-ng/issues/2763

(waiting for info to be able to test)


michaelkain:
my local test with profile ofsep worked, I merge

michaelkain (approved):

PR #2782, opened 2025-04-09
lazy processing in dataset & dataset in processing

julien-louis:
Add Hibernate Lazy loading for :

* Dataset.processing (optional input processing, aka processing where the dataset is an output)
* Dataset.processings (optional output processings, aka all proessings the dataset is an input of)
* Processing.inputDatasets

Test processing view / create / edit / tree


julien-louis:
@DuckflipXYZ please review this one

DuckflipXYZ (approved):
No noticeable performance approvement, but maybe the data tests were not relevant. At least, it's not breaking anything. We have to remember that one in case of issue about dataset_processing.

PR #2787, opened 2025-04-11
Laurent issue#2343 sh up no xml

LaurentPV:
This PR modifies the behavior of Shanoir Uploader to only have the import-job.json replacing the previous upload-job.xml and nominative-data-job.xml files associated with each imports.

The Patient and Study previously deleted from importJob before writing the json file are no longer deleted to allow the use of those informations.

UploadState and uploadPercentage were added as ImportJob variables in import ms.

Compatibility with previous ShUp versions imports is assured and data retrieved from xml files if absent from import-job.json


michaelkain:
Hi Laurent, thank you for this PR! It looks promising.
To finally really get rid of the XMLs, we should as well remove the libraries and see, that all works well.
E.g. I think of:
		<dependency>
		    <groupId>jakarta.xml.bind</groupId>
		    <artifactId>jakarta.xml.bind-api</artifactId>
		    <version>4.0.0</version>
		</dependency>
		<dependency>
		    <groupId>com.sun.xml.bind</groupId>
		    <artifactId>jaxb-impl</artifactId>
		    <version>4.0.0</version>
		    <scope>runtime</scope>
		</dependency>
		
With kind regards, Michael

michaelkain:
I will continue my review, if the libs are gone too (in the idea of the possible, of course)

michaelkain (approved):
Hi Laurent, good and solid PR. My local tests worked, fromFile and fromPACS,
to neurinfo- and ofsep-qualif. To progress and not increase again the size of
the PR, I merge. Work to do: 1) remove XML libs, 2) stop writing twice the
import-job.json in FinishAction(Listener or Runnable) 3) remove reference
to upload-job.xml in one file. Let's see, who does it. Cheers, Michael

Issue #2790, opened 2025-04-15
Download dialog (FF + Chrome): no selection dicom/nifti


michaelkain:
ofsep-qualif/prod: e.g. dataset 932913: qualif: no selection between dicom/nifti, prod: drop box to select

DuckflipXYZ:
Hello ! 

A check about the fact that datasets are outputDataset or not was updated, but the logic was changed : before, it checked that datasets were not output, but now it checks that datasets are not used as inputs. 

I just put a check doing : datasetType != Generic, as currently all the outputs have generic type (and if dicom is typed as generic, well, should it be typed as MR dataset instread ? . Moreover, outputs might be Dicoms in the futur, so the check  on output or not was wrong according to me.

PR #2791, opened 2025-04-15
#2790 : download pop up correction

DuckflipXYZ:
Tested in qualif. Check the issue for more details


michaelkain:
Hi @DuckflipXYZ,
thank you! Looks good. We introduced GenericDatasets for DICOM files, that have an imaging modality, we do not cover today. I think, GenericDatasets, should be accessible as DICOM download. If it works, is another topic. Cheers, Michael

pierrehenri-dauvergne (approved):
Tested on ofsep-qualif and worked well

PR #2795, opened 2025-04-17
ShUp: examination consistency checker

michaelkain:
- see JavaDoc in ExaminationConsistencyServiceJob
- calls server on FINISHED imports and checks images via DicomWeb interface
- new UIManager for better LookAndFeel
- add configuration menu to de-activate ECC
- add configuration menu for pseudo after copyOrDownload (not yet implemented)

Tested locally successfully:
- import from disk
- import from pacs


LaurentPV:
During tests I noticed that an empty exam on server is sufficient for the /dicomWeb/study/ http request to return a 200 code and then the status is CHECK_OK even if no serie is present on server side.

shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/CurrentNominativeDataController.java
```diff
@@ -79,7 +79,7 @@ public void mouseClicked(MouseEvent e) {
 						boolean uploadsToDelete = false;
 						for (int i = 0; i < rows; i++) {
 							String uploadState = (String) cuw.table.getModel().getValueAt(i, cuw.uploadStateColumn);
-							if (uploadState.equals(cuw.finishedUploadState)
+							if (uploadState.equals(cuw.checkOKUploadState)
```
LaurentPV, line 82: must add the finishedUploadState also to condition in case old imports still stored from before the setting of the check.on.server property.

shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/CurrentNominativeDataController.java
```diff
@@ -96,20 +96,20 @@ public void mouseClicked(MouseEvent e) {
...
-							e1.printStackTrace();
+						} catch (IOException eIO) {
+							logger.error(eIO.getMessage(), eIO);
 						}
 					}
 				// delete one import: ready (to gain disk space) or finished
 				} else if (col == cuw.deleteColumn && row != -1) {
 					String uploadState = (String) cuw.table.getModel().getValueAt(row, cuw.uploadStateColumn);
-					if (uploadState.equals(cuw.finishedUploadState)
+					if (uploadState.equals(cuw.checkOKUploadState)
```
LaurentPV, line 106: Same comment as before, people may switch from setting the check.on.server on and off so we need to delete also FINISHED folders even if they do not contain dicom files anymore.

LaurentPV: Also, the delete button do not work anymore for uploads with status check_KO, is it the expected behavior ? I would think that the user can delete whatever local upload he/she wants

LaurentPV (changes requested):
Tested, consistency checker works fine but minor issues on the currentUploadsPanel for the deletion of uploads. Must be approved with the expected behavior.

shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/CurrentNominativeDataController.java
michaelkain: Hi @LaurentPV, well seen. I changed accordingly. With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/CurrentNominativeDataController.java
michaelkain: Well seen, changed accordingly. Cheers, Michael

Issue #2647, opened 2025-02-13
Index 2 out of bounds for length 2


michaelkain:
Rouen

2025-02-13_14:49:07.711 INFO  [] o.s.n.i.service.ImporterService - createAllDatasetAcquisition: org.shanoir.ng.importer.service.ImporterService@5603b5ac instances: 3
2025-02-13_14:49:07.753 INFO  [] o.s.n.i.s.d.MrDatasetAcquisitionStrategy - Generating DatasetAcquisition for   : null - AX - Rank:0
2025-02-13_14:49:07.770 ERROR [] o.s.n.i.service.ImporterService - Error during import for exam: 78379 : {}
java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at org.shanoir.ng.importer.strategies.dataset.MrDatasetStrategy.generateSingleDataset(MrDatasetStrategy.java:204)
	at org.shanoir.ng.importer.strategies.dataset.MrDatasetStrategy.generateDatasetsForSerie(MrDatasetStrategy.java:93)
	at org.shanoir.ng.importer.strategies.datasetacquisition.MrDatasetAcquisitionStrategy.generateDatasetAcquisitionForSerie(MrDatasetAcquisitionStrategy.java:95)
	at org.shanoir.ng.importer.service.DatasetAcquisitionContext.generateDatasetAcquisitionForSerie(DatasetAcquisitionContext.java:74)
	at org.shanoir.ng.importer.service.ImporterService.createDatasetAcquisitionForSerie(ImporterService.java:375)
	at org.shanoir.ng.importer.service.ImporterService.generateAcquisitions(ImporterService.java:273)
	at org.shanoir.ng.importer.service.ImporterService.createAllDatasetAcquisition(ImporterService.java:161)
	at org.shanoir.ng.datasetacquisition.controler.DatasetAcquisitionApiController.createAllDatasetAcquisitions(DatasetAcquisitionApiController.java:150)
	at org.shanoir.ng.datasetacquisition.controler.DatasetAcquisitionApiController.createNewDatasetAcquisition(DatasetAcquisitionApiController.java:135)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:351)
[...]

PR #2804, opened 2025-04-24
Fix for issue #2647

michaelkain:


michaelkain:
Successfully tested locally with Enhanced MR Image and regression test made with standard MR Image, all good.

Issue #2760, opened 2025-03-27
Tree display although it should not


DuckflipXYZ:
Reproductible case (enough data in ofsep)  :

Open an exam in Medical data -> examination -> click on one random exam.
Then open a pathology model in Preclinical data -> Pathology model -> click on one random pathology model.

The tree relative to the study of the last opened exam should be opened. Refresh the webpage is enough for clearing it.

PR #2810, opened 2025-04-25
don't show tree if not implemented

julien-louis:
https://github.com/fli-iam/shanoir-ng/issues/2760

don't show tree for entities that are not implemented as nodes in the tree : pathologies, therapies, ...

* queueMicrotask is an optimization I just used in a previous pr, I put it here as well so it will be easier to merge
* some code is duplicated inside and outside queueMicrotask because one is for the first component load but when you navigate to another entity of the same subclass the component is not recreated thus the constructor is not called again, that why there is a router subcription


DuckflipXYZ (approved):
Seems all good

PR #2812, opened 2025-04-28
Light checkstyle

julien-louis:
This introduce checkstyle in github actions for the shanoir java code.

In this pr, almost every rule has been commented in checkstyle.xml and no java file has been cleaned yet.


pierrehenri-dauvergne (approved):
Tested on neurinfo-qualif and it deployed OK

PR #2813, opened 2025-04-28
Change the proxy access url test  depending on profile

LaurentPV:


shanoir-uploader/src/main/java/org/shanoir/uploader/service/rest/ServiceConfiguration.java
```diff
@@ -89,12 +93,24 @@ public Boolean isProxyEnabled() {
...
 	public String getTestURL() {
-		if (ShUpConfig.proxyProperties.getProperty("proxy.test.url") != null
-				&& !ShUpConfig.proxyProperties.getProperty("proxy.test.url").equals("")) {
-			return (ShUpConfig.proxyProperties.getProperty("proxy.test.url"));
-		} else {
-			return "https://shanoir.irisa.fr";
-		}
-	}
-
+		String profile = ShUpConfig.basicProperties.getProperty("profile");
```
michaelkain, line 96: Hi @LaurentPV, thank you for your PR.
What happens in case no profile is pre-configured in basicProperties? what is true in most of the cases.

shanoir-uploader/src/main/java/org/shanoir/uploader/ShUpConfig.java
```diff
@@ -84,6 +84,16 @@ public class ShUpConfig {
 
 	public static final String ANONYMIZATION_PROFILE = "anonymization.profile";
 
+	public static final String NEURINFO_SERVER = "https://shanoir.irisa.fr";
```
michaelkain, line 87: Hi, we should get these URL from the profile.properties?

michaelkain:
Hi, sorry, maybe my issue was not correctly stated.
I think, we should ask for the profile selection, before we check the proxy and then use the shanoir server URL to test the proxy?

LaurentPV:
> Hi, sorry, maybe my issue was not correctly stated. I think, we should ask for the profile selection, before we check the proxy and then use the shanoir server URL to test the proxy? We should in InitialStartupState switch this context.setState(proxyConfigurationState); to profileSelection and then go next to proxyConfigurationState

Done in last commit.

michaelkain (approved):

Issue #2742, opened 2025-03-25
ShUp (QueryPACSService + SeriesNumberAndDescriptionSorter): use acquisitionTime


michaelkain:
Hi @LaurentPV,
I hope you are doing fine. Thank you for this good proposal.
We might have to test it with different PACS vendors, but I think the idea is good.
We could extend the querySeries code for the acquisitionTime to ask, and in case
present use it, if seriesNumber is missing, before sorting by seriesDescription.
In case the acquisitionTime is returned by Q/R, of course.
With kind regards, Michael

michaelkain:
currently testing at Neurinfo

PR #2818, opened 2025-04-29
Shup query retrieve use acquisition time

michaelkain:
Implementation for Github issue #2742


michaelkain:
It might be interesting to review and test this PR please.
I was not able to fully test it, as it did never receive an acquisitionTime to really validate it.
Might be worth testing with a debugger and assume, that no seriesNumber is returned, or something like this.
With kind regards, Michael

michaelkain:
and I am uncertain for the seriesDesc order, at least at Neurinfo, it did not work

PR #2819, opened 2025-04-29
allow filtering of studies with first letters of the name

LaurentPV:
Filtering is slower than subject name filtering because we query shanoir server for study cards and subjects for the study that matches the filter.


michaelkain:
Hi @LaurentPV, thank you for this PR. Are you sure, this PR is complete?
Maybe I am wrong, but does the decorate call multiple times the server?
With kind regards, Michael

LaurentPV:
> Hi @LaurentPV, thank you for this PR. Are you sure, this PR is complete? Maybe I am wrong, but does the decorate call multiple times the server? With kind regards, Michael

I will try to improve the behavior but it is normal that it makes a new call to getSubjects() when typing a study name because if it matches an existing study name the study is directly selected in the combo box.

michaelkain (approved):

PR #2820, opened 2025-04-30
shanoir-issue#2809: warning popup before remove center from study

pierrehenri-dauvergne:
How to test:
Have a study with multiple center and multiple study cards (with equipments from each center of the study)
Open the study > go to the center tab > edit > remove a center

If the center is linked to a study, a warning pops up and asks the user if he/she wants to complete the removal of the center


pierrehenri-dauvergne:
We can chose to delete studycards of center removed of a study but then some dataset_acquisition are linked to that study_card (why?) and the deletion of studycard is cancelled.
Should we just warn the user that the center can't be removed because it's linked to a study_card?

PR #2824, opened 2025-05-06
Laurent bugfix shup mass import no xml

LaurentPV:
This PR fixes the use of mass import from an excel file without any xml upload files.

The wait mechanism for the copyOrDownload thread to finish was switched from a while loop to a .join() method.


shanoir-ng-import/src/main/java/org/shanoir/ng/importer/dicom/query/QueryPACSService.java
```diff
@@ -522,14 +522,8 @@ private void querySeries(Association association, Study study, DicomParam modali
 	private void processDICOMSerie(Attributes serieAttr, Association association, Study study, DicomParam modality, List<Serie> series) {
 		Serie serie = new Serie(serieAttr);
 		if (!DicomSerieAndInstanceAnalyzer.checkSerieIsIgnored(serieAttr)) {
-			if (serie.getNumberOfSeriesRelatedInstances() > 0) {
```
michaelkain, line 525: Hi @LaurentPV, thank you for this PR.
I wonder about this removal here. Normally, every PACS returns numberOfSeriesRelatedInstances, normally.
Is there a specific reason to remove it? Maybe to show it to the user in the tree? as empty. With kind regards, Michael

michaelkain: answered in other PR

michaelkain (approved):

PR #2801, opened 2025-04-24
keycloak v26.1.5

julien-louis:
https://github.com/fli-iam/shanoir-ng/issues/2696

**front** : instead of importing the keycloak.js (served and host independently by nginx) in index.html, the angular shanoir projet now uses the npm dependency "keycloak-js".  

**back** : update dependencies
(keycloak-admin-client version has to be set manually since the tags no longer stick to the main updates, see https://www.keycloak.org/docs/latest/release_notes/index.html#dedicated-release-cycle-for-the-client-libraries)

**keycloak container** : update version

Just check that you can log to the front and access to data.

PR #2825, opened 2025-05-06
Papaya lazy

julien-louis:
Must wait for https://github.com/fli-iam/shanoir-ng/pull/2801 to be merged first (only commits starting with [9b8a0ab](https://github.com/fli-iam/shanoir-ng/pull/2825/commits/9b8a0abce90b0f94156d70ae0fc5778a46ede9ee) are to consider)

* fixes problems with papaya displaying wrong images (images from the previous selected) in clinical context
* does download needed data only when the user clicks "preview" (lazy loading)
    * (we now give to the papaya component a function telling it how to get the data, then the papaya component is responsible for running it only when needed, for the error message and for the downloading status)


pierrehenri-dauvergne (approved):
Tested locally and working OK

PR #2827, opened 2025-05-07
bugfix compatible study and study card not selected by default

LaurentPV:
The issue was due to a refactoring, we were setting compatibleStudyCard back to False in ImportDialogOpener class if the compatible study card was not the last in the study card list.


michaelkain (approved):
Hi @LaurentPV, thank you! Very good fix, great, that you found it.
I merge by code review, as code very clear. With kind regards, Michael

PR #2848, opened 2025-05-20
shanoir-issue#2844: remove the deletion of subject from study update

pierrehenri-dauvergne:
This is the first step of fixing the concurrent update of the same entity. Starting with studies and the removal of subject (because this could lead to loss of data).

How to test:
Open a study
Go to subject tab
Edit the study
Remove a subject
Save the study

The subject is still in the study.


pierrehenri-dauvergne:
Tested locally and now we can't remove a subject from a study (from study > subject tab > edit)

PR #2851, opened 2025-05-22
shanoir-issue#2844-2: prevent users from removing a subject from study edit

pierrehenri-dauvergne:
How to test:

Open a subject > edit > remove a study > save

Now, users can't remove a subject from study > subject tab > edit.


pierrehenri-dauvergne:
Tested locally and on ofsep-qualif. Working OK, just need to refresh the tree once the subject is deleted but it's not a big issue.

PR #2852, opened 2025-05-23
Security optim

julien-louis:


pierrehenri-dauvergne:
Hi @julien-louis 
Lots of testing to do, maybe this could be checked in a test session.
I've made some tests on ofsep-qualif and I haven't seen big issues
The only thing is that for coils, acquisition-center and acquisition-equipment: as an expert, I can see some that I shouldn't and when I try to open on, it throws an "access denied" error.

These 3 entities are the only one where I saw an issue but maybe there are others.
Mass download was fine, as well as opening every other entities and importing dicom

PR #2858, opened 2025-05-28
develop2master

michaelkain:


michaelkain:
Hi @alexpron, I hope you are doing fine.
You can ignore all that is on shanoir-uploader, as we release the server only.
Cheers, Michael

alexpron (changes requested):
- Regression found by @pierrehenri-dauvergne during pre release test session:
when importing DICOM data and creating new subject, subject information are not pre filled as they are currently in production. 
- keycloak test ok 
- subject deletion from study ok

Issue #2855, opened 2025-05-26
Loading datasets takes too much time


pierrehenri-dauvergne:
For some datasets like 209024 (on ofsep), opening its details take too much time and ms-datasets logs end up throwing 'Broken pipe' errors.

This dataset has been processed, I don't know if it's related or not. It is also part of a very big study.

PR #2868, opened 2025-06-05
Performance optimizations : nb of queries in Datasets

julien-louis:
fixes https://github.com/fli-iam/shanoir-ng/issues/2855

!!! Must test VIP as this PR changes the amount of data returned when querying the server for a dataset or a processing

This PR can significantly reduce (75%) the amount of queries needed to display one dataset and subsequently other entities.

(note : accessing a LAZY n list property triggers n or more db requests while the EAGER strategy fetches everything once) 

* removed DatasetDTO.StudyTagDTO[ ].StudyDTO (huge gain)
* removed ExaminationDatasetAcquisitionDTO.DatasetWithProcessingsDTOs[ ]
* removed useless LAZY loadings in `public List<ExaminationDatasetAcquisitionDTO> datasetAcquisitionsToExaminationDatasetAcquisitionDTOs(
final List<DatasetAcquisition> datasetAcquisitions)`
* replaced `repository.findByExaminationId(examinationId)` by `examRepository.findByIdWithEagerAcquisitions(examinationId)`
    * find by dependency id creates n query
    * getting the parent and force the wanted dependencies to fetch eagerly makes 1 query 
* removed acquisitions from SubjectExaminationDTO
    * [GET] examinations/subject/{subjectId}/study/{studyId}
* removed complex objects from ProcessingDTO, just use ids instead.

### DIFF
--- a/shanoir-ng-anonymization/pom.xml
+++ b/shanoir-ng-anonymization/pom.xml
@@ -24,7 +24,7 @@
 		<maven.compiler.target>17</maven.compiler.target>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<dcm4che.version>5.30.0</dcm4che.version>
-		<poi.version>4.1.2</poi.version>
+		<poi.version>5.4.0</poi.version>
 	</properties>
 
 	<dependencies>
--- a/shanoir-ng-anonymization/src/main/java/org/shanoir/anonymization/anonymization/AnonymizationServiceImpl.java
+++ b/shanoir-ng-anonymization/src/main/java/org/shanoir/anonymization/anonymization/AnonymizationServiceImpl.java
@@ -215,17 +215,17 @@ public class AnonymizationServiceImpl implements AnonymizationService {
 					anonymizeTag(tagInt, action, datasetAttributes);
 				// even: public tags
 				} else if (anonymizationMap.containsKey(tagString)) {
-                                    switch (tagInt) {
-                                        case Tag.SOPInstanceUID -> anonymizeSOPInstanceUID(tagInt, datasetAttributes, mediaStorageSOPInstanceUIDGenerated);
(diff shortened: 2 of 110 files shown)


shanoir-ng-front/src/app/dataset-acquisitions/tree/dataset-acquisition-node.component.ts
```diff
@@ -71,8 +71,9 @@ export class DatasetAcquisitionNodeComponent extends TreeNodeAbstractComponent<D
                     this.input.studyRights.includes(StudyUserRight.CAN_ADMINISTRATE),
                     this.input.studyRights.includes(StudyUserRight.CAN_DOWNLOAD)
                 );
-                this.loadDatasets();
             }
+            this.node.registerOpenPromise(this.contentLoaded);
```
julien-louis, line 75: this is in order to make the tree work well with the fact that the acquisitions are no loaded in a separate time

julien-louis:
Now it works !!!

DuckflipXYZ (commented):
Performance comparison :

- Solr click / url search : hard improvement, for object display as for tree display (improvement factor between 2 and 3 upon various test).
- Navigation through tree : hard improvement, even if the order of magnitude of the loading values is lower than the one for line above (we're talking about secs, not dozen of secs).
- Processing data display : it indeed has to be loaded manually and afterwards by clicking specifically on the processing to display, but the loading time is really short, and clearly negligible comparing to the time saved at dataset loading. Moreover, it's kinda rare that a user wants to see the processing data, so the majority of the user won't see that point.


Feature checked : 
- Data deletion at every level under subject (subject included) : no problem
- DICOM import : no problem
- VIP exec : working well, a develop rebase had to be done for fixing a little issue, but there is nothing else

Didn't see any issue for now, really nice job bro =)

michaelkain (approved):
Hi, I approve after test session, that went well.
With kind regards, Michael

PR #2869, opened 2025-06-05
Shanoir issue#2861: fix import form

pierrehenri-dauvergne:
How to test:

Start import
Create new subject
The create-subject-form is pre-filled with datas from the DICOM uploaded

Also test: create examination, preclinical subject and acquisition equipment

Maybe it's not related but after filling the create-subject-form, I can't save the new subject, the "save" button is disabled.

### DIFF
--- a/shanoir-ng-front/src/app/import/clinical-context/clinical-context.abstract.component.ts
+++ b/shanoir-ng-front/src/app/import/clinical-context/clinical-context.abstract.component.ts
@@ -390,7 +390,7 @@ export abstract class AbstractClinicalContextComponent implements OnDestroy, OnI
 
         this.studycard = this.center = this.acquisitionEquipment = this.subject = this.examination = null;
         let studycardsOrCentersPromise: Promise<void>;
-        if (this.useStudyCard && this.modality == 'MR') {
+        if (this.useStudyCard && (this.modality == 'MR' || this.modality == 'bruker')) {
             studycardsOrCentersPromise = this.getStudyCardOptions(this.study).then(options => {
                 this.studycardOptions = options;
                 return this.selectDefaultStudyCard(options);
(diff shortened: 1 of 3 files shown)


pierrehenri-dauvergne:
By changing the way we send data to the create-subject-form from the import page, we added asynchronicity in the process.
Basically, we now use breadcrumbs as a Map<key, value> but it's async. So, when we open the create-subject-form, this is what happens:

Click the "create subject" button: 
- clinical-context.openCreateSubject: route.navigate + fillCreateSubjectStep

- route.navigate opens the subject.component, so it calls the init which calls super.init() (entity.abstract.component) which calls buildForm (in subject.component).

- Then, once the subject.component.init is done (from the route.navigate), the fillCreateSubjectStep is called.

It's hard to follow what's really going on because we use clinical-context, basic-clinical-context, entity.abstract, subject and breadcrumb.service.

-----

With theese lines:
```
        this.breadcrumbsService.currentStep.addPrefilled("entity", this.getPrefilledSubject());
        this.breadcrumbsService.currentStep.addPrefilled("firstName", this.computeNameFromDicomTag(this.patient.patientName)[1]);
        this.breadcrumbsService.currentStep.addPrefilled("lastName", this.computeNameFromDicomTag(this.patient.patientName)[2]);
```

we can have the subject-form filled with firstName and lastName, so I cheated a bit and added the same for subjectStudyList and birthDate (which should be in this.getPrefilledSubject() but I don't know why, it's not).


In the current state of things, this is what the subject-form is looking like:
image

If I fill the "name" field, I can save the subject. BUT the "is already anonymised" is not set. Its default value is false, I don't know why it's unset now.

DuckflipXYZ (commented):
Hey ! 


Various point : 

- There are lots of console.log remaining, I'm not sure that it's intentionnal
- For the isAlreadyAnonymized var, there is no way to add a setter in the init() method ? The value is probably lost while breadcrumbs mechanics are occuring
- I can not create examination when importing DICOM, either the subject is new or not.

Preclinical (it might deserves its own issue, edit : it's [here](https://github.com/fli-iam/shanoir-ng/issues/2884) : 
- I'm not sure about the impact, but when selecting the acquisitions I want when importing Brucker, contrary to DICOM, examination is not selected. I don't know if it's only visual or if it affects the import processing

PR #2872, opened 2025-06-11
ShUp embedded keystore in sslContext for shanoir.irisa.fr and shanoir-ofsep.irisa.fr

LaurentPV:
In httpService, if profile is not dev, we create a certs directory in .su folder and download the certificates in it to load them in the sslContext. 

The validity dates of each certificate is checked at launch and those are downloaded again if validity check fails.


michaelkain:
After our discussion, I allow myself to close the PR related to a security issue uncertainty.
Sorry for that and with kind regards, Michael

Issue #2837, opened 2025-05-15
ShUp: clean .su (old versions)


michaelkain:
Hi @LaurentPV,
today ShUp opens and migrates all old properties to the current version 9.0.0.
We should do this once and then delete all old properties, when a new version
is installed. This should align with our new version numbers, we discussed already.
Same topic here: https://github.com/fli-iam/shanoir-support/issues/21
With kind regards, Michael

LaurentPV:
Closed with https://github.com/fli-iam/shanoir-ng/pull/2873

PR #2873, opened 2025-06-12
Laurent shup jpackage

LaurentPV:
This PR is an attempt at automatizing the release process of ShUp and solving the java version issue for users.

A new GitHub action is created with the shup_release.yml file that triggers a packaging of ShUp executables for Linux, Windows and Mac OS with an associated Release and version tag.

- the GitHub release action is triggered when pushing on master branch a commit on shanoir-uploader folder content.
- existing release tags for Shanoir_NG are following the NG_x.x.x pattern, we retrieve the last created tag from GitHub repo and get the version for the ShanoirUploader tag from this Shanoir_NG tag. The ShanoirUploader tag follows the SHUP_x.x.x pattern.
- Then this version is injected into the pom.xml of shanoir-uploader folder before running the maven build.
- We create a custom java runtime with JLink stored in shanoir-uploader/target/custom-runtime
- We package the executables (.exe, .deb, .dmg) using Package targeting the JLink custom-runtime and the jar with dependencies, setting the java app.version parameter with the new tag version for logback to store the su.log file in the proper folder.
- We Release a new version of ShanoirUploader with a version synchronized with the Shanoir_NG version containing the archive used previously for running the app (.jar with .sh and .bat scripts), a .exe executable, a .deb for Linux and a .dmg for Mac OS.
- New vectorized icons were added
- As logback configuration is now getting the version from java parameter, when running shanoirUploader in debug we create by default a .su_dev folder storing the configuration and logs.
- At launch we run through previous shup versions folders as we did before but we remove them after copying the properties if necessary, excluding the .su_dev folder if exists, allowing us to update shanoirUploader versions from 9.0.0 to 2.6.5 without any issues (linked with https://github.com/fli-iam/shanoir-ng/issues/2837).

TODO : 

- [ ] add INRIA certificates

### DIFF
--- a/shanoir-uploader/src/main/java/org/shanoir/uploader/gui/AboutWindow.java
+++ b/shanoir-uploader/src/main/java/org/shanoir/uploader/gui/AboutWindow.java
@@ -66,7 +66,7 @@ public class AboutWindow extends JFrame {
 		addItem(aboutPanel, nameLabel, 0, 2, 1, GridBagConstraints.CENTER);
 
 		JLabel versionLabel = new JLabel(ShUpConfig.SHANOIR_UPLOADER_VERSION + " Released: "
-				+ ShUpConfig.basicProperties.getProperty(ShUpConfig.RELEASE_DATE).substring(0, 9));
+				+ ShUpConfig.basicProperties.getProperty(ShUpConfig.RELEASE_DATE).substring(0, 10));
 		addItem(aboutPanel, versionLabel, 0, 4, 1, GridBagConstraints.CENTER);
 
 		JLabel copyrightLabel = new JLabel(
@@ -77,8 +77,7 @@ public class AboutWindow extends JFrame {
 				resourceBundle.getString("shanoir.uploader.helpMenu.aboutShUp.copyrightPseudonymus"));
 		addItem(aboutPanel, copyrightPseudonymusLabel, 0, 6, 1, GridBagConstraints.CENTER);
 
-		final JLabel SupportLabel = new JLabel(
-				resourceBundle.getString("shanoir.uploader.helpMenu.aboutShUp.supportMail"));
+		final JLabel SupportLabel = new JLabel(supportMail);
 		addItem(aboutPanel, SupportLabel, 0, 7, 1, GridBagConstraints.CENTER);
 
 		SupportLabel.addMouseListener(new MouseAdapter() {
@@ -86,7 +85,7 @@ public class AboutWindow extends JFrame {
 			public void mouseClicked(MouseEvent event) {
 				try {
 					Desktop.getDesktop().mail(new URI(
-							"mailto:" + resourceBundle.getString("shanoir.uploader.helpMenu.aboutShUp.supportMail")));
+							"mailto:" + supportMail));
 				} catch (IOException e) {
 					e.printStackTrace();
 				} catch (URISyntaxException e) {
@@ -114,7 +113,7 @@ public class AboutWindow extends JFrame {
 		// frame.setLocationRelativeTo( null );
 		Point center = GraphicsEnvironment.getLocalGraphicsEnvironment().getCenterPoint();
 		int windowWidth = 415;
-		int windowHeight = 300;
+		int windowHeight = 327;
 		// set position and size
 		frame.setBounds(center.x - windowWidth / 2, center.y - windowHeight / 2, windowWidth, windowHeight);
 
--- a/shanoir-uploader/pom.xml
+++ b/shanoir-uploader/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>org.shanoir.uploader</groupId>
 	<artifactId>shanoir-uploader</artifactId>
-	<version>2.6.5</version> <!-- Version to be modified when pushing a release, if not will be updated for the release only with last NG_ tag version-->
+	<version>2.6.5</version> <!-- Version to be modified when pushing a release, if not will be updated anyway for the release with last NG_ tag version-->
 	<packaging>jar</packaging>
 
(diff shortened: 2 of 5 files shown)


.github/workflows/shup_release.yml
```diff
@@ -0,0 +1,225 @@
+name: Release ShanoirUploader App
+
+on:
+  push:
+    branches:
```
michaelkain, line 5: Hi @LaurentPV, thank you for this great PR.
I thought, we do not react on push, but add something like a link or button, where we can trigger a release, independent from the push action? Anthony is using this "on: workflow_dispatch:" on shanoir-deploy.
I hope this sounds logic?

shanoir-uploader/src/main/java/org/shanoir/uploader/ShUpConfig.java
```diff
@@ -18,9 +18,9 @@ public class ShUpConfig {
 	/**
 	 * Constants
 	 */
-	public static final String SHANOIR_UPLOADER_VERSION = "v9.0.0";
+	public static final String SHANOIR_UPLOADER_VERSION;
 	
-	public static final String RELEASE_DATE = "2025-04-29";
+	public static final String RELEASE_DATE = "2025-06-27"; // Do we keep it ?
```
michaelkain, line 23: Hi @LaurentPV, I think we can remove this? We will know the release date via Github and the artifacts produced? With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/gui/AboutWindow.java
```diff
@@ -21,15 +21,15 @@
...
-import org.slf4j.LoggerFactory;
 import org.shanoir.uploader.ShUpConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class AboutWindow extends JFrame {
 
 	private static final Logger logger = LoggerFactory.getLogger(AboutWindow.class);
 
-	String supportMail = "imagerie@ofsep.org";
```
michaelkain, line 32: Hi @LaurentPV,
we need to keep this for the ofsep-version. We should trigger this with the profile selected by the user.
We might extend the profile for a support mail value and use it to clean this up.
With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/gui/AboutWindow.java
```diff
@@ -65,7 +65,7 @@ public AboutWindow(final ResourceBundle resourceBundle) {
 				+ resourceBundle.getString("shanoir.uploader.helpMenu.aboutShUp.name") + "</B></body></html>");
 		addItem(aboutPanel, nameLabel, 0, 2, 1, GridBagConstraints.CENTER);
 
-		JLabel versionLabel = new JLabel(ShUpConfig.SHANOIR_UPLOADER_VERSION + " Released: 2023-03");
+		JLabel versionLabel = new JLabel(ShUpConfig.SHANOIR_UPLOADER_VERSION + " Released: 2025-06");
```
michaelkain, line 68: Hi @LaurentPV, please remove the date here.

shanoir-uploader/start-shup-linux-mac-java.sh
```diff
@@ -1,12 +1,19 @@
+JAR_FILE=$(ls shanoir-uploader-*-jar-with-dependencies.jar 2>/dev/null | head -n 1)
```
michaelkain, line 1: Hi @LaurentPV,
I am not a big fan of this ls/for calls to detect the jar-version.
One idea, why do we not always use a .jar, with the same name: shanoir-uploader-with-dependencies.jar?
The version will be in the execs produced by jlink and on Github.
Actually, as longer I think about it, I think we do too much here.
We can even fix the maven pom.xml to version 1.0.0 and never touch it.
As long as we know the release on Github with the packages produced,
what does it help us, that maven has the same version? as long as we tag
the code and can go back to it, what do we gain, when maven aligns with it?
With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/ShUpConfig.java
```diff
@@ -139,4 +141,20 @@ public static boolean isModeSubjectStudyIdentifier() {
 		return Boolean.parseBoolean(profileProperties.getProperty(MODE_SUBJECT_STUDY_IDENTIFIER));
 	}
 
+	/**
+	 * Static initialization block to set the ShanoirUploader version.
+	 * Used to manage the version only in pom.xml.
+	 */
+	static {
```
michaelkain, line 148: Hi @LaurentPV,
sorry, that I am quite late with my simplification proposals.
Why does Github not just produce a version.json or something into the produced execs
and the code reads this file and displays the version, independent from Maven?
We do something similar on Shanoir server. This will simplify a lot, normally hopefully.
Actually on the server it is the same. We tag and release the code, but Maven remains
stable and in line with Spring Boot. We would not gain anything, as all "frozen" in our
delivered docker containers (similar to your mac-linux-windows exec packages),
if Maven would align with the release version.
With kind regards, Michael

LaurentPV: Ok, I couldn't find the process editing this version.json in the Shanoir NG release, is it a file added to the jar ? then I would need the version in the pom.xml. In my latest push I kept the release.date in basic.properties (couldn't find a better place) with a value injected from the pom.xml again. So it can be displayed properly in the About window for the user.

.github/workflows/shup_release.yml
michaelkain: perfect

Issue #2882, opened 2025-06-19
Some issues relative to Solr


DuckflipXYZ:
- Turn date filter in left panels from excluded to included (maybe annotate with an info dialog)
- Refracto downloaded zip file name.... 
- Annotate the DICOM/NIFTI option in download pop up with 'only for dicoms datasets'
- Propose granularity for ordering when downloading (currently, only exam ordering). Cumulative ordering scale choice  would be nice
- At download, when choosing nifti format, turn the converter field mandatory

PR #2894, opened 2025-06-25
#2882 : date filters for solr from excluded to included + tiny update for mass-download pop-up

DuckflipXYZ:
Tested in local, not on qualif.

Check that both firefox and not firefox downloads from solr work correctly + Check that date filters (in left panels) for solr are working well.


pierrehenri-dauvergne (changes requested):
When I test locally, I've examination with date = 04/03/2010 and when I filter the exam dates from 01/03/2015 to 31/03/2010, some other exam appear in the list (15/11/2010 shouldn't):
image

Also, you changed the html error from dd/mm to mm/dd but not the rest of the html. Which format should we aim for?

PR #2898, opened 2025-07-03
ShUp - add a tooltip icon to inform users

LaurentPV:
Add an "info" icon to inform user that a tooltip explains the field in the query form.
Capture d’écran 2025-07-03 à 16 36 44


michaelkain (approved):
Locally tested and it works.

PR #2909, opened 2025-07-15
Shup no jboss seam repository

michaelkain:


michaelkain:
Hi, I tested locally with an old version (v8.0.0) and compared the hashs for OFSEP-Qualif profile
and all hash remain equal after the removal of the lib and deletion of the old StreamGobbler.
I tested an error case as well with a wrong input and the error display and logs are the same too.
I merge to get rid of the Github build issue with this very legacy repository. With kind regards, Michael

PR #2922, opened 2025-07-21
Single-study subject: backend

michaelkain:
This PR prepares the full implementation of single-study subjects:
It creates new subjects into subject, in case a subject is part of multiple studies and
then aligns the subject_study objects to point to the new subjects. In the end only
subject study rows exist, that point to exactly one subject.
It adds all fields of subject_study to subject and copies the values on the database.
In case a new subject is created, the new values are filled similar to subjectStudy.
Removes group of subjects.
Blocks adding more than 1 study to a subject, after the migration script run.

Tests made (by order of importance):
- ShUp: DICOM import (ofsep-profile): create new subject (auto-create): **OK** (tested) [method createAutoIncrement]
- ShUp: DICOM import (neurinfo-profile): create new subject: **OK** (tested) [method create, same as WebGUI]
- ShUp: DICOM import: ofsep-profile: use existing subject and add to another study: not working anymore, as forbidden by backend **OK** (tested, error message displayed) [method update]
- WebGUI: DICOM import, create new subject: **OK** (similar values in subject_study and new subject) [method create]
all above tested and values compared on database level (pseudo hash values included)
- WebGUI: Manage subjects, create new subject: **OK** (similar values in subject_study and new subject)
- WebGUI: Manage subjects, create new subject with 2 studies: not working anymore, as forbidden by backend **OK** (tested, error message displayed) [method update]
- WebGUI: Edit existing subject and add to new 2. study: not working anymore, as forbidden by backend **OK** (tested, error message displayed) [method update]


DuckflipXYZ:
to fix: 

```json
{"code":500,"message":"A collection with orphan deletion was no longer referenced by the owning entity instance: org.shanoir.ng.subject.model.Subject.tags","details":null}
```

open a subject form, change something and click on update

michaelkain:
Hi, my summary after a lot of testing:
- today, on qualif and prod, when I edit an existing subject and add a new subject tag, it does not work (200, but no tag added); this PR fixes this and allows to add any number of tags to subjects during subject edit
- today, on qualif and prod, when I edit a subject, that has a tag (e.g. added during the web GUI import, not possible with ShUp) and try to remove the tag, it does not work (200, but no tag removed); my PR does not fix this, I tried, but as in any case SubjectStudyTag will disappear and it works with SubjectTag (I tested), I stop here for my PR
- today, on qualif and prod, deleting a subject with a tag does not work, same here for my PR (it will work in the future with SubjectTag, but there is a problem with SubjectStudyTag)

michaelkain:
- Open a subject form, edit it, remove its only study, and try to update: throws error, this is expected because we do not want to create orphan subjects (in no study)

PR #2929, opened 2025-07-22
Vip features to merge

DuckflipXYZ:
Currently in test on qualif : 

Hi ! Here are the 2 commits to merge into master for being able to process both SIMS and PMAP pipelines :
- 2890 : The longer commit between both, but the more important. Relative to that [not merged PR](https://github.com/fli-iam/shanoir-ng/pull/2923) and that [merged PR](https://github.com/fli-iam/shanoir-ng/pull/2891)
- 2904 :  The shorter commit between both, and the less important. Relative to that [not merged PR](https://github.com/fli-iam/shanoir-ng/pull/2905)

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
@@ -68,6 +68,8 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.shanoir.ng.dataset.dto.DatasetLight;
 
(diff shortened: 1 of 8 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -394,6 +396,10 @@ public List<Dataset> findByExaminationId(Long examinationId) {
 		return Utils.toList(repository.findByDatasetAcquisitionExaminationId(examinationId));
 	}
 
+	public List<Dataset> findDatasetAndOutputByExaminationId(Long examinationId) {
+        return StreamSupport.stream(repository.findAllById(repository.findDatasetAndOutputByExaminationId(examinationId)).spliterator(), false).toList();
```
michaelkain, line 400: Hi @DuckflipXYZ, thank you for your PR.
This looks like a very database intensive pile of code and I am not sure I understand, what it does.
With kind regards, Michael

michaelkain (commented):
Hi @DuckflipXYZ, thank you for this PR, it looks promising.
On the other hand I am not sure, if I can take it into production, without a download test session with Chrome and Firefox. External scripts depend completely on our download output and we need to be careful here. With kind regards, Michael

PR #2935, opened 2025-07-25
Laurent shup mass import multiple exams

LaurentPV:


michaelkain:
Hi @LaurentPV,
I am sorry, but I think we might close this PR entirely, sorry.
- You use the studyInstanceUID from the hospital, that disappears entirely.
- The studyInstanceUID is replaced by a random-shanoir-UID during pseudonymization (code in module anonymization)
- We should use the already existing studyInstanceUID coming from the server on examination, see my ECC PR
https://github.com/fli-iam/shanoir-ng/pull/2125
- You can not use the dicomWeb interface with a studyInstanceUID, the dicomWeb interface of Shanoir only knows examinationUIDs, if not you can not check the rights and it will not work.
With kind regards and sorry for the confusion,
Michael

PR #2941, opened 2025-07-29
Adapt subject name auto-create to 4 digit centerIds

michaelkain:
- Can be tested only with ShUp and profile ofsep (or switch flags in dev-profile accordingly)
- Tested locally with ShUp (only usage of autoCreate method)
- Tested with 3 digit center code OK
- Tested with 4 digit center code OK


LaurentPV:
Tested and approved on 31st of July by Celine on Ofsep qualif.

PR #2945, opened 2025-07-30
shanoir-issue#2936: single-study subject frontend

pierrehenri-dauvergne:
Before:
image

After:
image

Creating a subject from the subject-list > new form should work
Creating a subject from the import > clinical-context > new subject should also work (for both DICOM and EEG)


michaelkain:
Hi @pierrehenri-dauvergne,
my local tests are very promising and I propose to approve and merge it?
I only found 2 minor bugs:
1) Edit study, subjects tab, add new **subject tags** and assign directly (no click on update) to a subject below: KO
already existing
TO IGNORE: as we have to refactor this subject-study table anyway to new subject structure
2) Tag removal after edit subject, did not work, keeps the 3 tags and does not remove one: KO
It might be interesting to have only a look on 2), maybe the problem is in the backend.
Please find after my test protocol.

michaelkain (approved):
All local tests: OK for major use cases

Issue #2949, opened 2025-08-04
Date standardization


DuckflipXYZ:
CARE /!\ The text below is with an english browser language. As we tried it, angular is following the browser langage, and some parts of the date code is following, but not every code instruction (for example, error labels are written in raw, so they are not following the browser langage) : 

\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\

Date format is not unique across the different date fields in Shanoir. For example, all dates in Solr page are ruled by the MDY format : 

<img width="1430" height="752" alt="Image" src="https://github.com/user-attachments/assets/f24bc6e4-21a2-4c2c-98cc-f496822e7817" />

But in an other hand, dates from exam forms are not regular : 

- DMY in view mode 

<img width="645" height="466" alt="Image" src="https://github.com/user-attachments/assets/0c49d26d-076c-4413-bc57-ba884028b5ae" />

- MDY in edit mode 

<img width="681" height="517" alt="Image" src="https://github.com/user-attachments/assets/13ac7dc6-fb73-4dda-a2dc-9c6cf58a1868" />

- DMY in error label 

<img width="681" height="517" alt="Image" src="https://github.com/user-attachments/assets/84a8cf9b-6702-461c-86b4-64c31c3daded" />

According to https://en.wikipedia.org/wiki/List_of_date_formats_by_country, DMY format seems relevant as a great part of the world is using it. Discussion needed.

PR #2950, opened 2025-08-05
#2949 : date according to browser language refracto

DuckflipXYZ:
Tested in local, not in remote

To test it, check date fields in view mode and edit mode, and change your browser langage (it might need to restart). Pay attention to Solr filtr

Dev logic is located in shanoir-ng-front/src/app/shared/localLanguage/


julien-louis (changes requested):
Hi, instead of setting a property in any component that uses the pipe to access the static import, you should set a custom date pipe, then just call {{dataset.startDate | datePipeOrSmth}} without giving any format here, since it's always the same.

When there is time it could be either with a parameter or another pipe 

```
import { Pipe, PipeTransform } from '@angular/core';
import { DatePipe } from '@angular/common';

@Pipe({
  name: 'datePipeOrSmth'
})
export class DatePipeOrSmth implements PipeTransform {
```

Tell me if you need help or if you don't agree, no problem.

DuckflipXYZ:
I did not create a pipe for the localDateDisplay, as there is no data to transform. Tell me what u think of the impl :)

shanoir-ng-front/src/app/shared/localLanguage/localDateFormat.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Pipe({
+    name: 'localDateFormatPipe',
+})
+
+export class LocalDateFormatPipe implements PipeTransform{
+
+    private datePipe = new DatePipe(BROWSER_LANGUAGE);
+
+    transform(date: Date, time: Boolean = false) {
```
julien-louis, line 13: not a big deal but it is better to use the primitive 'boolean' instead of the class 'Boolean'

julien-louis (changes requested):
It works with firefox but sorry I have an error in the debug console with chrome : 

```
InvalidPipeArgument: 'NG0701: Missing locale data for the locale "fr".' for pipe 'DatePipe'
    at invalidPipeArgumentError (common_module-BukMeiI7.mjs:3503:12)
    at DatePipe.transform (common_module-BukMeiI7.mjs:4006:19)
    at LocalDateFormatPipe.transform (localDateFormat.pipe.ts:19:34)
```

when displaying a study's details

DuckflipXYZ:
I changed Boolean class to boolean type, and modified the code for chrome, it should be fine now

julien-louis (approved):

Issue #2932, opened 2025-07-25
Subject type filter for Solr migth be broken


DuckflipXYZ:
<img width="506" height="418" alt="Image" src="https://github.com/user-attachments/assets/6e0e39db-8adf-4caa-9141-22a8a945285c" />

As u can see below, the number of elements in the Solr tab is not the same as the total of elements splitted between subject type, the filter migth have lost some them.

PR #2953, opened 2025-08-06
#2932 : adding empty filter option to solr

DuckflipXYZ:
Tested in local, not in remote.

To test it, check that empty filter option is available only when here is a number next to it (meaning that there are some datasets corresponding to that empty filter option). Check that the numbers correspond to the number of returned elements in the tab.


shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/solrj/SolrJWrapperImpl.java
```diff
@@ -357,8 +425,50 @@ private SolrResultPage<ShanoirSolrDocument> getSearchResultsWithFacets(ShanoirSo
...
 
+		LOG.info("test");
+		Map<String, FacetPageable> cleanedFacetPage = cleanFacetPage(response, shanoirQuery);
+
 		/* build the page object */
-		return buildShanoirSolrPage(response, pageable, shanoirQuery.getFacetPaging());
+		return buildShanoirSolrPage(response, pageable, cleanedFacetPage);
+	}
+
+	private Map<String, FacetPageable> cleanFacetPage(QueryResponse response, ShanoirSolrQuery shanoirQuery) {
```
julien-louis, line 435: Can you explain why is that for please ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/solr/solrj/SolrJWrapperImpl.java
```diff
@@ -357,8 +425,50 @@ private SolrResultPage<ShanoirSolrDocument> getSearchResultsWithFacets(ShanoirSo
 
 		QueryResponse response = querySolrServer(query);
 
+		LOG.info("test");
```
julien-louis, line 428: test log to remove

julien-louis (changes requested):

DuckflipXYZ:
Log removed. 

The cleanFacetPage method is relative to that line : 

					query.set("f." + facetName + ".facet.missing", true);

That line always allows empty value as solr filter option, whatever the count (so even if the dataset count for that empty value is 0). The cleanFacetMethod get that element count for empty values afterwards (once the solrQuery is done), and remove the empty filter option if no dataset is corresponding to it.

DuckflipXYZ:
Renamed the empty filter option as 'None' and removed the changes for dateRange and numericRange as examiantionDate and magneticField

Issue #2634, opened 2025-02-10
VIP execution tracking


DuckflipXYZ:
The idea is to have a vip exec tracking file constructed while executions are launched (meaning not constructed afterwards)

Some points : 
- .csv
- Referencing all exams, even those which can not be sent to VIP, with the explanation
- Exhaustive list of columns : exec conditions / sending / results.yaml / error.yaml (the 2 lasts columns are specific to CometeMoelle)

DuckflipXYZ:
After first real session, some changes needed : 
- add subject column

some other issues : 
- tracking file removed when new released

PR #2954, opened 2025-08-06
#2634 : tracking file modified for being more generic

DuckflipXYZ:
Tested in qualif.

To test it, just run an exec and check that the tracking file is correctly filled. It's located in data/ofsep/qualif/datasets-data/vip-data/{TrackingFile} (There is a pyNoir API to download them too). Particularly, the 2 last columns have change from SIMS specific columns to generic status and output file names columns.


LaurentPV (dismissed):
tested, works.

DuckflipXYZ:
It's currently not ? It should be something like Failed from what I remember

LaurentPV (approved):
Tested OK, tracking file as expected.

Issue #2836, opened 2025-05-15
values of column "Studies" in export CSV for user list missing


chomo-ofsep:
when exporting as CSV from page https://shanoir-ofsep.irisa.fr/shanoir-ng/user/list, the column "Studies" is empty for all the ligns even though on the front there are values.

PR #2960, opened 2025-08-11
wait all async fields in csv export

julien-louis:
fix https://github.com/fli-iam/shanoir-ng/issues/2836#issuecomment-3163930320
(= study name now appears in user list csv export)

also fixes study list table export


julien-louis:
I created a new issue : https://github.com/fli-iam/shanoir-ng/issues/3033

Issue #2916, opened 2025-07-17
Preclinical: create subject/examination issue


michaelkain:
Caused by: org.shanoir.ng.shared.exception.ShanoirException: Failed to create subject. Animal subject can't be created.
	at org.shanoir.ng.preclinical.subjects.controller.AnimalSubjectApiController.createSubject(AnimalSubjectApiController.java:132)
	at org.shanoir.ng.preclinical.subjects.controller.AnimalSubjectApiController.createAnimalSubject(AnimalSubjectApiController.java:91)
	... 101 common frames omitted
Caused by: org.shanoir.ng.shared.exception.ShanoirException: Created subject id is null.
	at org.shanoir.ng.preclinical.subjects.service.AnimalSubjectServiceImpl.createSubject(AnimalSubjectServiceImpl.java:130)
	at org.shanoir.ng.preclinical.subjects.controller.AnimalSubjectApiController.createSubject(AnimalSubjectApiController.java:128)
	... 102 common frames omitted

fragkiadamis:
Update from test session 02/12/2025:

Preclinical subject creation fails with:

2025-12-02T13:29:50.855Z ERROR 1 --- [mcat-handler-17] o.s.n.s.c.GlobalExceptionHandler : Unexpected error in the rest service.
2025-12-02T13:29:50.861647245Z
2025-12-02T13:29:50.861660819Z org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Cannot deserialize value of type java.time.LocalDate from String "2015-03-09T00:00:00.000Z": Failed to deserialize java.time.LocalDate: (java.time.format.DateTimeParseException) Text '2015-03-09T00:00:00.000Z' could not be parsed, unparsed text found at index 10
2025-12-02T13:29:50.861682389Z

PR #2965, opened 2025-08-13
#2916 : missing * for required property fixed

DuckflipXYZ:
Tested in local and qualif, seems fine, no NgModel error raised anymore.


julien-louis:
fixed in the refacto

PR #2991, opened 2025-08-27
set custom selection model for dicomTree to allow only unique patient selection

LaurentPV:
In SHUP, user cannot select multiple DICOM studies or series from different patients.

Multiple selections in the dicom query tree can only be made if they come from the same patient.

To test : 
Query a PACS from ShanoirUploader and try to select multiple nodes from several patients.


michaelkain (approved):
Hi @LaurentPV, thank you for this clean PR. locally tested and it works.
With kind regards, Michael

PR #2992, opened 2025-08-27
Upgrade docker actions

a-ba:
This partially  fixes the freeze we haven when building the production images (the freeze still happens, but after the images are pushed)
See https://github.com/fli-iam/shanoir-ng/actions/runs/17263394694
No change for the qualif images (they still build normally).

Also fixes warnings in the dockerfile.


michaelkain (approved):
Hi @a-ba, perfect, thank you!

PR #3014, opened 2025-09-05
check value change of center / serialNumber

pierrehenri-dauvergne:
How to test:

Open an equipment
Edit
Change the center: the "update" button should be available

The same should happen with the serialNumber field

image

### DIFF
--- a/shanoir-ng-front/src/app/acquisition-equipments/acquisition-equipment/acquisition-equipment.component.ts
+++ b/shanoir-ng-front/src/app/acquisition-equipments/acquisition-equipment/acquisition-equipment.component.ts
@@ -217,6 +217,18 @@ export class AcquisitionEquipmentComponent extends EntityComponent<AcquisitionEq
     viewCenter(center: Center) {
         this.router.navigate(['center/details/' + center.id]);
     }
+
+    onCenterChange() {
+        console.log("center change");
+        this.form.markAsDirty();
+        this.form.updateValueAndValidity();
+    }
+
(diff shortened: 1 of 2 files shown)


shanoir-ng-front/src/app/acquisition-equipments/acquisition-equipment/acquisition-equipment.component.ts
```diff
@@ -30,7 +30,8 @@ import { ManufacturerModel } from '../shared/manufacturer-model.model';
 import { ManufacturerModelService } from '../shared/manufacturer-model.service';
 import { Center } from '../../centers/shared/center.model';
 import { ManufacturerModelPipe } from '../shared/manufacturer-model.pipe';
-
+import {of, timer} from "rxjs";
+import {catchError, map, switchMap} from "rxjs/operators";
```
julien-louis: please put a space after { and before }

julien-louis (changes requested):
I think the reason of the bug is the unicity check in edit falsely invalidate the current (manuf + serial) tuple. I mean you shouldn't be able to set an existing tuple to your equipment UNLESS it is the tuple currently saved in the db for this equipment we want to edit.

So, `uniqueEquipmentValidator` needs to deal with that case (it doesn't now and that is the problem). Maybe save the tuple as private propertie(s) in the component in `initEdit()` then in the validator, if the checked tuple equals this one, return null (null = valid).

The you can remove the manual uniqueEquipmentValidator and markAsDirty

PR #3032, opened 2025-09-22
fix uploader unit tests : update surefire

julien-louis:
Update surefire and specify the version for the uploader.

The version used was not compatible anymore and the `@BeforeAll public static void setup()` was not executed.

### DIFF
--- a/shanoir-ng-back/pom.xml
+++ b/shanoir-ng-back/pom.xml
@@ -192,10 +192,6 @@ along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
-				<version>3.0.0-M9</version>
-				<configuration>
-					<useSystemClassLoader>false</useSystemClassLoader><!-- due to a bug with openJDK -->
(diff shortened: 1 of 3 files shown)


michaelkain:
Hi @julien-louis,
thank you for this PR. Could you please give us some context on the issue?
Has it happened locally? Is it an issue on qualif or Github? develop seems to compile and normally the unit tests run.
With kind regards, Michael

julien-louis:
> Hi @julien-louis, thank you for this PR. Could you please give us some context on the issue? Has it happened locally? Is it an issue on qualif or Github? develop seems to compile and normally the unit tests run. With kind regards, Michael

Hi I'm not sure, I think it didn't fail in GitHub but has failed for a long time on my setup. On my computer the effective version of Surefire was very old and not compatible with the other libs (I believe JUnit). This silentely make the `@BeforeAll` to never be executed, provoking a test failure. The AI explained that as the Surefire version was not configured for the ShanoirUploader, the effective Surefire version was a **default value that depends on the maven version**. So maybe the maven version of Anthony is greater than mine and the default value for the Surefire plugin was compatible. I coulnd'nt find this version, mine is 3.8.7, providing a default surefire 2.12.4. The PR force a compatible version (I took the spring boot version).

PR #3041, opened 2025-09-23
Allow users to type in the study's name at import

pierrehenri-dauvergne:
How to test:

Open import page
Select zip, go to step 3
Instead of selecting a study, click on the input field and start typing a study's name
No error is thrown, you can select the study and finish the import
image

This PR also includes the ability to type in the name of centers, study cards, equipment and subjects
This PR also includes the alphabetical order of studies, centers, study cards, equipments and subjects

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -1,4 +1,4 @@
-name: Checkstyle Java Code
+name: Check code style
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


shanoir-ng-front/src/app/import/clinical-context/clinical-context.abstract.component.ts
```diff
@@ -366,43 +370,52 @@ export abstract class AbstractClinicalContextComponent implements OnDestroy, OnI
...
-            this.acquisitionEquipment = options[0].value;
+            newValue = options[0].value;
         }
+
         let founded = options?.find(option => option.compatible)?.value;
         if (founded) {
-            this.acquisitionEquipment = founded;
+            newValue = founded;
         }
+        Promise.resolve().then(() => {
```
julien-louis, line 383: `Promise.resolve().then()` seems useless to me, use directly `            this.acquisitionEquipment = newValue;`

pierrehenri-dauvergne: In this case, I added this to prevent the ExpressionChangedAfterItHasBeenCheckedError. I have this error when I select a center that has some equipment then select a center without equipment. The value of this.acquisitionEquipment go from a value to null in the same angular cycle, which throws the error mention above. By doing this Promise.resolve().then(), we change the value in another angular cycle. It could also be done with a setTimeout() but I saw elsewhere in the code that we often used Promise.resolve() so I did the same.

PR #3070, opened 2025-10-01
OHIF: final migration to v3.11.0

michaelkain:
Hi, this PR migrates the viewer OHIF to v3.11.0 (latest stable version without beta on docker by yesterday)
@pierrehenri-dauvergne Could you please test it on qualif? It works locally on my machine.
@a-ba Could you please check if it requires something more on your side for the MEP?
Thank you both and with kind regards,
Michael

### DIFF
--- a/docker-compose/Dockerfile
+++ b/docker-compose/Dockerfile
@@ -316,8 +316,6 @@ FROM ohif/app:v3.11.0 AS nginx-viewer
 
 FROM nginx AS nginx
 
-RUN apt-get update && apt-get install -y gzip
-
 COPY --link \
     nginx/http.conf \
     nginx/https.conf \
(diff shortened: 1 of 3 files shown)


pierrehenri-dauvergne:
Hi @michaelkain 
You're right, I forgot to check those boxes.
I deployed again : https://github.com/fli-iam/shanoir-deploy/actions/runs/18162574044 but the result is the same. I emptied my cache and removed cookies but I still have a blank page

michaelkain:
Hi @a-ba,
thank you for your help.
The problem seems to be when you ask for the .js file the server does not send the .js.gz file, what he should.
It works locally. I mean to remember you had an own nginx-template in your Ansible scripts and as it works
locally, if the new flag is just not propagated to qualif? We wait for your answer please. Thank you and with kind regards, Michael

a-ba:
this should do the trick

a-ba (approved):

PR #3075, opened 2025-10-01
Fix subjects in ms-datasets

pierrehenri-dauvergne:
Before running this script we have this:
image

Some subject have a study_id = null in datasets.
The subjects with this are the one that used to be in multiple studies. Since the single-study subject was started, we split subjects that used to be in multiple studies into multiple subject with a different study_id.
Something was missing in the script and we ended up with NULL values.

Also, some mixing up of the study_id of the newly created subject (here, subject with id=1 should have study_id=45 instead of 2).

After running the script:
image


michaelkain:
Hi,
after a lot of comparison and testing and 2 meetings. Error in previous migration script found
and fix is working on all columns in ms.datasets, so I approve. With kind regards, Michael

michaelkain (approved):

PR #3084, opened 2025-10-08
DICOMWeb STOW-RS (Extension for Karnak)

michaelkain:
Voted on the roadmap.

Depends on this pull request, to merge before: https://github.com/fli-iam/shanoir-ng/pull/2125
Usage of existing StudyInstanceUIDs and SeriesInstanceUIDs is necessary.

- Contains: **harmonize usage of dcm4che library**, now all using 5.31.1 as version and moved dependencies to ng-back
- Contains **bug fix for BIDS-import** (not working anymore after single-study-subject)
- MS Datasets: index datasets to Solr without using RabbitMQ (unnecessary (and slow) as all in same microservice)

Tests

- [x] Import of 1 DICOM serie works (all in viewer, all in download after) 
- [x] Test with very big DICOM study (30 series, 2.500 DICOM files): 3 minutes run-time
- [x] File-by-file comparison with sample1.zip (web import zip vs. STOW-RS): all equal 2.535 images in total)
- [x] Add StudyCenter in case new center added via DICOMWeb import
- [x] Tested with Enhanced DICOM MR

Regression tests:
- [x] Regression test with DICOM SEG in viewer OHIF (segmentation view)
- [x] Regression test with DICOM SR in viewer OHIF (standard view)
- [x] Create new center (no study): replicated to MS Datasets
- [x] Edit study add/remove center: replicated to MS Datasets
- [x] Create study, add center: replicated to MS Datasets
- [x] Delete center in MS Studies: replicated to MS Datasets (now working, fix in here)

### DIFF
--- a/docker-compose/database/db-changes/datasets/0061_stowrs.sql
+++ b/docker-compose/database/db-changes/datasets/0061_stowrs.sql
@@ -17,12 +17,12 @@ CREATE TABLE study_center (
 );
 
 INSERT IGNORE INTO `center` (id)
-SELECT DISTINCT center_id 
-FROM studies.study_center 
-WHERE center_id NOT IN (SELECT id FROM `center`);
+SELECT DISTINCT id 
+FROM studies.center 
+WHERE id NOT IN (SELECT id FROM `center`);
 
 # Replicate study_center into MS Datasets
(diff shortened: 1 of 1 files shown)


fragkiadamis:
I tested the DICOM SEG & SR importation and it works fine

PR #3098, opened 2025-10-15
[HOTFIX]: Download input datasets in DICOM format.

fragkiadamis:
Ensure downloading the input and output dataset in the correct format from the DatasetProcessingApiController.

The format control of the dataset name was not considering cases where the input dataset's name has not a format extension.

Example:
`format = dataset.getName().endsWith(".dcm") ? "dcm" : "nii";`
will assume a NIfTI format for the input dataset name `TOF WE TRA 2 SLABS PAT`, even though it's DICOM.


DuckflipXYZ (approved):
Nice management for non extensionned dataset name

PR #3100, opened 2025-10-15
Keycloak MFA

pierrehenri-dauvergne:
How to test:

Login and see if the shanoir theme is unchanged
MFA should be asked

This PR includes lots of changes in .ftl files because they weren't up to date with keycloak current version.
I tested it in neurinfo qualif and I just had to set to "required" the OTP config in keycloak. After that, when I logged in I got redirected to the MFA page, I used my phone to register and it worked well. After login out and login back in, I had to enter the code from my phone's app.


michaelkain:
Dear @fragkiadamis,
I hope you are doing fine. It would be really great to merge this big dev effort into develop,
to allow 2-FA with Shanoir and Keycloak. It is de-activated by default, but we have one instance
waiting for this feature. Could you please solve the conflict, luckily only in .yml, test it and merge it,
it it works? The feature should by default NOT be activated and will not at the moment be used on
our instances Neurinfo and OFSEP. But instance/server owners should be able to switch it on in Keycloak.

PR #3105, opened 2025-10-16
Fix async loading of prefilled values in dataset processing form

### LINKED ISSUE #3104: Import processed dataset bug
When trying to import a processed dataset in Shanoir, we need to fill out a form with details such as the study, subject, type, name, etc. After that, we must create a processed dataset entity to link the imported dataset with an input dataset and optionally an output dataset.

In this form, the study field should be preselected based on the previous form and should be disabled to prevent modification. However, due to a bug, the study is not preselected, while the field remains disabled, making it impossible to import the dataset from the UI.

Image

fragkiadamis:
Closes #3104

### DIFF
--- a/shanoir-ng-front/src/app/datasets/dataset-processing/dataset-processing.component.ts
+++ b/shanoir-ng-front/src/app/datasets/dataset-processing/dataset-processing.component.ts
@@ -128,15 +128,12 @@ export class DatasetProcessingComponent extends EntityComponent<DatasetProcessin
             this.breadcrumbsService.currentStep.getPrefilledValue('subject').then(res => this.prefilledSubject = res)
         ])
         .then(([study, subject]) => {
-            this.prefilledStudy = study;
-            this.prefilledSubject = subject;
-
-            this.studyOptions = [new Option(this.prefilledStudy, this.prefilledStudy.name)];
-            this.study = this.prefilledStudy;
+            this.study = study;
+            this.subject = subject;
             this.datasetProcessing.studyId = this.study?.id;
 
-            this.subjectOptions = [new Option(this.prefilledSubject, this.prefilledSubject.name)];
-            this.subject = this.prefilledSubject;
+            this.studyOptions = [new Option(study, study.name)];
+            this.subjectOptions = [new Option(subject, subject.name)];
         })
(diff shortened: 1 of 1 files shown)


DuckflipXYZ (approved):
Clean

shanoir-ng-front/src/app/datasets/dataset-processing/dataset-processing.component.ts
```diff
@@ -123,23 +123,22 @@ export class DatasetProcessingComponent extends EntityComponent<DatasetProcessin
...
-            }
+        return Promise.all([
+            this.breadcrumbsService.currentStep.getPrefilledValue('study').then(res => this.prefilledStudy = res),
+            this.breadcrumbsService.currentStep.getPrefilledValue('subject').then(res => this.prefilledSubject = res)
+        ])
+        .then(([study, subject]) => {
+            this.prefilledStudy = study;
+            this.prefilledSubject = subject;
+
+            this.studyOptions = [new Option(this.prefilledStudy, this.prefilledStudy.name)];
```
DuckflipXYZ: I'm not sure that this line is still required. Can you confirm ?

shanoir-ng-front/src/app/datasets/dataset-processing/dataset-processing.component.ts
```diff
@@ -123,23 +123,22 @@ export class DatasetProcessingComponent extends EntityComponent<DatasetProcessin
...
+            this.study = this.prefilledStudy;
+            this.datasetProcessing.studyId = this.study?.id;
+
+            this.subjectOptions = [new Option(this.prefilledSubject, this.prefilledSubject.name)];
```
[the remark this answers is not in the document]
fragkiadamis: Those are necessary too to populate the DatasetProcessingComponent, as well as to populate the subject select box field.

shanoir-ng-front/src/app/datasets/dataset-processing/dataset-processing.component.ts
DuckflipXYZ: Ok, I was wondering as we are not able to select a new study neither subject, as they are readonly

Issue #2940, opened 2025-07-29
Shanoir global statistics displayed on welcome page


LaurentPV:
As discussed during meeting with Alban Gaignard and Camille Maumet : 
Aggregated statistics to be displayed on Shanoir welcome page : 
- Number of "images"
- Number of Subjects
- Data size
- Number of queries per month
- Number of users
- Number of public datasets

e.g : Radboud data repo (https://data.ru.nl) writes "It has 9255 users. It contains 4374 data collections in editable state, 442 internally archived collections and 633 published collections that are accessible to the public ")

These metadata will have to be machine-readable (cf. schema.org).

cmaumet:
@LaurentPV - Thanks for reaching out. I would say the number of downloads on the top of my head but I don't have a clear visibility on what other operations could be. This would also have to be discussed with Isabelle.

PR #3128, opened 2025-10-28
Laurent 2940 welcome stats

LaurentPV:
Linked to Issue https://github.com/fli-iam/shanoir-ng/issues/2940

Displays statistics as below : 
Capture d’écran 2026-01-06 à 09 47 53


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/dto/OverallStatisticsDTO.java
```diff
@@ -0,0 +1,46 @@
+/**
+ * Shanoir NG - Import, manage and share neuroimaging data
```
julien-louis: I'm not sure we need a DTO when it's the same as the original class.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -507,6 +508,27 @@ public DatasetAcquisition getAcquisition(Dataset dataset) {
     }
 
     @Override
+    public OverallStatisticsDTO getOverallStatistics() {
```
julien-louis: There could be good reason to use a DTO inside the service layer and I might not understand everything here but if it's not justified we must use the original class.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/repository/DatasetRepository.java
```diff
@@ -157,6 +162,20 @@ public interface DatasetRepository extends PagingAndSortingRepository<Dataset, L
...
+    void computeOverallStatistics();
+
+    @Modifying
+    @Transactional
+    @Query(value = "UPDATE overall_statistics os SET os.storage_size = :totalStorageVolume WHERE os.stats_date = CURDATE()", nativeQuery = true)
+    void addTotalStorageVolume(@Param("totalStorageVolume") Long totalStorageVolume);
+
+    @Query(value = "SELECT studies_count, subjects_count, dataset_acquisitions_count, storage_size "
+            + "from overall_statistics WHERE stats_date = CURDATE()", nativeQuery = true)
+    List<Object[]> getOverallStatistics();
```
julien-louis: Why not use a custom OverallStatistics class instead of Object ?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/CreateStatisticsService.java
```diff
@@ -48,6 +51,15 @@ public class CreateStatisticsService {
     @Autowired
     private ShanoirEventService eventService;
 
+    @Autowired
+    private DatasetRepository datasetRepository;
+
+    @Autowired
+    private DatasetService datasetService;
```
julien-louis: I believe that mapper and service aren't used

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
julien-louis: Actually it seems it's just a naming issue sorry

shanoir-ng-front/src/app/datasets/shared/dataset.service.ts
```diff
@@ -32,6 +32,12 @@ import { DatasetType } from './dataset-type.model';
 
 export type Format = 'nii' | 'dcm';
 export type DatasetLight = {id: number, name: string, type: DatasetType, subject: IdName, hasProcessings: boolean, study: IdName, creationDate: Date};
+export type OverallStatisticsDTO = {
```
julien-louis: Same, if it is used as the model, don't name it DTO

julien-louis (changes requested):

shanoir-ng-datasets/src/main/java/org/shanoir/ng/configuration/security/SecurityConfiguration.java
```diff
@@ -106,4 +107,15 @@ public FilterRegistrationBean shanoirCorsFilter() {
         bean.setOrder(Ordered.HIGHEST_PRECEDENCE);
         return bean;
     }
+
```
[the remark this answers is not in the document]
LaurentPV: J'ai du procéder comme ça parce que oauth2ResourceServer valide le JWT sur datasets meme si le endpoint est en permitAll, ce n'est pas le cas sur Users surement parce que nous n'avons pas le multipartRelatedRequestFilter sur la conf de Users

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/dto/OverallStatisticsDTO.java
LaurentPV: Should I place the class in Model ?

michaelkain:
Hi @LaurentPV, hi @julien-louis,
I hope you are doing fine. I had a quick review on the PR and I see nothing alarming.
Thx for the screenshot of the GUI! My proposal for the GUI, one line each, centered:

**Statistics**
X active user(s)
X querie(s) made in the last month
X public data collection(s) detailed below, with X subject(s)?, X imaging dataset(s)?
X private? data collection(s), with X subject(s), X imaging dataset(s)
X Gb(s) of total storage volume

Are the data collections including public, or are they separated? If separate, we might add private.

cmaumet:
Hi @LaurentPV, 

As requested, in complement to @michaelkain review, I would suggest amending the text as follows:
 - Having first the number of data collections, images and subjects (most important).
 - Having last the number of active users and queries.
 - Typo: I think it is "GB" instead of "Gb"
 - Change "data collection" into "collection"
 - Text shortening 
   -  "representing a total volume of 1 GB" into "for a total of 1 GB"
   - "and are detailed below:" into "(see below)
 - If possible, making it optional to add a sentence at the beginning that would describe the instance. By default this could be "This is an instance of the Shanoir database" (but we could imagine instance owner customizing in their dashboard to say something like "This is the Shanoir instance of xxx").
 - Plural / singular
   -  When there is only one public collection change conjugate the verb to be "1 data collection *is* public" and for zero "There is no public collection".
   - When there is only 1 or 0 user : "There is 1 active user" / "There is no active user"
   - When there is 1 or 0 query: "1 query was run" / "No query was run"
 - Also using bold face for all the numbers  to emphasize all the number would be useful.

What is a query (Maybe we should try a new formulation but not sure what this means)? I am not so sure about this last figure (e.g. why over the last month, should it be the total number just for the other metrics).


For the example above we could get: 
> This is an instance of the Shanoir database. It has **2 collections** containing **16 images** of **6 subjects**, for a total of **1 GB**. **1 collection** is public (see below). There is **no active user** and **no query** was run during the last month.

or possibly using Michael's formatting above but still reordering as above. 

If you can share an update screenshot that would be great!

Issue #2717, opened 2025-03-17
Modify Shanoir in/output for POST /rest/executions to identify Shanoir instance


DuckflipXYZ:
Due to VIP infra update, they need to identify the emitting instance. The idea is to change the prefixe of (this is an example) : 

shanoir:/resource_id+b49d20f2-f92a-4555-ad9a-f5d79fa4f840+dataset.nii.gz?resourceId=b49d20f2-f92a-4555-ad9a-f5d79fa4f840&format=nii&converterId=2&token=[....]

to :
shanoir-ofsep-qualif:/resource_id .... or whatever similar

PR #3129, opened 2025-10-29
#2717 : Prefix update for VIP instance detection [HOTFIX]

fragkiadamis:


DuckflipXYZ (approved):
Changes specific to VIP, no risk and tested. Sounds good to me

PR #3141, opened 2025-11-06
Draft VIP local

michaelkain:


michaelkain:
Hi @DuckflipXYZ, just to note it down, that I might hopefully remember it.
Normally, next step to make: use either keystore "cacerts" or certificate "shanoir-ng-nginx.crt",
all in /etc/ssl/certs/java:ro, to have it used by VIP (Java behind running Tomcat).
As a reminder for me, the rights logs are here: /vip/tomcat/logs, not all info in vip.log and vip-errors.log

PR #3152, opened 2025-11-17
Angular deprecations

julien-louis:
* delete app.module and make every component standalone
* replace any `*ngIf` and `*ngSwitch` by `@if` and `@switch`
* replace `router.getCurrentNavigation()` by `router.lastSuccessfulNavigation`
* replace `xxx.toPromise()` by `firstValueFrom(xxx)`
* use angular core `inject()` instead of custom injector hack
* remove deprecated animation file, remove some animations, migrate some
* migrate subscribe syntaxe
* add depreciation warnings in the linter


julien-louis:
Done

PR #3156, opened 2025-11-18
fix tree download - PLEASE REVIEW

julien-louis:
downloading from the tree doesn't work if the downloaded exam / acq are not fully deployed

Instead, here, we ask the front for all the dataset ids that are implied in the selection.

* if datasets are selected, they are dl
* if an acquisition is selected, every dataset from this acquisition is dl
* if an exam is selected, every dataset from this exam is dl


fragkiadamis (approved):
looks good to me

Issue #2630, opened 2025-02-07
Refactor dataset copy feature


michaelkain:
in all related mappers
remove @transactional from mapper
only search info, where necessary
optimize perf in all cases and test with mass data

michaelkain:
Hi @pierrehenri-dauvergne, please work on it after the current release. Thank you, Michael

pierrehenri-dauvergne:
Hi @michaelkain I feel like this has been done already, I don't see any @Transactional in any mapper

michaelkain:
Hi @pierrehenri-dauvergne, please wait with this feature request.
I think we require a refactor of copy datasets, but we should take
into account a new demand on the roadmap.

michaelkain:
Question for CH: how will the copy take place: script of via web GUI?

chomo-ofsep:
Hi @michaelkain,
we do not require the exam-date shift.
For the hidden-pseudonymus-hash-values, the users with downloading rights should not have access to it but it would be better if the administrator still had access to it. I think a role user with downloading right does not have access to the export statistics feature so yes can the hidden-pseudonymus-hash-values be copied please. (can you confirm that the hidden-pseudonymus-hash-values are only available through the statistics export ?)
I confirm that at the beginning we won't need to copy processed datasets but we will need to in the future so to discuss further at a later point maybe ?

ebannier:
Hi, thanks for tagging us and adding us to the discussion. Is the creation of the derived study to be done from the study tree GUI only (and study by study) or also possible via SOLR / script ? For a start, starting from the study tree to select the series to link to the derived study would be very useful already. To the questions adressed to Celine : we would at Neurinfo NOT want to have an exam date shift, unless this is required by MR004 which I do not know.  As for processed data, if a segmentation is hanging to the dataset/examination, it would be nice to have the possibility to transfer it too. Regarding data provenance - are we supposed to keep a link to the original study (NCT reference, DUA) or to remove all provenance information on the contrary ?

michaelkain:
Hi, I close as initial feature in production. With kind regards, Michael

PR #3169, opened 2025-11-21
Write subjectName in dicom metadata at download (Prepa MR004)

pierrehenri-dauvergne:
How to test:

Open solr
Select a dataset
Download
Unzip and open a file: check the PatientID and PatientName tags, it should be the one of the subject

Copy that dataset to another study, with another subject (other subjectName too), do the same steps: the PatientID and PatientName must be those of the new subject.


pierrehenri-dauvergne:
This PR is in WIP because we do not offer the possibility for users to change the subject's name when we do a copy of datasets.

When this feature https://github.com/fli-iam/shanoir-ng/issues/2630 is available, this PR can be tested again to make sure the new subject's name appears in the datasets tag PatientID and PatientName

PR #3175, opened 2025-11-25
neurobagel

pierrehenri-dauvergne:
How to test:

Call the endpoint /datasets/bids/neurobagel/studyId/1 with postman for instance
A file should be returned looking like this:
```
participant_id	subject_identifier	
sub-1_subject1	1	
sub-2_DemoSubject2	2	

```


pierrehenri-dauvergne:
In addition to the participant_id and the subject_identifier, we also add the subject_age and subject_sex to the participants.tsv file. To do this, we have to create the file in ms-studies since those datas are stored in ms-studies and not in ms-datasets.
This forces us to call ms-studies from ms-datasets and create the file in a folder that is shared between both ms.

This PR adds a new endpoint in ms-studies that returns this:
```
participant_id	subject_identifier	subject_age	subject_sex	
sub-1_DemoSubject	1	12	F	
sub-2_DemoSubject2	2	1	F	

```

In addition to this endpoint, the participants.tsv file is also needed in the BIDS tab of a study's details.

Tested locally and this is the result:
image

PR #3200, opened 2025-12-04
Subject create form at eeg import

pierrehenri-dauvergne:
How to test:

Start EEG import
At step 3, click the "create new subject" button
The subject-creation form appears and the Study can't be changed:

image

Tested locally and on ofsep qualif and both were OK


michaelkain:
Hi @julien-louis, could you document this somewhere on the wiki?
With kind regards, Michael

PR #3202, opened 2025-12-04
VIP Execution: remove CAN_ADMINISTRATE check

### LINKED ISSUE #3195: VIP processing - CAN_ADMINISTRATE rights
Currently a user must have CAN_ADMINISTER rights in order to start a VIP processing execution.

Normally the user should be able to do that with the CAN_EXECUTE right #3138, so the CAN_ADMINISTRATE control should be removed.

fragkiadamis:
Remove CAN_ADMINISTRATE check from ExecutionApiController.

The rights check is now handled by the API @PreAuthorize using the CAN_EXECUTE right.

Closes #3195


DuckflipXYZ (approved):
Approved by code review only. Execution right check is on interface, so it's checked from UI as from PyNoir

Issue #3215, opened 2025-12-05
Center and Equipment disappear from Study Card update menu


LaurentPV:
After creating a study card with a specific center equipment linked to it as shown below : 

<img width="557" height="237" alt="Image" src="https://github.com/user-attachments/assets/8951b1a8-47fe-49e0-84d3-84ca085db208" />

When trying to update this study card the equipment is no longer selected and no equipment is shown in the menu : 

<img width="570" height="327" alt="Image" src="https://github.com/user-attachments/assets/e935c5f3-3461-45e6-9c77-822b0d0ea7fe" />

PR #3222, opened 2025-12-08
fix center in study card

julien-louis:
fix https://github.com/fli-iam/shanoir-ng/issues/3215


LaurentPV (approved):
Tested, works as it should !

PR #3227, opened 2025-12-09
[HOTFIX]: Download NIfTI conversion

fragkiadamis:
Fix NIfTI compression bug on dataset download process.


DuckflipXYZ (approved):
Sounds good to me, tested on qualif, correctly compressed at nifti converison

Issue #3255, opened 2026-01-08
Info bubble on password policies: update password page


michaelkain:
Hi @DuckflipXYZ,
I hope you are doing fine.
I am transferring you a demand of Céline.
Could you please move the Info bubble today on login page, to update password page: login-update-password.ftl?
The info bubble contains the policies and is more helpful on the actual page, where you change your password.
It would be great to integrate this into the release next week.
Thank you and with kind regards, Michael

PR #3260, opened 2026-01-09
#3255 : polices info bubble moved to password update page instead of login page

DuckflipXYZ:


DuckflipXYZ:
HI @michaelkain, yes, it has been tested succesfully (it's still deployed on ofsep qualif if you wanna take a look)

PR #3282, opened 2026-01-15
feature#3274: Assign automatically dataset type for VIP imported datasets

### LINKED ISSUE #3274: Processed dataset importation -->  Type assignment error.
During the imprtation of processed datassets, even though a type is assigned (MRDataset or EEGDataset for example), the dataset type is not saved in the database.

Also, develop feature to assign automatically the type of the dataset in case of DICOM dataset.

fragkiadamis:
### VIP Importation
1. The imported datasets from VIP were assigned with the GENERIC dataset type.
2. This PR introduces code to assign the type based on the DICOM's SOPClassUID.
3. If the dataset is not DICOM, then the GENERIC dataset type is assigned.

### Manual importation
1. Assign datasets to the correct (chosen by user) type.
2. Add missing types.
3. Closes: #3274


DuckflipXYZ (approved):
All good from code review and test

Issue #3284, opened 2026-01-16
Create subject issue (human or preclinical)


michaelkain:
Failed to evaluate expression 'hasRole('ADMIN') or (hasAnyRole('EXPERT', 'USER') and @studySecurityService.hasAnyRightOnStudy(#studyId))

PR #3286, opened 2026-01-16
Fix for Github #3284

michaelkain:


DuckflipXYZ (approved):
Tested with expert role, it worked well, but I found an other bug (not directly related), while testing : https://github.com/fli-iam/shanoir-ng/issues/3287

Issue #3300, opened 2026-01-23
Shanoir keycloak template styles


fragkiadamis:
The PR #3100 introduced some updates for the keycloak.

This broke some of the styles in the pages for reseting/updating the password

<img width="620" height="528" alt="Image" src="https://github.com/user-attachments/assets/2d61d1cf-3f9d-4838-bbb0-3f62eb2d4980" />

PR #3304, opened 2026-01-27
#3300: Fix the keycloak template styles

fragkiadamis:
Fix the style issues in keycloak templates (.tfl)

Closes: #3300 

How to test.
1) Compile and run the develop branch of shanoir-ng
2) Check templates for password update (for example) --> They should appear like in #3300 
3) Compile and run the version of this PR
4) Check the templates again, they should have the expected styles (like in shanoir.irisa.fr)


DuckflipXYZ (approved):
Tested and validated by test, but not by code review. Sounds good

PR #3314, opened 2026-01-29
Revert keycloak update

DuckflipXYZ:
Here is the revert of the Anthony's keycloak update ! Git revert would have removed all the changes added to develop after that commit, and git rebase -i was raising confilcts for some reason ... so it's not clean but i just did a new commit for that :/

Can not merge my own PR, so I let you do it


michaelkain (approved):
Hi @DuckflipXYZ, thank you!

PR #3317, opened 2026-02-02
feature#3063: ROLE_ADMIN only creates studies.

### LINKED ISSUE #3063: Create study, only by sysadmin (role ADMIN)
Today the role EXPERT can create a new study (research project) on a Shanoir server instance.
This can cause multiple difficulties to a server instance. If the disk space left on the instance
is not sufficient to host the new study during its runtime, this can lead to a server saturation
shut-down. Or if the new study and its importing users produce a lot of traffic, this can cause
performance impacts too.

For this reason, this issue proposes to remove the right to create new studies from the role
EXPERT and allow this only to sysadmins (role ADMIN), that have a more global view on the
entire server instance and its limitations.

We could structure it in 3 versions:
-  v1 "Just" switch rights in the backend (result: expert role gets a deny while click on create study), quick implementation
-  v2 Work on a form "Study creation request" and its process
-  v3 Work on a volume definition in the form and a volume control during study runtime, to declare the volume, that is required and control, that volume constraints are not exceeded

### LINKED ISSUE #3287: During import, proposed studies are not related to can_import right
During import, proposed studies (including the pre-selected one) are not related to the can_import right, but only to the membership, although it should be both. I saw it when I tried to create a new subject while importing, I had not the can_import (or can_administrate right, not sure about the relevant one), so clicking on the save button just raised a ms-studies error. In all cases, I should have not be able to select the study in the first import form page, as I had not can_import right.

fragkiadamis:
In this PR, the study creation procedure is modified in order for the ROLE_ADMIN only to create a new study.

ROLE_EXPERT creates a study draft which has to be approved by a ROLE_ADMIN user.

Close #3063 
Close #3287


shanoir-ng-users/src/main/java/org/shanoir/ng/email/EmailServiceImpl.java
```diff
@@ -93,6 +94,8 @@ public class EmailServiceImpl implements EmailService {
 
     private static final String LINK = "link";
 
+    private static final String IS_DRAFT = "isDraft";
```
michaelkain: uncertain if this is used here

michaelkain:
Hi @fragkiadamis, thank you for this rapid corrections.
I think, I found some kind of bugs/proposals:
- [x] We might visually better have to separate drafts from active studies, not only by the State, with some color and visual hints. It would be great if you could check this with Julien next week, he might have some CSS ideas about it.
- [x] **Bug**: when I create a draft as expert and mark it as "Display as public" true, it is directly shown on the Welcome page, what should be avoided as long as studies are drafts only (please fix)
- [x] Between bug and optimization: It would be great to just filter out studies from the import, that are drafts, that they can not even be selected for import, this would avoid error messages and end user confusion. So, I mean in Clinical Context during the import: do not show draft studies please. Please do it for all imports: BIDS, Bruker etc.
- [x] Same for Copy datasets and create new subject: better do not propose draft-studies as studies to select
- [x]  **Good point**: adding centers, quality cards and study cards is working and should be possible to rapidly start with the study, once approved, so this should be able as preparation work (in case necessary)
- [x] Potential optimization: when I click as ROLE_ADMIN on Draft Studies, it takes 5 seconds to load the 2 draft studies, what is very, very slow
- [x] **Bug**: I approved a draft study as ROLE_ADMIN and tried a normal MR-Web-GUI import into the study and got the following error:
"Failed to evaluate expression '!@importSecurityService.isDraftStudy(#importJob.getStudyId()) and (hasRole('ADMIN') or (hasAnyRole('EXPERT', 'USER') and @importSecurityService.hasRightOnStudy(#importJob.getStudyId(), 'CAN_IMPORT')))'"
- [x] same bug as above, when I try to make a preclinical import
- [x] Provide a fix for #3287

shanoir-ng-studies/src/main/java/org/shanoir/ng/study/controler/StudyApiController.java
```diff
@@ -163,6 +163,26 @@ public ResponseEntity<List<StudyDTO>> findStudies() {
         return new ResponseEntity<>(studyMapper.studiesToStudyDTOs(studies), HttpStatus.OK);
     }
 
+    @Override
+    @Transactional
+    public ResponseEntity<List<StudyDTO>> findApprovedStudies() {
```
michaelkain: Hi @fragkiadamis, I hope you are doing great.
A thing came in my mind, you told me yesterday, that you fixed the filtering-out of draft studies via the frontend on multiple places. Why do you not manage this in the backend?
On example: ShanoirUploader will call: /shanoir-ng/studies/studies/namesAndCenters to get the list of studies, he displays. The same might apply to the Python tools we have, that query the API to get the study list. Same for findStudies() above. From my point of view, you might not require a method findApprovedStudies as this might be the default behavior for findStudies(), maybe? This might simplify the changes in the frontend too, and include ShanoirUploader and Python API callers. So, when a study is a draft, it is only returned to its added members (CAN_ADMINISTRATE) with ROLE_EXPERT, to edit the study together, but to nobody else. So, it will not be displayed for any standard user, the might try an import and might even have CAN_IMPORT right to prepare the study. ROLE_ADMIN sees all studies, draft or active. What do you think, I hope this is not contrary to the previous spec? With kind regards, Michael

michaelkain (approved):

PR #3328, opened 2026-02-09
oneshot script cleaning subject_id

julien-louis:
* dataset.subject_id should match subject > acquisition > exam > subject
* except when the subject has a processing as origin because it doesn't even have an acquisition


michaelkain:
Done on develop already, so I close.

Issue #3309, opened 2026-01-28
[ShanoirUploader] Quality Control - Execute and filter at dataset acquisition level


LaurentPV:
For now, Quality cards can contain multiple quality examination rules and each rule can contain multiple conditions in order to tag the examination VALID, WARNING or ERROR.

With the implementation of quality control at import from ShanoirUploader we added the behavior of blocking an import if the quality tag is ERROR.

## Proposal

We would like to modify the behavior of quality cards so that instead of defining rules concerning all/at least X/none of the datasets of an examination we target the dataset acquisition (= DICOM serie) contained in the examination.

The Quality tag would be no more at Subject level but at dataset acquisition level and only dataset acquisition with a VALID or WARNING quality tag would pass the ShanoirUploader quality control at import.

This would allow the study administrator to precisely define quality filters and collect in his/her study only valuable DICOM series for an examination.

LaurentPV:
Closed with Pull Request #3557

Issue #2651, opened 2025-02-18
Apply quality card on big studies


pierrehenri-dauvergne:
When we apply a quality card on a study with a lot of subject/examinations (on ofsep-qualif there is one QC on study "Cohorte Principale - CHU Lyon" that crashes ms-datasets), it ends in error with below error:

```
2025-02-14_14:38:38.468 ERROR [phdauvergne(203)] o.s.n.s.c.GlobalExceptionHandler - Unexpected error in the rest service.
jakarta.servlet.ServletException: Handler dispatch failed: java.lang.OutOfMemoryError: Java heap space
        at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1103)
        at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:978)
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
        at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:564)
        at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:195)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
        at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:164)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
        at org.shanoir.ng.utils.MDCFilter.doFilter(MDCFilter.java:32)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:164)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:140)
        at org.shanoir.ng.dicom.web.StowRSMultipartRelatedRequestFilter.doFilter(StowRSMultipartRelatedRequestFilter.java:93)
[...]

PR #3330, opened 2026-02-09
Add qualityTag to dataset acquisition model

LaurentPV:
Linked to https://github.com/fli-iam/shanoir-ng/issues/3309

Solve https://github.com/fli-iam/shanoir-ng/issues/2651

**Refactorization of the Quality Control**

- [x] new field qualityTag is added to dataset_acquisition table in datasets database
- [x] fixed quality control performance issues, multiple quality controls can run  on 3000+ exams studies
- [x] tags added to dataset acquisition in study tree
- [x] quality card if "check at import" is true filter out DICOM series that have the quality tag ERROR at import
- [x] all quality card of a study to be checked at import are applied for each serie and the worst qualityTag is applied at the end
- [x] if all DICOM series are in ERROR the import is set in status ERROR with the quality control report
- [x] Quality control is applied on client side by ShanoirUploader and filters out DICOM series in ERROR the same way the import does it on server side.
- [x] QualityTags are properly stored in dataset_acquisition table


shanoir-ng-datasets/src/main/java/org/shanoir/ng/download/WADODownloaderService.java
```diff
@@ -341,7 +341,7 @@ public AcquisitionAttributes<Long> getDicomAttributesForAcquisition(DatasetAcqui
             try {
                 dAcquisitionAttributes.addDatasetAttributes(dataset.getId(), getDicomAttributesForDataset(dataset));
             } catch (PacsException e) {
-                throw new RuntimeException("Could not get dataset [" + dataset.getId() + "] dicom attributes from pacs", e);
+                LOG.warn("Could not get dataset [" + dataset.getId() + "] dicom attributes from pacs", e);
```
julien-louis, line 344: I'm not sure it's a good thing to remove the throw and put a log.warn instead. Here if a PacsException occurs it will be silent and could lead to inconsistencies in the quality check. After reading the code I believe the dataset will be ignored during the quality check. That's probably the correct behavior but It should still at least be reported as "unknown" (see method writeConditionsReport()). In general I believe that an exception should be catch only if we know how to handle the exception. With catch and log we hide it, with catch and throw we don't really catch it, we just add info.

for instance here I don't think it even makes sense to start the study card check if the pacs request failed : 

```
AcquisitionAttributes<Long> dicomAttributes = downloader.getDicomAttributesForAcquisition(acquisition);
                changeInAtLeastOneAcquisition = studyCard.apply(acquisition, dicomAttributes);
```

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/controler/QualityCardApiController.java
```diff
@@ -157,13 +158,20 @@ public ResponseEntity<QualityCardResult> applyQualityCardOnStudy(
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         }
         LOG.info("apply quality card: name:" + qualityCard.getName() + ", studyId: " + qualityCard.getStudyId());
-        QualityCardResult results = cardProcessingService.applyQualityCardOnStudy(qualityCard, true);
+        QualityCardResult results = null;
+        try {
+            results = cardProcessingService.applyQualityCardOnStudy(qualityCard, true);
+        } catch (MicroServiceCommunicationException e) {
+            LOG.error("Error during microservice communication while applying quality card on study", e);
```
julien-louis, line 165: Am I wrong or those catches would result in http 200 OK with null results ?

Instead of the logs you could 
* return a ResponseEntity with 500 
* OR throw a new RestServiceException (see the exemples in the others api controlers), il will be catch by GlobalExceptionHandler and will result in a ResponseEntity 500 and a warn log

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/controler/StudyCardApiController.java
```diff
@@ -202,7 +202,11 @@ public ResponseEntity<Void> applyStudyCard(
         StudyCard studyCard = studyCardService.findById(studyCardApplyObject.getStudyCardId());
         LOG.debug("re-apply studycard n° " + studyCard.getId());
         List<DatasetAcquisition> acquisitions = datasetAcquisitionService.findById(studyCardApplyObject.getDatasetAcquisitionIds());
-        cardProcessingService.applyStudyCard(studyCard, acquisitions);
+        try {
```
julien-louis, line 205: Same

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/rule/QualityCardRule.java
```diff
@@ -47,17 +44,19 @@
 
 @Entity
 @GenericGenerator(name = "IdOrGenerate", strategy = "org.shanoir.ng.shared.model.UseIdOrGenerate")
-public class QualityExaminationRule extends AbstractEntity {
```
julien-louis, line 50: Ok I see that now quality cards are applied at the acquisition level instead of the examination level.

Then is there a difference between QualityCardRule and DatasetAcquisitionRule which is applied to an acquisition as well ?

julien-louis (changes requested):

julien-louis:
@LaurentPV I've merged develop, don't forget to pull before commit.

LaurentPV:
All developments were transferred to PR [3557](https://github.com/fli-iam/shanoir-ng/pull/3557). 

Closing this PR.

PR #3343, opened 2026-02-16
run compute statistics at startup

LaurentPV:
Fixes compute overall statistics for welcome page allowing it to run at startup of ms datasets in addition to the CRON at 6 am.


michaelkain (approved):
Hi @LaurentPV, thank you! I approve by code review only and it looks just fine.
With kind regards, Michael

PR #3344, opened 2026-02-17
Download window bug fixes

DuckflipXYZ:
Some bug fixes : 
- Converter required when "nii" format is selected
- Window closed when we click on the 'x' button next to a selection (to remove 'nii' format when selected for example)

The reason why the HostListener to check the oustide click is removed is because the click on the 'x' button mentionned above is considered as an external window click for some reason


julien-louis (approved):

julien-louis:
I have fixed the clear button bug here : #3438

It was a vicious one : the click also were updating a var that was in the ngIf of the close button, so apparently the button was disappearing before we check if the target element (the button) is inside or outside the modal.

So I have put back the HostListener here.

PR #3356, opened 2026-02-23
feature#3355: Separation of database from migrations

### LINKED ISSUE #3355: Separate migrations from database
The migration scripts should exist in a different MS from the databse. The database MS should only host the database and nothing else.

fragkiadamis:
Extract database migrations into a dedicated container. The database container now only runs MySQL, with no migration logic. This separation prepares the application for a transition to Kubernetes, where migrations can be handled cleanly using init containers.

Close #3355


fragkiadamis:
@a-ba 

Also, the OFSEP and oneshot-updates directories in the database-migrations, I guess we should keep them, right?

a-ba (approved):

PR #3358, opened 2026-02-23
Use case-insensitive matching for Git error "Not a valid object name"

### LINKED ISSUE #3357: Use case-insensitive matching for Git error message "Not a valid object name"
## Summary

Git is changing the capitalization of its `fatal: Not a valid object name` error message to lowercase (`fatal: not a valid object name`) to follow Git's [CodingGuidelines](https://github.com/git/git/blob/master/Documentation/CodingGuidelines#L410-L416), which state that error messages should begin with a lowercase letter.

This change is under review on the Git mailing list:
https://lore.kernel.org/git/pull.2052.git.1771836302101.gitgitgadget@gmail.com

## Problem

Shanoir-NG currently uses **case-sensitive** string matching to detect this Git error, which will break when Git ships the updated message.

### Affected files

- **docker-compose/database/check_migration_names.py** — `"Not a valid object name" in proc.stderr`

## Suggested fix

Use case-insensitive comparison:
```python
"not a valid object name" in proc.stderr.lower()
```

This will work with both the current and future Git versions.

## Timeline

The Git change is currently under review and will likely be included in a future Git release. Fixing the matching now ensures forward compatibility.

mdferdousalam:
Fixes #3357

Git is lowercasing the `fatal: Not a valid object name` error message
to follow its CodingGuidelines. This change makes the string matching
case-insensitive so it works with both the current and future Git versions.

Changes:
- Use `.lower()` before checking for the error substring

See: https://lore.kernel.org/git/pull.2052.git.1771836302101.gitgitgadget@gmail.com

### DIFF
--- a/docker-compose/database/check_migration_names.py
+++ b/docker-compose/database/check_migration_names.py
@@ -23,7 +23,7 @@ for path in pathlib.Path("db-changes").glob("*"):
         if proc.returncode == 0:
             released_migrations = set(line.split("\t", 1)[1] for line in proc.stdout.splitlines())
 
-        elif "Not a valid object name" in proc.stderr:
+        elif "not a valid object name" in proc.stderr.lower():
             released_migrations = []
             latest_migration = ""
         else:


michaelkain:
Hi, I will close it as we already have a PR, that fixes this. With kind regards, Michael

PR #3370, opened 2026-02-27
issue#3369: fix for study card rule copy button not working 

### LINKED ISSUE #3369: Copy button on study-card and quality-card does not work
If you create/edit a study/quality -card, you can press the copy icon to duplicate the rule. Nothing happens

fragkiadamis:
The copy button on study card rules was silently doing nothing when clicked. The @Output in StudyCardRuleComponent is named copyRule but the parent study-card-rules.component.html was binding to (copy) instead, so the event was never caught.

Fixed by changing the binding in the parent template from (copy) to (copyRule) to match the child's output name.

Note: QualityCardRuleComponent has the same copyRule output and is bound the same way in the parent template, so that is fixed in this PR as well.

Close #3369


DuckflipXYZ (approved):
Sounds good by use test

PR #3372, opened 2026-02-27
issue#3371: restore import rules functionality on study cards

### LINKED ISSUE #3371: Import rules feature broken - rules cannot be selected and form not dirty after import
Select mode never activated: When navigating to /study-card/select-rule/select/:id, the StudyCardComponent opens in view mode but selectMode is never set to true despite the route declaring data: { select: true }. As a result, clicking rules does nothing and the "Import selected rules" button never appears.

fragkiadamis:
Fixes two bugs that together made the import rules feature completely unusable.

Bug 1 - Select mode never activated
The route /study-card/select-rule/select/:id correctly declares data: { select: true } in the routing module, but StudyCardComponent.initView() never read it. Added a single line to detect the flag and set selectMode = true

Bug 2 - Form not dirty after import
Fixed by setting a rulesImported flag on the breadcrumb step data, then checking it in buildForm() — which is called at exactly the right moment during form construction — and marking the rules control dirty there.

Close #3371


LaurentPV (approved):
Tested and approved

PR #3382, opened 2026-03-04
issue#3381: Allow blob URLs in frame-src for PDF iframe rendering

### LINKED ISSUE #3381: Blocked DUA notification image
The DUA notification image is blocked as it can be observed in the attached image.

If your user has DUA notifications, it can be observed in the dashboard, as well as in the notifications tab.

It is probably related with the recent CSP header addition.

Image

fragkiadamis:
Hello @michaelkain 

This is a very small and quick fix for the CSP header.

It was blocking iframe blobs and as a result, the DUA preview was not loading (check PR's related issue)

Closes #3381

### DIFF
--- a/docker-compose/nginx/http.conf
+++ b/docker-compose/nginx/http.conf
@@ -4,7 +4,7 @@ server {
 	server_name         SHANOIR_URL_HOST;
 
 	# Content Security Policy (CSP) configuration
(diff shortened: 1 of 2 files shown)


DuckflipXYZ (approved):
Not sure about eventual side effect, but it works, all good

PR #3389, opened 2026-03-06
3103 : Multi-job for VIP

### LINKED ISSUE #3103: Turning executionMonitoring 1 - 1 datasetProcessing to executionMonitoring 1 - * datasetProcessing
In fact, the core implementation is already made, we're just using lists of a single datasetProcessing. The goal of that issue is to review and adapt the aside features to work with a list of datasetProcessing.

DuckflipXYZ:
Not merged but deployed in prod in version 2.11.2.

A fix for DB is required with the last commit

Don't forget to git pull main PyNoir


michaelkain:
Thank you for your PR. I have not started to review it, but I have one question:
Is your PR designed in a way, that both operation modes: single-job (today) and
multi-job (future) are supported? In the idea, when we require scaling adaptations
for multi-job, we can still run single-job as before to remain operable?
With kind regards, Michael

DuckflipXYZ:
Hi @michaelkain, it's just a draft, no point to start a review, it's easier to visualize the changes I did some times ago.

I'm not sure to understand your reflexion, like, there is no point to keep the single-job if multi-job is available. It's juste an optimization of the feature thanks to a reshape, but the logic of  use is not changed at all. Run a single process in multi-job context won't cost more than in a single-job context, if it's your question.

michaelkain:
Hi @DuckflipXYZ, thank you for your answer.
Sorry, I did not do any review on your PR, as it was draft. I am just asking an overall question on this PR.
What I mean is, that with your PR the previous functioning remains retro-compatible.
In the idea, imagine after your PR, we have to run 3.000 exams and the server crashes with multi-job, that we tested nicely with 100 exams. I want to be sure, we remain operational and can still run 3.000 exams as single-job option.
Does your PR take these 2 modes into consideration or is it a break with the single-mode as today? or is there no risk at all?
With kind regards, Michael

fragkiadamis:
@DuckflipXYZ for reference I leave this comment on the minor UI issues of the PR.

1) The Execution monitoring has no dataset-acqusition related to it, and as a result, when you access it in the UI, we get a front-end error.

2) When starting a processing from the UI with multiple datasets, the job show that only 1 dataset is being processed.

fragkiadamis (approved):

PR #3395, opened 2026-03-06
issue#3394: Fix for failing bootstrap.sh execution

### LINKED ISSUE #3394: Bootstrap script fails
The bootstrap script fails after the introduction of the draft studies feature.

fragkiadamis:
Insert `study_extra_details` data into SQL initialization scripts

How to test:
Just execute the `./bootrapt --clean` command on the project's root dir (it should install without any errors).
No need for qualif deployment, since the fix is about the local installation of shanoir-ng

Should be merged before next release to avoid any issues with the clean installations from the users

Close #3394

### DIFF
--- a/shanoir-ng-studies/src/main/java/org/shanoir/ng/study/model/Study.java
+++ b/shanoir-ng-studies/src/main/java/org/shanoir/ng/study/model/Study.java
@@ -100,6 +100,7 @@ public class Study extends HalEntity {
     private boolean downloadableByDefault;
 
     /** End date. */
+    @NotNull
     @LocalDateAnnotations
     private LocalDate endDate;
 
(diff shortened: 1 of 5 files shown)


shanoir-ng-studies/src/main/java/org/shanoir/ng/study/model/Study.java
```diff
@@ -100,6 +100,7 @@ public class Study extends HalEntity {
     private boolean downloadableByDefault;
 
     /** End date. */
+    @NotNull
```
DuckflipXYZ: There are studies with null values in ofsep (I did not check for neurinfo). I might raise a lack of consistency, I'm not sure how to deal with it

DuckflipXYZ (commented):
It works, all goods except one tiny comment.

michaelkain (approved):

PR #3405, opened 2026-03-11
split keycloak internal url and frontend url in backend

AxelCouturierCIBM:
improvement for k8s support, without this, k8s microservices fail to reach auth server


a-ba:
closed in favour of #3471 which includes additional fixes

Issue #3403, opened 2026-03-11
Download fails for some acquisitions


julien-louis:
* download fails with acquisitions 710039, 727350, ...
* on Shanoir OF***
* firefox (don't know for chrome)

PR #3410, opened 2026-03-12
set WebClient maxInMemorySize to 500 Mb

LaurentPV:
Fixes issue https://github.com/fli-iam/shanoir-ng/issues/3403


michaelkain (approved):
Hi @LaurentPV, perfect, thank you a lot!

Issue #3124, opened 2025-10-22
Object storage (s3) compatibility


courvose:
The addition of S3 compatibility to Shanoir would greatly enhance its functionality. 
This feature would allow the storage of DICOM, BIDS related files, and other large files on S3 infrastructures (whether cloud-based or on-premise). This would facilitate easier and secure interfacing with other services by enabling the use of tokenized URLs. For instance, an authorized researcher using a high-performance compute cluster could request signed URLs of DICOMs from a Shanoir Study, allowing the cluster to directly request data from the storage infrastructure without going through the Shanoir VM. This would reduce the workload on Shanoir. It is important to note that the Shanoir PACS (DCM4CHE) is already compatible with S3.

S3 provides by design a lot of advantages in term of performance, scalability and security.
We plan to use on-premise S3 infrastructures and may use national trusted institutional partner infrastructure.

This issue has a strong link with the Kubernetes migration and therefore we will try to prioritize it higher.

courvose:
indeed for dcm4chee-arc S3 compatibility we can contribute with a pull request based on our config.
I did not get the BIDS import would not get supported anymore... It may cause some performance issues for large dataset download (send to compute cluster for ex). Topic should be discussed.
S3 compatibility would allow replacing the named volumes. All files can be stored (raw, DICOMs, EEG, (and Niftis if needed))
Topic should be discussed also.

michaelkain:
Hi @courvose, thank you for your fast answer. I have to correct: BIDS import will be supported
and I have to check on our side, in which volume we store it today and how.
I have been talking only of DICOM studies and the on-the-fly NIfTI/BIDS tree creation.
Not studies, that originally receive only NIfTI in BIDS.
With kind regards,
Michael

DuckflipXYZ:
Hi @michaelkain ! Only execution tracking files are contained in this folder, and in fact they are not used at all, so yes this folder does not require a long term storage in s3

a-ba:
> To estimate the impact, this issue will have on the application developer side, it would be great, if you could please answer my questions. I list below the named volumes, we have today:  
> * tmp: not in S3 (nothing to do for app devs)

The `tmp` volume is used for inter-container communications and its access patterns are undocumented. It could be replaced with a shared bucket to make the interactions very clear, with a recommended retention policy.

michaelkain:
Hi @a-ba, hi @courvose, thank you for your answers.
[...]

PR #3412, opened 2026-03-13
S3

michaelkain:
This pull request relates to Github issue #3124

It starts to introduce the usage of **S3 storage** into the microservices studies, datasets and preclinical via a
configuration placed into application.yml to make the switch.

The configuration option has been extended too for docker-compose usage: .env variables and entrypoints.

**CHANGE:**
Each of the 3 microservices has its own folder/bucket now, to clearly separate the concerns and align things.
**MS Studies: /var/studies-data OR studies bucket**
**MS Datasets: /var/datasets-data OR datasets bucket**
**MS Preclinical: /var/preclinical-data OR preclinical bucket** (called /var/extra-data today)
bids-data NOW outside of datasets-data
vip-data NOW outside of datasets-data

The BIDS-tree (bids-data) is as **by design a file-system based structure** and the BIDS validator only works on
a file-system based solution. Furthermore BIDS tree is generated by Shanoir and not considered as a persistent storage, as this would duplicate many data. Therefore bids-data, same as for vip-data exist only in the file-system.
vip-data are only temporary too, as they are imported into Shanoir. A persistent storage would again duplicate the data, what is not desired.

- /var/studies-data OR studies bucket contains study files (protocols, challenge zip files) OR DUA-files.
- /var/datasets-data OR datasets bucket contains extra-data files to human examination
- /var/preclinical-data OR preclinical bucket contains either extra-data files attached to a preclinical examination OR
pathology model files, attached to each pathology model.

**Implementation decisions:**
The long term goal of Shanoir is to produce as small and as efficient single microservices,
with the less amount of unnecessary libs and code used, so a minimalist approach to each
microservice is applied. Therefore architecture decisions have been taken:
- 1) we have NOT put the config for S3 into shanoir-shared.yml to avoid to overload none-persistent
storage microservices to have any lib or config, that do not need, keep it small
- 2) this results in some code duplication: S3 config in different application.yml
- 3) a new Maven project has been introduced in this PR: shanoir-ng-storage, that only
microservices depend on, that require persistent storage, this will allow the other microservices
to remain slim

More on this PR:
- BIDS tree generation: use prefix "study-", instead of "stud-"
- Move bids-data into a separate folder

**--------------------------------TESTING:---------------------------**
**File-system implementation**
- Study files and DUA: add files, delete files and download files
- OK, all files found in corresponding file-system studies-data

- Extra data on examination:
1. Create new exam with extra-data: OK
2. Create new exam, no extra-data: OK
3. Edit existing exam, add extra-data (1 or more files): OK
4. Delete exam with extra-data: OK
5. Delete exam without extra-data: OK

- Preclinical: extra-data and pathology model data

**S3 Implementation with S3-Ninja**
- Study files and DUA: add files, delete files and download files
- OK, all files found in corresponding bucket studies

- Extra data on examination:
1. Create new exam with extra-data: **OK**
2. Create new exam, no extra-data: **OK**
3. Edit existing exam, add extra-data (1 or more files): **OK**
4. Delete exam with extra-data: **OK**
5. Delete exam without extra-data: **OK**

- Preclinical: extra-data and pathology model data


michaelkain:
Hi @a-ba, I hope you are doing great.
I invite you as first reviewer Anthony, as I hope the general principal to have a bucket by microservice is fine for you?
I will ask another developer for a more specific Java code review, in case okay for you? I hope to make the switch via the application.yml variable sounds good to you too. With kind regards, Michael

a-ba:
Here it is ;-)

Bucket auto-creation should be an opt-in feature. In production the bucket would typically be created and preconfigured by the sysadmin and the access key provided to the shanoir microservice would have minimal permission (only GetObject+PutObject+DeleteObject+ListObjects on its own bucket, no ability to create a bucket, no access to the other buckets).

As the bucket name is not stored in the db (which is fine for me), i have two points :
- how do the urls in `dataset_file` look like? (i guess they are unchanged, eg: `file:/var/datasetr-data/...`)
- for each bucket we need to allow configuring an optional alternate target (another bucket accessed in read-only mode) to provide snapshot-like semantics (thus the qualif instances would use the qualif bucket as primary target and the prod bucket as secondary target, if the service gets a 404 on the from the primary bucket, it would fallback to the secondary bucket) 

How does the storage service gets to decide which storage implementation to use ? can they coexist in the same process?

Is the resulting filesystem layout identical on FileStorageService and on S3StorageService (in other words: can we migrate from file storage to the s3 storage without transformations).

Are the objects served directly from S3 to the client (using a http redirect) or do they go through the shanoir microservice?

> The BIDS-tree (bids-data) is as by design a file-system based structure and the BIDS validator only works on a file-system based solution.

Which file-system properties does it require?

If you need a herarchically organised layout, then this is possible in S3 using directory buckets:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-overview.html

> only temporary too, as they are imported into Shanoir. A persistent storage would again duplicate the data, what is not desired.

S3 buckets can be configured to be non-persistent.
<https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html>

From a sysadmin perspective it is generally easier to work S3 storage, it is truly elastic and is the bucket lifecycle can be finely tuned. A filesystem storage (even temporary) still has to be provisioned.

> /var/datasets-data OR datasets bucket contains extra-data files to human examination

How about the niftis? we still have plenty

> this results in some code duplication: S3 config in different application.yml

I do not mind having to provide the configuration for each microservice separately, however it is very confusing to repeat the config keys for *all* buckets in *all* microservices. I would pprefer that the keys are initialised with an empty or a sentinel value by default, eg:
```java
@Value("${storage.s3.studies-bucket-name:UNUSED}")
```
and that the S3StorageService raise a specific exception when it encounters the sentinel value.

The application.yml would only list the buckets actually used by the service.

> for clean up: on qualif, in datasets-data are many folders, that start with "sub-", which should
be old and come from an erroneous first bids-tree implementation, they can be deleted.

It would be interesting to have an admin feature that exports a lists of all files known by shanoir, so that we can identify the stale files (would be useful right now, and at any point in the future).

michaelkain:
Hi @a-ba, thank you for your great review. I will work on it in the next days, cheers, Michael

michaelkain:
Dear @a-ba, I hope you are doing fine.
I worked the entire day on your comments and I think I am done, with what I could in this iteration.
- Now each application.yml contains only the property used to address the specific resource. Thank you for your great proposal. Now when a microservice tries to address e.g. a bucket from another one, an exception is thrown and I use your UNUSED proposal. Thx!
- I use now longer keys, with e.g. var/datasets-data/examination-id/etc., to allow you to easily align the existing path with the new paths in S3. This was quite tricky to add, but I did it.
- Now the good point is: we can even store everything in one bucket, if we configure it like this.
- I will not introduce, see my comment on the original issue, some tmp/file hierarchy aspect in this PR, as it is already quite big and will double my development time effort, and was not initially planned like that. I can create an issue for an upcoming roadmap/development.
- The export of all files used, will have to be done manually (sorry, I can list you the tables) and not be part of this PR. I agree, useful indeed, but another roadmap issue from my point of view to create.
- At the moment, the objects go through the specific microservice, that protects them at the moment, via the rights he is aware of. Sorry, same here, a change here requires a new issue on the roadmap and was not specified on the original issue.
- The StorageService runs up with the implementation, either file-system or S3, so exclusively, not parallel, depending on the value of **storage.type**. The used bean is selected depending on the value.
With kind regards, Michael

michaelkain:
@a-ba PS: thx for your comment, I have to manage the nifti-download, for existing nifti-files. Normally the conversion is now temporary, on-the-fly, and I prefer to keep these non-persistent resources in the file-system, BUT I have to manage the case, where we kept the files in /var/datasets-data for legacy reasons.

a-ba:
>     * Now each application.yml contains only the property used to address the specific resource. Thank you for your great proposal. Now when a microservice tries to address e.g. a bucket from another one, an exception is thrown and I use your UNUSED proposal. Thx!
> 
>     * I use now longer keys, with e.g. var/datasets-data/examination-id/etc., to allow you to easily align the existing path with the new paths in S3. This was quite tricky to add, but I did it.

fine

For clarity, i would like to suggest the changes in: https://github.com/a-ba/shanoir-ng/commit/c87e9594ade2ebd51dfa274d761cbe5bbad4a146
Instead of setting a 'basedir' (reused from the file-storage properties, which kind of odd) i prefer having a separate 'prefix' key. 'prefix' is the usual term in s3 for this purpose and it has well-defined semantics.

>     * The export of all files used, will have to be done manually (sorry, I can list you the tables) and not be part of this PR. I agree, useful indeed, but another roadmap issue from my point of view to create.

agree
 
>     * At the moment, the objects go through the specific microservice, that protects them at the moment, via the rights he is aware of. Sorry, same here, a change here requires a new issue on the roadmap and was not specified on the original issue.

fine for me (actually that preferable for now because the s3 endpoint is not exposed to the public and the amount of data is rather low if we no longer store niftis)

>     * The StorageService runs up with the implementation, either file-system or S3, so exclusively, not parallel, depending on the value of **storage.type**. The used bean is selected depending on the value.

ok i missed that key. That's fine for me.

> @a-ba PS: thx for your comment, I have to manage the nifti-download, for existing nifti-files. Normally the conversion is now temporary, on-the-fly, and I prefer to keep these non-persistent resources in the file-system, BUT I have to manage the case, where we kept the files in /var/datasets-data for legacy reasons.

My point is that i did not find how these legacy files are served in S3, the only datasets-related method i found is `loadExtraData()`.

michaelkain:
Hi @a-ba, a big thank you for your PR, I merged it, much better now! I fully agree.
I work on the BIDS file storage into S3 and still have to add the NIfTI - management, testing and working on it.
Cheers, Michael

michaelkain:
Hi (thx to @a-ba!),
concerning the URL stored in path of dataset_file,
we go for solution 2) abstraction of the s3 endpoint, you store only the bucket and prefix names:
s3:datasets-bucket/datasets-prefix/study-2/sub-06-big-removed/ses-6/nirs/sub-06-big-removed_task-fingernonautodual_channels.tsv
With kind regards, Michael

michaelkain:
Hi @all,
I have tested all types of downloads between qualif (my PR) and production, for file-system storage.
ALL IS WORKING (study files DUA+protocol files, extra-data files, dataset_files (all 6 types)) as in production.
I merge, further test session for S3 only on qualif necessary, but already all working locally.
With kind regards, Michael

michaelkain:
Dear @fragkiadamis, I hope you are doing fine.

Could you please review my PR in the next 2-3 days?
I have already tested for file-system a lot on qualif.

My idea is to merge, to keep the existing and then
test further for S3 on qualif, and provide (if) minor
fixes for S3 only. With kind regards, Michael

fragkiadamis:
Hello @michaelkain 

If that's ok with you, I can review it tomorrow.

Best,

fragkiadamis (approved):

PR #3435, opened 2026-03-24
style draft studies

julien-louis:
image
image

### DIFF
--- a/shanoir-ng-front/src/app/studies/study/study.component.html
+++ b/shanoir-ng-front/src/app/studies/study/study.component.html
@@ -157,18 +157,18 @@ along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 							</ng-template>
 						</span>
 					</li>
-					<li [ngSwitch]="mode">
-                        <label i18n="Study detail|Draft state label@@studyDetailApprovalStatus">Draft</label>
-                        <span class="right-col">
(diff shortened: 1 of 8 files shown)


shanoir-ng-front/src/app/studies/study/study.component.html
```diff
@@ -152,14 +157,18 @@ <h2 class="header command-zone" i18n="Create study|Title@@studyDetailCreateTitle
...
+                            <label i18n="Study detail|Draft state label@@studyDetailApprovalStatus">Draft</label>
+                            <span class="right-col">
+                                    @if (study.isDraft) {
+                                        <span class="bool-true"><i class="fas fa-check"></i></span>
+                                    } @else {
+                                        <span class="bool-false"><i class="fas fa-times"></i></span>
+                                    }
+                            </span>
+                        </li>
+                    </ng-container>
```
fragkiadamis, line 171: Hello @julien-louis 

All good with the styles.

I added only one commit for the draft state display.
We had discussed with Michael to make the state always visible for `edit` and `view` modes, but I had completely forgotten to implement it.
Since it's a very small change I decided to include it in you PR to avoid opening a new PR only for this.

If it's not ok with you, I can always reset the commit :)

fragkiadamis (approved):

PR #3445, opened 2026-03-30
#3114 : data reuse agreement added

### LINKED ISSUE #3114: Checkbox to track participant agreement for data re-use
Hi,

In our latest consent form we are asking the participant two questions :   
- if they agree to take part in the study - mandatory
- if they also agree for the data to be re-used/in the context of neuroscience or neuroimaging research - optional

We would like to track in Shanoir, whether the subject has agreed to data re-use or not so that we can include them or not in derived datasets. 
(=add a check box at the exam level or which we could filter the examinations when creating derived datasets ) 

Elise

DuckflipXYZ:
Some notes: 
- Field is called dataReuseAgreement
- Not visible in solr tab, and not yellow left side filter tool, only filterable from text field at the top
- Can be set in the excell for the mass import in shUp, simply add a new column at the end (the name is not taken in account), the field will be true only if 1 or true (caps ignored) in the excell
- In the import dialog, dataExtraUseAgreement checkbox is disabled and ignored if "New exam check box" is not true

Don'nt forget to reindex Solr, and to remove "/var/solr/data" before in solr container.


docker-compose/database-migrations/db-changes/datasets/0066_add_data_extra_use_agreement_for_exam.sql
```diff
@@ -0,0 +1,14 @@
...
+--
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
+
+ALTER TABLE examination ADD COLUMN data_extra_use_agreement bool;
```
michaelkain: Hi @DuckflipXYZ, thank you a lot for this great PR.
I wonder, if the term "**data reuse**", as in the original Github issue of Elise might not be the better term to use in general for this PR. We have already extra-data in Shanoir, files attached to examinations (clinical or preclinical). The term "**data reuse**" might be better understandable by our users and hopefully by all developers too, as it relates probably better with the scientific papers/terms used. What do you think?

shanoir-ng-front/src/app/examinations/examination/examination.component.html
```diff
@@ -119,6 +119,12 @@ <h2 class="header command-zone"i18n="Create examination|Title@@examinationDetail
 							</ng-template>
 						</span>
 					</li>
+                    <li>
+                        <label i18n="Examination detail|Data extra use agreement label@@examinationDetailDataExtraUseAgreement">Data extra use agreement</label>
```
michaelkain: Please change at least here for "Data reuse agreement"

shanoir-uploader/src/main/resources/dicom_server.properties
```diff
@@ -42,5 +42,5 @@ dicom.server.truststore.password=password
 #******************************************************
 
 local.dicom.server.aet.calling=SHANOIR-UPLOADER
-local.dicom.server.host=localhost
+local.dicom.server.host=172.17.0.1
```
michaelkain: Hi @DuckflipXYZ, please clean up this file. Thank you, Michael

shanoir-uploader/src/main/resources/messages_fr.properties
```diff
@@ -184,6 +184,7 @@ shanoir.uploader.import.table.column.dicom.filter.serie.description=FILTER_DICOM
 shanoir.uploader.import.table.column.studycard=StudyCard (nom)
 shanoir.uploader.import.table.column.common.name=Identifiant
 shanoir.uploader.import.table.column.comment=Commentaire
+shanoir.uploader.import.table.column.data.extra.use.agreement=Accord de r\u00E9uilisation des donn\u00E9es
```
michaelkain: Perfect naming!

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/model/ImportJob.java
```diff
@@ -427,4 +429,11 @@ public void setStudyInstanceUID(String studyInstanceUID) {
         this.studyInstanceUID = studyInstanceUID;
     }
 
+    public Boolean getExaminationDataExtraUseAgreement() {
```
michaelkain: Hi @DuckflipXYZ, I think this extension might not be necessary, as the exam creation happens outside the ImportJob before it is actually send?

michaelkain (changes requested):
Hi @DuckflipXYZ,
thank you again for this great and complete PR.
All necessary inside and complete! Well done!
Please see my minor comments, please fix them.
My major proposal, probably rename the attribute
to a more widely used term like "Data reuse".
With kind regards, Michael
PS: except from that, ready to merge! :-)

docker-compose/database-migrations/db-changes/datasets/0066_add_data_extra_use_agreement_for_exam.sql
DuckflipXYZ: Ok, sounds good to me.

shanoir-uploader/src/main/resources/dicom_server.properties
DuckflipXYZ: yes, my bad

shanoir-ng-import/src/main/java/org/shanoir/ng/importer/model/ImportJob.java
DuckflipXYZ: For me, it's required, as it's used for the importDialog window to prefill the value when imported from Excell after the job has failed during the multi import, and it's used for exam creation, at ligne 386 of ImportFromTrableRunner (in the current PR)

michaelkain: Perfect, you are right!

michaelkain:
Hi @DuckflipXYZ, thank you. Perfect!
I am testing locally, is the filter on Solr missing to select only exams with dataReuse=true?
I mean as facet on the left.
I mean it was in the issue, that Elise wanted to do this. With kind regards, Michael

michaelkain:
Hi @DuckflipXYZ, I am testing locally and the new field, indexed into solr, is not in the column (empty column) and not in the column settings, please see screenshot. It would be great to put it as "Exam Reuse" behind "Exam Date". With kind regards, Michael
Screenshot From 2026-03-31 09-25-56

michaelkain:
Hi @DuckflipXYZ,
I try to summarize please:

- [x] Add facet to filter during Copy datasets from Solr
- [x] Fix empty column
- [x] Fix wrong db-changes script (typo)
- [x] Test feature on qualif and please give feedback to MK
- [x] Add Exam Reuse to Column settings
- [x] ShUp: I do not see the new field
Screenshot From 2026-03-31 09-51-18


With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/gui/ImportDialog.java
```diff
@@ -772,11 +774,32 @@ public ImportDialog(MainWindow mainWindow, String title, Boolean trueOrFalse, Re
         importDialogGBC.gridwidth = 1;
         formPanel.add(mrExaminationCommentTF, importDialogGBC);
 
+        mrExaminationDataReuseAgreementLabel = new JLabel(
+                resourceBundle.getString("shanoir.uploader.mrExaminationDataReuseAgreementLabel") + " *");
```
michaelkain: Hi @DuckflipXYZ, please remove asterix. THX!

michaelkain:
Hi, now I have this and the dialog does not open.
2026-03-31_10:02:22.835 [AWT-EventQueue-0] ERROR o.s.u.action.ImportDialogOpener - Cannot invoke "java.lang.Boolean.booleanValue()" because "examinationDataReuseAgreement" is null
java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "examinationDataReuseAgreement" is null
	at org.shanoir.uploader.action.ImportDialogOpener.updateImportDialogForNewExamFields(ImportDialogOpener.java:277)
	at org.shanoir.uploader.action.ImportDialogOpener.openImportDialog(ImportDialogOpener.java:93)

DuckflipXYZ:
For processed_data, I consider there is an agreement if all the exam relative to its inputs have dataReuseAgreement = true

michaelkain:
Hi @DuckflipXYZ, thank you for your corrections. I will test again.
One more important point came in my mind: the moment we release, many old ShUps will still be importing.
So, we need to assure, that it still works, even if the ExamDTO does not contain the new attribute. Is this already the case, I did not test or check the code?

michaelkain:
@DuckflipXYZ concerning processed_data: let's start like this.

DuckflipXYZ:
> Hi @DuckflipXYZ, thank you for your corrections. I will test again. One more important point came in my mind: the moment we release, many old ShUps will still be importing. So, we need to assure, that it still works, even if the ExamDTO does not contain the new attribute. Is this already the case, I did not test or check the code?

Hi @michaelkain, I just checked, it works without issue with older shUp versions (I used the develop one)

michaelkain:
Hi @DuckflipXYZ, perfect, thank you for your test with old ShUp, great. I will test locally.

docker-compose/database-migrations/db-changes/datasets/0066_add_data_reuse_agreement_for_exam.sql
```diff
@@ -0,0 +1,14 @@
...
+-- This program is free software: you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation, either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
+
+ALTER TABLE examination ADD COLUMN data_reuse_agreement bool DEFAULT false NOT NULL;
+UPDATE examiantion SET data_reuse_agreement = false WHERE true;
```
michaelkain: Hi @DuckflipXYZ, I hope you are doing fine. I wonder how this can work with the wrong table name?
Have you tested your PR on qualif?

michaelkain:
During my tests, when I configure a study without-study-card, the import does not work.
My last dev/tests on this topic worked, so is this a new issue on your PR or does it exist on develop?
With kind regards, Michael

docker-compose/database-migrations/db-changes/datasets/0066_add_data_reuse_agreement_for_exam.sql
DuckflipXYZ: Hi ! Wow, nice catch. I guess the data remain consistent when loaded in the back even by being null. The script had already run when I added that line, my bad.

DuckflipXYZ:
> Hi @DuckflipXYZ, I hope you are doing fine. During my tests, when I configure a study without-study-card, the import does not work. My last dev/tests on this topic worked, so is this a new issue on your PR or does it exist on develop? With kind regards, Michael

Hi @michaelkain, can you a explain a bit more ? with this branch, I have no issue to import in a study wihtout study card from the Web page, and from shUp, the field study card seems to be mandatory, so it does not seem to be any issue, just it's not allowed (or I did not understand you well)

image

michaelkain:
Hi @DuckflipXYZ, thank you for your answer. The import should work without the error message.
A study, where study-card is disabled, should just work, without study-card. Center and equipment
to select, same as for web-zip-import. I mean to have implemented it and tested it, last year.
With kind regards, Michael

michaelkain (approved):
Hi @DuckflipXYZ, my local tests look all very fine, thank you.
We have to clarify the no-study-card use case, but it does not look like related
to your acutal dev, therefore I merge. Thank you again and with kind regards, Michael

PR #3451, opened 2026-04-03
feature#2745: Database migration: MySQL 5.7 → MariaDB 11.4 LTS

### LINKED ISSUE #2745: Update MySQL server 5.7 to 8.4

fragkiadamis:
Migrated the database from MySQL 5.7 to MariaDB 11.4 LTS.

The data directory was reused as-is, no dump/restore needed. Just in case I did a dumb of my local dev databases, that would be imported to mariadb after (as described [here](https://mariadb.com/docs/server/server-management/install-and-upgrade-mariadb/migrating-to-mariadb/moving-from-mysql/mysql-to-mariadb-migration-the-master-guide#logical-migration-dump-and-restore)), but finally it was not necessary.

```bash
mysqldump --user=root --password --databases datasets import  migrations preclinical studies users \
--single-transaction --routines --events \
--triggers --hex-blob | mariadb --host=new_server_ip --user=root --password
```

The only things that required attention were running `mariadb-upgrade` once to fix the internal system tables, and switching the JDBC driver from mysql-connector-j to mariadb-java-client in the microservices config.

`mariadb-upgrade -uroot -ppassword`

If mariadb-upgrade is not performed, there will be an issue with the procedures

Everything else just worked

Close #2745


fragkiadamis:
Hello @a-ba 

The PR to change the database to MariaDB is ready and you can review it.

Just one major issue:
Locally it works perfectly but when I try to deploy on neurinfo-qualif, I get the following failure:

```
Container neurinfo-qualif-database-1  Started
 Container neurinfo-qualif-keycloak-database-1  Started
 Container neurinfo-qualif-keycloak-database-1  Waiting
 Container neurinfo-qualif-database-1  Waiting
 Container neurinfo-qualif-keycloak-database-1  Healthy
container neurinfo-qualif-database-1 has no healthcheck configured


MSG:

non-zero return code
PLAY RECAP *********************************************************************
shanoir-neurinfo-qualif-vm : ok=56   changed=27   unreachable=0    failed=1    skipped=51   rescued=0    ignored=0   
error: ansible error (exit 2)
Error: Process completed with exit code 1.
```

It seems that there is not healthcheck configured for the database?? Not really sure what is going on.
Could you please have a look and give me a feedback on this?

Have a good weekend!

a-ba:
This is because the docker-compose.yml in qualif has a `service_healthy` dependency on the database container and mariadb image does not have a [healthcheck](https://docs.docker.com/reference/dockerfile/#healthcheck) config while mysql image did have one.

I changed it to `service_started` and i added a `mariadb-upgrade` stage in the deployment procédure. The qualif should work correctly now.


PS: how about the keycloak-database image ? (migrating to mariadb should be a matter or changing `KC_DB=mysql` to `KC_DB=mariadb`)

fragkiadamis:
Hello @a-ba,

I talked about the keycloak DB with @michaelkain and we decided to stick only in our database for the change.

Should we do it for keycloak too you think?

a-ba:
> Should we do it for keycloak too you think?

it would make sense to me because keycloak officially supports mariadb 11.4 and it no longer supports mysql 5.7 and it is simpler to have all databases running on the same server version.
  https://www.keycloak.org/server/db#_supported_databases

fragkiadamis:
@a-ba I have made the changes for the keycloak database too. I guess now that we will need something similar to the shanoir database for the health checks and the mariadb-upgrade in the qualif.

fragkiadamis:
@a-ba also please have in minf that in this [PR](https://github.com/fli-iam/shanoir-ng/pull/3486), I updated the keycloak version from 26.2.1 to 26.2.5 due to a keycloak bug that existed and was interfering with our work.

a-ba:
> I have made the changes for the keycloak database too. I guess now that we will need something similar to the shanoir database for the health checks and the mariadb-upgrade in the qualif.

yes i already change that yesterday.
Currently both database containers have the following config which works indifferently with mysql and mariadb
```yml
    healthcheck:
      test: ["CMD", "/bin/bash", "-c", "echo -n > /dev/tcp/127.0.0.1/3306"]
```


> also please have in minf that in this PR, I updated the keycloak version from 26.2.1 to 26.2.5 due to a keycloak bug  > that existed and was interfering with our work.

fine for me

actually i have noticed that bug and discarded it as just a display bug, i did not expect it could have destructive effects when creating new attribuets

a-ba (approved):

PR #3472, opened 2026-04-15
populate the users database with dummy accounts

a-ba:
Rather that creating initial accounts with real users and real email address, this PR creates dummy accounts.

The dummy account have a non-routable email address (so that they can only be activated through the keycloak admin console).


DuckflipXYZ:
Not sure about the interest of giving them invalid adresses, but ok, it works

DuckflipXYZ (commented):
All good

DuckflipXYZ (approved):

PR #3476, opened 2026-04-16
ci: compose deployment

cmdoret:
## Objective
This PR adds a CI workflow to run the docker compose deployment and post-install steps as described in readme, and validate the installation success.

## Changes

It adds `.github/workflows/deployment-test.yaml` which:

* overrides /etc/hosts
* Builds the migration image.
* Runs the bootstrap script (without build, as I could not get image to build on CI)
* Runs smoke tests in case one service fails.
* Runs functional tests to verify that databases are correctly initialized, authentication works, and API endpoint respond.
* Exports compose logs in case of failure.

It also adds a compose override `docker-compose.ci.yml` which changes env variables for the CI environment.

The bootstrap script now uses the standard `COMPOSE_FILE` environment variable to easily enable this override with `COMPOSE_FILE: "docker-compose.yml:docker-compose.ci.yml"`.

## Questions

* At the moment, the migration is built in CI because the image tag referred to in `docker-compose.yml` does not exist on the registry (`ghcr.io/fli-iam/shanoir-ng/database-migrations:NG_v2.9.3`) is this an error? It would be better to avoid hard-coding a tag in CI and use the compose file as a single source of truths for image tags anyways.
* Currently, the job runs on pull requests to main and develop. I was not sure what would be the best time to run it and verify the deployment process. As it currently pulls image from the registry, it should probably run after latest images have been pushed. Or maybe I should override all image tags to point to latest?

> Disclaimer: This is implemented with the help of Claude AI.


a-ba:
>At the moment, the migration is built in CI because the image tag referred to in docker-compose.yml does not exist on the >registry (ghcr.io/fli-iam/shanoir-ng/database-migrations:NG_v2.9.3) is this an error?

The database-migration container is fairly new, it was introduced in [v2.10.0](https://github.com/fli-iam/shanoir-ng/pkgs/container/shanoir-ng%2Fdatabase-migrations).

> It would be better to avoid hard-coding >a tag in CI and use the compose file as a single source of truths for image tags anyways.

or maybe set the image tag via an environment variable


>Currently, the job runs on pull requests to main and develop. I was not sure what would be the best time to run it and verify >the deployment process. As it currently pulls image from the registry, it should probably run after latest images have been >pushed. Or maybe I should override all image tags to point to latest?

Running on every PR might be too much for our quotas.

A periodic run should be fine. We already have a daily build of the docker images using the latest commit of the 'develop' branch. They are pushed to https://github.com/orgs/fli-iam/packages?repo_name=shanoir-deploy

jbardet (changes requested):

jbardet:
Status: the automatic PR run (pull path) fails with `missing table card_condition`. This is the workflow doing its job, not a workflow bug: the PR path uses the latest PUBLISHED GHCR images, which predate the @Check fix (#3541, on develop as 3598907c1). Hibernate doesn't halt on DDL errors, so the bad @Check makes CREATE TABLE card_condition silently fail at init, then validate fails.

Proof the fix deploys cleanly: a manual `build_from_source=true` run on this branch passed end-to-end (build from source, schema-drift check with no card_condition error, smoke + functional tests, restart persistence).

Added a workflow_run trigger so the smoke test runs right after "Publish Docker images" succeeds (validates each release at publish time), and demoted the weekly cron to monthly since all images are pinned. Note: workflow_run only activates from the default branch, so it takes effect once this merges to develop.

For maintainers:
1. Publish an image set containing #3541 (cut a release tag or run deploy-qualif on develop HEAD); then the pull path / auto goes green.

### What it validates
Full stack via bootstrap.sh (SHANOIR_MIGRATION=auto), schema-drift check, smoke + functional tests, restart persistence. Two modes: pull pre-built GHCR images (default), or build_from_source=true (manual) to validate unpublished code.

### Notable fixes
- bootstrap.sh: drop docker -t/-i and fix HOME/tmp path for non-interactive CI Maven build
- post-restart validation: wait for microservice readiness + refresh nginx after restart
- new fail-fast "schema/DDL drift" assertion (turns a 5-min timeout into an immediate, clear error)

### Scope / non-goals
Fresh install only. No upgrade tests, no real data, CI-only stack. Upgrade migrations (e.g. study_card_condition -> card_condition) are NOT exercised here.

jbardet:
Hi @a-ba. PR is ready for another look. The branch is up to date with develop, includes the deployment smoke test workflow, and the card_condition / #3541 blocker is resolved now that NG_v2.12.1 images are published.

CI isn’t running yet because workflows need maintainer approval on this fork PR. Could you please:

- Approve and run workflows on the PR
- Re-review and approve when green
- The deployment test uses the default pull path (image_tag=auto → newest NG_v*.*.*). No build_from_source needed.

Thanks!

jbardet:
Validated against NG_v2.13 on our GitLab DinD pipeline. Two fixes needed for current Keycloak images:

JWT issuer: docker-compose.ci.yml must use the nginx frontend issuer (post hostname-v2 / #3306).
Bootstrap race: wait_db_network_ready before Keycloak init on fresh --clean deploys.

jbardet:
Hi @a-ba, the branch is up to date with develop. Could you: 

1. **Approve and run workflows on this PR**: fork PRs need a maintainer click, and no CI has ever run here, so the checks have been empty since April rather than failing.
2. Review and merge once green. The default path (`image_tag=auto` → newest `NG_v*.*.*`) needs no configuration.

Note that the `workflow_run` trigger only becomes active once this is on the default branch, so per-release validation starts working after the merge, not before.

PR #3489, opened 2026-04-27
"Add" button in subject & exam

julien-louis:
According to the rights spec : https://github.com/fli-iam/shanoir-ng/wiki/Roles-and-rights-specification 

* on subjects list & and exam list : show "add" button for USER role
* in the form of those 2 entities, on the studies select-box, list studies that have CAN_IMPORT (or CAN_ADMINISTRATE) for this user


shanoir-ng-front/src/app/subjects/subject/subject.component.ts
```diff
@@ -86,7 +87,8 @@ export class SubjectComponent extends EntityComponent<Subject> implements OnDest
                 private subjectService: SubjectService,
                 private studyService: StudyService,
                 private downloadService: MassDownloadService,
-                private studyRightsService: StudyRightsService) {
+                private studyRightsService: StudyRightsService,
+                private userRightsService: StudyRightsService) {
```
fragkiadamis, line 91: Hello @julien-louis 

All good with me, the PR works great. Just a minor thing. The `private studyRightsService: StudyRightsService` here seems that it is not used at all.

fragkiadamis (approved):

PR #3492, opened 2026-04-29
add refresh button to BIDS validator

julien-louis:
image


michaelkain:
Hi @julien-louis, I let you take into consideration the comments of copilot, and once done, let me know, when I should have a look. With kind regards, Michael

michaelkain (approved):

PR #3496, opened 2026-04-30
adding SpringSecurity Annotation

LaurentPV:
fix issue https://github.com/fli-iam/shanoir-ng/issues/3491


LaurentPV:
Tested with user fmakki today, fix the issue.
Reday to be merged

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/service/SubjectServiceImpl.java
```diff
@@ -256,7 +256,7 @@ private Subject mapSubjectStudyListToSubject(Subject subject) throws ShanoirExce
 
         if (subject.getStudy() != null && subject.getStudy().getId() != null) {
             Long studyId = subject.getStudy().getId();
-            Boolean isDraft = studyRepository.findIsDraftById(studyId);
+            Optional<Boolean> isDraft = studyRepository.findIsDraftById(studyId);
             if (Boolean.TRUE.equals(isDraft)) {
```
michaelkain: Hi @LaurentPV, thank you for the adaptation. Does the comparison work between Optional and Boolean.TRUE?

michaelkain (changes requested):

shanoir-ng-studies/src/main/java/org/shanoir/ng/subject/service/SubjectServiceImpl.java
LaurentPV: You are right It returned false every time that's why the test worked. Fixed in last commit

michaelkain (approved):

Issue #3302, opened 2026-01-26
Center creation during import from ShanoirUploader and Center rights management


LaurentPV:
# This issue is a proposal from the CHU of Nantes for the Roadmap 2026

## Context and Issue
For now, study promoters enter the centers contributing to the study with a specific name.

This name is linked to a specific prefix in the study menu and during import the subject common name is already prefilled with the prefix associated to the selected center. This feature is helpful to prevent errors and heterogeneity in subjects pseudonyms.

When using ShanoirUploader for the imports (and especially when using the mass import feature), the center is automatically created for the selected study based on the acquisition center name content of the DICOM data. This results of newly created centers for the study without any validation from the study promoter and with ambiguous names as they come from the DICOM tag.

Here is an example :  

<img width="417" height="372" alt="Image" src="https://github.com/user-attachments/assets/788b6454-190d-4646-bc9e-a56bb649c15a" />

For the user doing the import, this behavior is problematic because the new center is created but the user has no importing rights concerning this center unless a study administrator give him/her those rights, and with DICOM center names as ambiguous as "IRM 3T" it could be difficult for the study admin to know what to do.

This issue underline the differences between concepts of Acquisition Center (as defined in Shanoir, one acquisition device belongs to a center and must correspond to the DICOM data) and the contributions center, which matches the understanding of a study promoter that collects data from specific centers, without the necessity of knowing exactly from which acquisition center it belongs in the beginning.

For now, [developments have been made](https://github.com/fli-iam/shanoir-ng/issues/3183) to allow an acquisition device to be duplicated from an acquisition center to a contributing center. This will help this feature linking acquisition devices to contributing center.

The idea of this feature would be to make imports only using the centers defined by the study administrators to keep the study management cleaner even if the examination took place in another center and the data was collected afterwards in the defined contributing center.

**The original acquisition center informations would anyway still be kept in the DICOM data.**

## Proposed solutions

[!WARNING] Another Issue is open to fix the fact that a user could have rights on a center he/she created during import : https://github.com/fli-iam/shanoir-ng/issues/3303

### The most simple Solution

* For the "mass import from an excel file" feature we add a SHANOIR_ACQUISITION_CENTER column to the excel template and specify to ShanoirUploader the center to select for the import. This center must match the center on which the user has rights.

### Managing Center rights in the Shanoir Study Menu

[...]

PR #3511, opened 2026-05-06
Extend mass import for centerId

michaelkain:
This implements the crucial part (MVP) of the roadmap issue #3302


michaelkain:
I updated already the mass import doc on the wiki for the new column SHANOIR_CENTER_ID.
I tested locally:
- ROLE_ADMIN: OK with centerId given, OK no centerId given
- ROLE_USER, right can import: OK with centerId given, OK no centerId given
all worked nice.
Happy testing and with kind regards, Michael

LaurentPV:
it works fine but I see one issue: 
- it allows to link examination to a center that is not defined in the study (I think it's best to let the administrator of the study define which center is linked to his/her study)

We can either let this happen but then add the new center to the study and in the end it recreates the issue we tried to fix. 
Or we can just send an error message if center id does not exists in the study.

LaurentPV (approved):

PR #3514, opened 2026-05-07
AutoIndex fix

DuckflipXYZ:
the idea of the fix is that even if the cmp command line was returning a non 0, it was not going to the second block, also it should, that's why the logic is modified by removing the "||" part and the "set -e"


michaelkain:
Hi @DuckflipXYZ, thank you for your PR. Could you successfully test it on qualif? With kind regards, Michael

DuckflipXYZ:
Hi @michaelkain, sry for lateness, sadly not yet, at first sight it seems to work, but an other issue is raised, so the deployement is not succesfull ... I'm not sure it's linked to this or not, that's why I'm asking Anthony if he knows about it

DuckflipXYZ:
It's successfully tested

michaelkain (approved):

michaelkain:
Hi @DuckflipXYZ, thank you for this important fix and your qualif-test! I approve by code review only. Cheers, Michael

PR #3526, opened 2026-05-12
Adding feature to put acq dateTime in acq folder layer when downloading outputs

DuckflipXYZ:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -18,6 +18,7 @@
 import jakarta.annotation.PostConstruct;
 import jakarta.servlet.http.HttpServletResponse;
 import org.apache.commons.io.FileUtils;
+import org.assertj.core.util.Lists;
```
fragkiadamis: AssertJ is a test-scoped library, it shouldn't exist in the production code.

In line 231 `Lists.list("AcquisitionDate", "AcquisitionTime")`, use `List.of("AcquisitionDate", "AcquisitionTime")` instead.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetServiceImpl.java
```diff
@@ -585,6 +585,26 @@ public File extractDicomMetadata(List<Long> datasetIds, List<String> metadataKey
...
+            metadatas = mapper.readTree(metadataStr).get(0);
+        } catch (Exception ignored) {
+        }
+
+        for (String metadataKey : metadataKeys) {
+            if (Objects.nonNull(metadatas) && metadatas.has(metadataKey) && metadatas.get(metadataKey).has("Value")) {
+                results.put(metadataKey, metadatas.get(metadataKey).get("Value").get(0).asText());
+            }
+        }
+        LOG.info("TEMPO : " + results);
```
fragkiadamis: This seems like a leftover debug. Do we need it for the logs?

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -226,14 +225,29 @@ protected Map<Long, String> getDatasetDownloadPath(List<Dataset> datasets, Strin
             }
 
             if (sorting.contains("acquisition")) {
-                path += "/Acq_id_" + relevantDataset.getDatasetAcquisition().getId();
+                if (sorting.contains("acquisitionDate")) {
+                    String dateTime;
+                    if (Objects.isNull(relevantDataset.getDatasetAcquisition().getAcquisitionStartTime())) {
+                        Map<String, String> dateTimeMap = datasetService.getSpecificDicomMetadataValues(relevantDataset, Lists.list("AcquisitionDate", "AcquisitionTime"));
+                        dateTimeMap.putIfAbsent("AcquisitionDate", "NoDate");
+                        dateTimeMap.putIfAbsent("AcquisitionDate", "NoTime");
```
fragkiadamis: Copy-paste error. The second key should be "AcquisitionTime"

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetService.java
```diff
@@ -174,4 +174,7 @@ public interface DatasetService {
 
     @PreAuthorize("hasRole('ADMIN') or (hasAnyRole('EXPERT','USER') and @datasetSecurityService.hasRightOnEveryDataset(#datasetIds, 'CAN_SEE_ALL'))")
     File extractDicomMetadata(List<Long> datasetIds, List<String> metadataKeys) throws Exception;
+
+    @PreAuthorize("hasRole('ADMIN') or (hasAnyRole('EXPERT','USER') and @datasetSecurityService.hasRightOnEveryDataset(#datasetIds, 'CAN_SEE_ALL'))")
+    Map<String, String> getSpecificDicomMetadataValues(Dataset dataset, List<String> metadataKeys);
```
fragkiadamis: Here, the first argument of `getSpecificDicomMetadataValues` is a `Dataset` entity, and not a list of `datasetIds`. That means that the `@PreAutorize` is a copy-paste error. It should be `hasRightOnDataset` and not `hasRightOnEveryDataset`

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -226,14 +225,29 @@ protected Map<Long, String> getDatasetDownloadPath(List<Dataset> datasets, Strin
...
+                        dateTime = LocalDateTime.parse(dateTimeMap.get("AcquisitionDate"), DateTimeFormatter.ofPattern("yyyyMMdd")).format(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
+                            + "_"
+                            + dateTimeMap.get("AcquisitionTime").substring(0, 2) + "-" + dateTimeMap.get("AcquisitionTime").substring(2, 4);
+                    } else {
+                        dateTime = relevantDataset.getDatasetAcquisition().getAcquisitionStartTime().format(DateTimeFormatter.ofPattern("dd-MM-yyyy_HH:mm"));
```
fragkiadamis: This seems that will create a folder named: `Acq_date_15-01-2023_14:30_acq_id_42`. The colon (:) is an illegal character in the OS for using in paths

fragkiadamis: Also it is not consistent with line 236 where you use a dash (-).

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
fragkiadamis: Also, the putIfAbsent will add the key-value pair if the key is absent. If the key (AcquisitionDate or AcquisitionTime) is there but with an empty value, nothing will happen and the keys will remain with empty values.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -226,14 +225,29 @@ protected Map<Long, String> getDatasetDownloadPath(List<Dataset> datasets, Strin
...
 
             if (sorting.contains("acquisition")) {
-                path += "/Acq_id_" + relevantDataset.getDatasetAcquisition().getId();
+                if (sorting.contains("acquisitionDate")) {
+                    String dateTime;
+                    if (Objects.isNull(relevantDataset.getDatasetAcquisition().getAcquisitionStartTime())) {
+                        Map<String, String> dateTimeMap = datasetService.getSpecificDicomMetadataValues(relevantDataset, Lists.list("AcquisitionDate", "AcquisitionTime"));
+                        dateTimeMap.putIfAbsent("AcquisitionDate", "NoDate");
+                        dateTimeMap.putIfAbsent("AcquisitionDate", "NoTime");
+                        dateTime = LocalDateTime.parse(dateTimeMap.get("AcquisitionDate"), DateTimeFormatter.ofPattern("yyyyMMdd")).format(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
```
fragkiadamis: Here it seems that you format only date (yyyyMMdd), so you should use `LocalDate` instead of `LocalDateTime`. It could throw an error if the value to fromat represents only date and no dateTime.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -226,14 +225,29 @@ protected Map<Long, String> getDatasetDownloadPath(List<Dataset> datasets, Strin
...
+                            + "_"
+                            + dateTimeMap.get("AcquisitionTime").substring(0, 2) + "-" + dateTimeMap.get("AcquisitionTime").substring(2, 4);
```
fragkiadamis: Again, what happens in the case that there are the placeholder ("NoDate"/"NoTime") values signed?? The result will be like `No-Ti`. It will not make sense.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
fragkiadamis: Also, what happens in the case that there are the placeholder ("NoDate"/"NoTime") values signed??

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetDownloaderServiceImpl.java
```diff
@@ -226,14 +225,29 @@ protected Map<Long, String> getDatasetDownloadPath(List<Dataset> datasets, Strin
...
+                    } else {
+                        dateTime = relevantDataset.getDatasetAcquisition().getAcquisitionStartTime().format(DateTimeFormatter.ofPattern("dd-MM-yyyy_HH:mm"));
+                    }
+                    path += "/Acq_date_" + dateTime + "_acq_id_" + relevantDataset.getDatasetAcquisition().getId();
+                } else {
+                    path += "/Acq_id_" + relevantDataset.getDatasetAcquisition().getId();
+                }
```
fragkiadamis: The nesting goes to deep here... Difficult to read/maintain the code. See if it can be simplified a bit.

fragkiadamis (changes requested):
Hello @DuckflipXYZ 

I reviewed the PR and I found some minor issues. Have a look at them please and let me know when to comeback and re-review.

DuckflipXYZ:
Hi @fragkiadamis ! Thanks for reviewing, that was such a messy branch, good catches

fragkiadamis (approved):
All good!

PR #3529, opened 2026-05-13
issue#3364: Fix study user removal

### LINKED ISSUE #3364: Removing access to a study for a user from the user page not working.
Image

Removing access to a study for a user from the user page as above is not working.

fragkiadamis:
Fix the study user removal from the user template.

Closes #3364


LaurentPV (approved):
Tested, works as expected.

Issue #3530, opened 2026-05-13
ShanoirUploader - 500 ERROR when calling SubjectApi.findSubjectByIdentifier() endpoint


LaurentPV:
Users from OFSEP with property mode.subject.study.identifier=true have 500 ERRORS during the SubjectApi.findSubjectByIdentifier CALL when clicking on "importer" in the Import tab.

Logs on server side : 
2026-05-13T16:23:19.071+02:00 ERROR 1 --- [mcat-handler-64] o.s.n.s.c.GlobalExceptionHandler         : Unexpected error in the rest service. 
studies-1  | 
studies-1  | java.lang.NullPointerException: Cannot invoke "org.shanoir.ng.subject.dto.SubjectDTO.getId()" because "subjectDto" is null
studies-1  | 	at org.shanoir.ng.study.security.StudySecurityService.hasRightOnSubjectForEveryStudies(StudySecurityService.java:394)

PR #3531, opened 2026-05-13
add PostAuthorize condition

LaurentPV:
linked to https://github.com/fli-iam/shanoir-ng/issues/3530


LaurentPV:
@michaelkain waiting for your advise on this subject as OFSEP rules can be specific.

If mode.subject.study.identifier=true and subject is not found with its identifier as it is a new subject then the SubjectAPI.findSubjectByIdentifier() @PostAuthorize does not return a 500 as studySecurityService.hasRightOnSubjectForEveryStudies() on the empty subjectDTO is irrelevant.

We could also avoir the NullPointer in StudySecurityService.hasRightOnSubjectForEveryStudies() but not sure if this method must return false or true if subjectDTO is NULL.

michaelkain (approved):

Issue #3532, opened 2026-05-18
ShUP Mass Import - let user fill import form if examination already exists with same date + comment


LaurentPV:
For now if an examination already exists for the selected subject with the same examination date the user has to fill the import form to choose what he/she wants to do : 

- fill the existing examination with new data
- create a new examination

For some cases users would need to separate different series from one examination to severals containing only specific DICOM series fulfilling their needs.

In those cases adding the examination comment to the date would allow mass import to classify the series selected with the filter series description column of the mass import table in different examinations with specific comment.

PR #3533, opened 2026-05-18
adding exam comment to check if exam already exists

LaurentPV:
Linked to issue #3532


shanoir-uploader/src/main/java/org/shanoir/uploader/action/ImportFromTableRunner.java
```diff
@@ -360,13 +364,13 @@ private boolean importData(ImportJob importJob, org.shanoir.uploader.model.rest.
...
                     Examination examination = (Examination) iterator.next();
-                    // Existing exam found with the same study date: stop importJob and take next one
+                    // Existing exam found with the same study date and comment: stop importJob and take next one
                     Date examinationDate = examination.getExaminationDate();
                     LocalDate examinationLocalDate = examinationDate.toInstant()
                         .atZone(ZoneId.systemDefault())
                         .toLocalDate();
-                    if (examinationLocalDate.equals(studyDate)) {
-                        logger.info("Import job only downloaded, manual user decision needed: existing examination with the same date.");
+                    if (examinationLocalDate.equals(studyDate) && ((examination.getComment() != null || studyDescription != null) && examination.getComment().equals(studyDescription))) {
```
michaelkain: Hi @LaurentPV, thank you for this good PR!
Hm, I would maybe prefer, to create a new examination only, if the Excel SHANOIR_EXAM_COMMENT
is different from the existing exam.comment? This forces the mass importing user, to explicitly set its
new comment in the Excel, and therefore force a new creation. I think, very less of the Excel table creators will know the actual studyDescription. Maybe I read the if wrong. Cheers, Michael

LaurentPV: Corrected in last commit.

michaelkain (approved):
Hi @LaurentPV, perfect! Thank you, I approve by code-review only.
With kind regards, Michael

PR #3542, opened 2026-05-22
feature#3430: Set database names with variables

### LINKED ISSUE #3430: customisable database names
#3356 technically allows hosting the databases on an external mysqld server but their names are still hardcoded to `users`, `datasets`, `studies`, `import`, `preclinical` and `migrations`, which renders the feature impractical.

Three changes are needed:
- all the db parameters for the `migrations` db must be configurable (should be in the same way as in #3380)
- we have multiple db migration scripts working on multiple databases simultaneously:
  ```
  $ git grep -E '(datasets|users|studies|preclinical|import)\.' docker-compose/database-migrations/db-changes/|wc -l
  184
  ```
  These names have to be translated to the actual name configured by the sysadmin. This is not feasible in native SQL, we have to expand the variables before calling mysql (eg. like the entrypoint in the nginx container).
  ```sh
  sed "s/USERS_DB/$SHANOIR_DB_NAME_USERS/ ; s/DATASETS_DB/$SHANOIR_DB_NAME_DATASETS/ ; ..."
  ``` 
- we must document that these six databases are required to be hosted on the same server

fragkiadamis:
This PR makes every schema name independently configurable via environment variables, while preserving the existing names as defaults so current deployments are unaffected.

Closes #3430

### DIFF
--- a/docker-compose/database-migrations/shanoir-entrypoint.sh
+++ b/docker-compose/database-migrations/shanoir-entrypoint.sh
@@ -33,15 +33,56 @@
 DB_CHANGES_DIR="/opt/db-changes"
 DB_INIT_PROCEDURES_DIR="/opt/db-init-procedures"
 
-MIGRATION_DB=migrations
-MIGRATION_USER=migrations
-MIGRATION_PASSWORD=password
+# Configurable database names — defaults preserve backward compatibility.
+# Note: all six databases must reside on the same MariaDB server.
+: ${SHANOIR_DATASETS_DB_NAME:=datasets}
+: ${SHANOIR_STUDIES_DB_NAME:=studies}
+: ${SHANOIR_IMPORT_DB_NAME:=import}
+: ${SHANOIR_USERS_DB_NAME:=users}
+: ${SHANOIR_PRECLINICAL_DB_NAME:=preclinical}
+: ${SHANOIR_MIGRATIONS_DB_NAME:=migrations}
+
+MIGRATION_DB="$SHANOIR_MIGRATIONS_DB_NAME"
+MIGRATION_USER="${SHANOIR_MIGRATIONS_DB_USER:-migrations}"
+MIGRATION_PASSWORD="${SHANOIR_MIGRATIONS_DB_PASSWORD:-password}"
 MARIADB_HOST="${MARIADB_HOST:-database}"
 
 HEADER="[Shanoir Entrypoint]"
 
 MARIADB="mariadb            -h$MARIADB_HOST -u$MIGRATION_USER -p$MIGRATION_PASSWORD"
 MARIADBADMIN="mariadb-admin -h$MARIADB_HOST -u$MIGRATION_USER -p$MIGRATION_PASSWORD"
+
+# Map a db-changes directory name to the configured database name
+get_db_for_dir()
+{
+	case "$1" in
+		datasets)    echo "$SHANOIR_DATASETS_DB_NAME" ;;
+		studies)     echo "$SHANOIR_STUDIES_DB_NAME" ;;
+		import)      echo "$SHANOIR_IMPORT_DB_NAME" ;;
+		users)       echo "$SHANOIR_USERS_DB_NAME" ;;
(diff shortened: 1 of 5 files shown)


fragkiadamis:
Hello @a-ba

I started working on #3430 I have put the names of the databases in variables, have a look and let me know if it's good to go or not.

Have a good weekend!

shanoir-ng-datasets/src/main/resources/scripts/test-data-h2.sql
```diff
@@ -10,17 +10,6 @@
 -- You should have received a copy of the GNU General Public License
 -- along with this program. If not, see https://www.gnu.org/licenses/gpl-3.0.html
 
--- Manually create this table
```
a-ba: this seems unrelated to the PR

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/CardCondition.java
```diff
@@ -38,7 +38,7 @@
 import jakarta.validation.constraints.NotNull;
 
 @Entity
-@Check(constraints = "(dicomTag IS NOT NULL AND shanoirField IS NULL) OR (dicomTag IS NULL AND shanoirField IS NOT NULL)")
```
a-ba: this seems unrelated to the PR

docker-compose/database-migrations/shanoir-entrypoint.sh
```diff
@@ -33,16 +33,57 @@
...
+get_db_for_dir()
+{
+	case "$1" in
+		datasets)    echo "$SHANOIR_DATASETS_DB_NAME" ;;
+		studies)     echo "$SHANOIR_STUDIES_DB_NAME" ;;
+		import)      echo "$SHANOIR_IMPORT_DB_NAME" ;;
+		users)       echo "$SHANOIR_USERS_DB_NAME" ;;
+		preclinical) echo "$SHANOIR_PRECLINICAL_DB_NAME" ;;
+		migrations)  echo "$SHANOIR_MIGRATIONS_DB_NAME" ;;
+		*)           echo "$1" ;;
```
a-ba: i would prefer the default to yield an error (if a new db is introduced it will certainly have to be translated too)

docker-compose/database-migrations/shanoir-entrypoint.sh
```diff
@@ -33,16 +33,57 @@
...
+	esac
+}
+
+# Substitute hardcoded database names in SQL content with configured names.
+# Database names must only contain alphanumeric characters and underscores.
+apply_db_name_substitutions()
+{
+	sed \
```
a-ba, line 75: kind of fragile
(but i don't have a better solution that would not be overengineered ;-)

shanoir-ng-datasets/src/main/resources/scripts/test-data-h2.sql
fragkiadamis: Hello @a-ba 

These changes were cherry-picked from another PR as there were some errors with the DB. Now that I rebased the branch with develop, you will notice that these commits and their changes are not in the PR anymore

shanoir-ng-datasets/src/main/java/org/shanoir/ng/studycard/model/condition/CardCondition.java
fragkiadamis: The same as above.

docker-compose/database-migrations/shanoir-entrypoint.sh
fragkiadamis: Fixed!

docker-compose/database-migrations/shanoir-entrypoint.sh
fragkiadamis: Agreed. I've added a comment documenting the known failure modes so the next person who touches this knows what to watch out for.

a-ba (approved):

PR #3545, opened 2026-05-26
fix(front): enable save when form becomes dirty without status change

jbardet:
## Summary

Fixes the footer **Update** / **Save** button staying disabled when the form becomes dirty without a validation status change (e.g. adding a center on study edit).

## Issue

Update button stays disabled when adding a center.

When editing a study that already has at least one center, adding another center on the **Centers** tab does not enable the **Update** button, although the form was changed.

### Steps to reproduce

1. Open a study that already has one or more centers.
2. Click **Edit**.
3. Go to **Centers**, add a center with the **+** control.
4. Observe the **Update** button in the footer.

### Expected

**Update** becomes enabled so the new center can be saved.

### Actual

**Update** stays gray/disabled until another field triggers a validation status change (e.g. editing **End date** on General).

### Cause

In `EntityComponent.manageFormSubscriptions()`, `footerState.valid` is updated only on `form.statusChanges`. Adding a center marks the form dirty via `studyCenterList` but often does not change validation status, so the footer is not refreshed.

### Environment

- Shanoir-NG front (Angular), study edit mode
- Reproduced on 2.11.x

## Changes

- `entity.component.abstract.ts`: recompute `footerState` on `statusChanges`, `valueChanges`, and `form.events`.

## How to test

1. Edit a study with ≥1 center → **Centers** → add a center → **Update** enables (no need to edit end date).
2. Regression: edit any entity form, change a field, confirm save still enables when valid.

### DIFF
--- a/shanoir-ng-front/src/app/shared/components/entity/entity.component.abstract.ts
+++ b/shanoir-ng-front/src/app/shared/components/entity/entity.component.abstract.ts
@@ -25,7 +25,7 @@ import {
 } from '@angular/core';
 import { AbstractControl, FormArray, FormGroup, UntypedFormBuilder, UntypedFormGroup, ValidationErrors } from '@angular/forms';
 import { ActivatedRoute, NavigationExtras, Router } from '@angular/router';
-import { firstValueFrom, Subject, Subscription } from 'rxjs';
+import { firstValueFrom, merge, Subject, Subscription } from 'rxjs';
 
 import { Selection, TreeService } from 'src/app/studies/study/tree.service';
 import { SuperPromise } from 'src/app/utils/super-promise';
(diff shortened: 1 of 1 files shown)


fragkiadamis:
Hello @jbardet 

I cannot reproduce the bug. Check my steps in the following video.

https://github.com/user-attachments/assets/816a9bcf-f669-415a-9b5b-ee9da5eafbbf

jbardet:
Yes, it is because it is a new study that has all the required fields. If you take one that has been created before Shanoir version 2.11, it doesn't have the new mandatory fields: 
image
And thus update fails. See here: 

https://github.com/user-attachments/assets/e498e561-6340-4893-8d02-b477621a4416

julien-louis:
Hi @jbardet , thanks for your pr. 

I cannot reproduce the described bug either. 

Also the problem you have in the video seem to to a date validation problem. The dateOrdervalidator in StudyComponent returns a negative validation result if the startDate and the endDate are the same : 

```if (endDay <= startDay) return { order: true };```

jbardet:
Hi, thank you for your work. It is indeed hard to reproduce the bug because it has been induced by Shanoir's version update. So it would need to create a Study in a previous version of Shanoir, do an update to v2.11 and then modify the study. But then it is a minor bug so I can close it if it is too much of a pain for a study editing validation. Concerning the Study Start and End Dates, does this condition has been just added? Or how is it possible that my Start and End Dates are the same in my Study?

fragkiadamis:
I close the PR since we tried with @jbardet to reproduce the bugs and we did not manage. They seem to be resolved after the merge from the PR #3546

PR #3551, opened 2026-05-28
fix(dicomweb): return buffered DICOM body from findInstance

jbardet:
## Summary

- Fixes `DICOMWebService.findInstance()` returning an **empty** response body because the PACS HTTP entity was read twice (`EntityUtils.toByteArray(entity)` after the stream was already consumed for subject-name rewriting).
- OHIF/dcmjs then failed with `Invalid DICOM file, expected HEADER is missing` when saving/exporting segmentations (full instance fetch via DICOMweb), while frame viewing could still work via `findFrameOfStudyOfSerieOfInstance()`.
- Regression introduced in `27d632fadd` (*Introduce subject-name cache…*, MR-004 subject-name in DICOMweb).

## Changes

- Use the existing `dicomBytes` buffer for `ByteArrayResource` and set `Content-Length` from `dicomBytes.length`.
- Add `DICOMWebServiceTest` with mocked PACS responses: asserts non-empty body and valid DICOM Part 10 preamble (`DICM` at byte 128), including with `subjectName` set.

## How to verify

1. Open an examination in OHIF with a segmentation.
2. Save/export segmentation to Shanoir (not only “Download DICOM SEG”).
3. In browser devtools → Network, confirm `GET …/dicomweb/studies/…/series/…/instances/…` returns a non-zero body starting with the DICOM preamble (`DICM` at offset 128).
4. Unit test: `mvn test -pl ../shanoir-ng-datasets -Dtest=org.shanoir.ng.dicom.web.service.DICOMWebServiceTest` (from `shanoir-ng-back`).

## Test plan

- [x] `DICOMWebServiceTest` passes in CI
- [x] OHIF segmentation export/save to Shanoir succeeds on a test examination
- [x] DICOMweb instance GET returns non-empty `application/dicom` with valid preamble
- [x] Subject name still appears correctly in OHIF when MR-004 anonymization applies


michaelkain:
Dear @jbardet,
excellent work and very clear PR. Tested locally and it works! Could save a segmentation with success.
And after dataset-copy, new name still displayed.
With kind regards, Michael

michaelkain (approved):

PR #3554, opened 2026-05-29
Fix BIDS import examination date from scans.tsv and sessions.tsv.

jbardet:
## Problem:
BIDS upload often fails to get acq_time and thus goes to fallback (today)

## Solution
- Parse BIDS `acq_time` from `*_scans.tsv` and `*_sessions.tsv` with flexible ISO formats
- Use the **earliest** valid `acq_time` in scans.tsv for the examination date (visit date)
- Match session labels with or without `ses-` prefix
- Log and surface a **warning** when falling back to folder creation time


michaelkain (approved):
Dear @jbardet,
I approve by code-review only, as we have very few BIDS imports here in Rennes, and as I assume you have tested your use-cases well, with specific BIDS test files.
Thank you for your great unit-tests, that are very complete. The idea, to extract
the date management into a separate component (BidsTsvDateParser), makes
the code much more readable, excellent.
With kind regards, a big thank you and see you soon, Michael

Issue #2388, opened 2024-08-30
Auto-execute pipeline on new imported data (e.g. SIMS)


michaelkain:
Hi @jcomedouteau,
I hope you are doing fine.
This should be a very interesting extension to work on.
The idea is to configure a pipeline-run in VIP with a specific pipeline,
once the import has been done. So this might be done with a daily
job, that runs and initiates the necessary executions depending on
some schema. With kind regards, Michael
PS: a simple start might be: "on every new exam" run "X"

jcomedouteau:
Hi Michael,
This is an idea I discussed with benjamin too, it would be a great forward step for shanoir!
The question is how we implement it:
- Integrated in a study card ? (instead of renaming/adding metadata, start a process) 
- Add a new "automatic action" in study, with a possibility to filter / manage new datas.

This can be quite simple at the beginning and become something quite huge, we need to think of it to be able to evolve.

michaelkain:
Hi @DuckflipXYZ, please provide me a short todo-list, that you see behind this issue in relation with the PR already existing and how much time it might require you please. Thank you!

DuckflipXYZ:
Hi @michaelkain ! Jean-Come almost did everything ! What's missing is the exec logic (Should we use CRON or deal with the exec directly ? If CRON, what freq ? Or maybe the freq is a parameter ? 

I didn't test it yet, so there is a debug/test stage too. I think a work time of 3 days until a week seems relevant.

PR #3556, opened 2026-05-29
feature#3553: Auto-exec execution templates with offline token authentication

### LINKED ISSUE #3553: Auth issue for auto-exec
Fix the authorization issue for the auto [execution feature](https://github.com/fli-iam/shanoir-ng/pull/2832)

fragkiadamis:
## From #2388

Adds **execution templates**: a study-level configuration that automatically runs a VIP
pipeline on newly imported datasets when the datasets microservice restarts.

A template binds a study to a VIP pipeline with a priority, a set of regex **filters**
(on acquisition / examination / subject fields combined via a boolean `filterCombination`
expression), and the pipeline **parameters** (export format, converter, dataset grouping
level, etc.). On datasets-MS startup, `ExecutionTemplateRunner` detects acquisitions
imported since the last run (tracked by the `import_exec_count` watermark in
`miscellaneous_parameter`), matches them against each study's templates, persists the
matches to `planned_execution` (crash-safe queue), and submits them to VIP ordered by
template priority, with dataset-level locking so dependent runs are serialized.

## Authentication (no connected user at startup, continuation of #2388)

VIP needs a real user token both to start an execution and to call back into Shanoir to
download datasets. Since the runner fires at startup with no HTTP session, we use a
**Keycloak offline token captured per template**:

- When a template is created, the frontend obtains a true **offline token** for the
  current user (via `KeycloakService.getOfflineToken()`, a dedicated silent `check-sso`
  requesting the `offline_access` scope) and stores it on the template
  (`execution_template.offline_token`). Unlike a session refresh token, an offline token
  survives the creator's logout and SSO session expiry.
- At execution time, the datasets MS refreshes it via
  `KeycloakServiceAccountUtils.refreshUserToken` to obtain a live access token, rebuilds the
  security context for the original user, and embeds both the access token and the
  refresh token in the `shanoir://` data URIs handed to VIP.

See `docs/auto-exec-authentication.md` for details.

## Schema changes

- `0073_add_offline_token_to_execution_template.sql`

### Create a template
- [x] Open a study --> **Executions** tab --> create an execution template 
- [x] Pick a VIP pipeline, set a priority, configure parameters (export format, converter, group level)
- [x] Save and verify the row exists in `execution_template`, with `offline_token` populated

### Trigger auto-execution
- [x] Import a dataset into the study (matching the filters, if any)
- [x] Restart the datasets microservice
- [x] In the datasets database, confirm `planned_execution` rows are created
- [x] Confirm `import_exec_count` advanced to the highest new acquisition id

### Verify the VIP execution
- [x] In VIP, confirm the execution moves **In queue --> Running**
- [x] Confirm the execution completes and the `planned_execution` row is removed
- [x] Confirm the resulting processing/monitoring is attributed to the template creator

### Offline token survives logout (the key regression)
- [x] Create a template, then **log the creator out**
- [x] Import a dataset and restart the datasets MS
- [x] Confirm `refreshUserToken` succeeds (no `Session not active` / `invalid_grant`) and the execution runs

### Regression
- [x] Trigger a **manual** VIP execution from the UI; verify it still works end to end

Closes #3553


fragkiadamis:
Hello @DuckflipXYZ 

The authentication for the auto-execution is ready. It's working great.

You can test the offline token by setting up a template. After you import the data , logout, restart the MS. When you see in the VIP that the processing has finished, log back in to verify from the jobs that everything went well.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
```diff
@@ -123,10 +128,12 @@ protected synchronized void manageExecutionsQueue() {
                     //Execution in executor submission has to be final
                     ExecutionInQueue execution = next;
 
-                    // Submit execution
+                    // Submit execution.
+                    // NB: threadExecution manages its own (short) transaction internally; the VIP status
```
DuckflipXYZ: 1. PlannedExecutionManager.132: Ok for processing_resource to be available (committed in db) for vip's download, but I don't understand what's the relation between this and the VIP status polling

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
```diff
@@ -159,37 +166,77 @@ public synchronized void addToInvolvedDatasetIds(List<Long> longs) {
...
+            SecurityContextUtil.initAuthenticationContext("ROLE_ADMIN", username, userId, tokenResponse.getToken());
+        } catch (SecurityException e) {
+            LOG.error("Failed to refresh user token for template {}. Execution aborted.", template.getId(), e);
+            return;
+        }
+
+        // Create the monitoring, processing resources and submit to VIP in a SHORT transaction that commits
+        // before VIP starts downloading. The processing resources (resourceId -> datasets mapping) must be
+        // committed and visible to VIP's /carmin-data/path callback, otherwise the download fails with HTTP 400.
+        final String[] vipIdentifierHolder = new String[1];
```
DuckflipXYZ: 2. PlannedExecutionManager.190: "final String[] vipIdentifierHolder = new String[1];" is kinda weird, as there is no list returned/used at all in this method, in the end this is just string manipulation

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
```diff
@@ -159,37 +166,77 @@ public synchronized void addToInvolvedDatasetIds(List<Long> longs) {
...
                 }
+            });
+        } catch (RuntimeException e) {
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
+            LOG.error("Execution from template {} for {} {} failed.", template.getId(), executionLevel, objectId, cause);
+            return;
+        }
 
-                ExecutionStatus status = ExecutionStatus.RUNNING;
+        // Transaction committed: processing resources are now visible to VIP. Poll status OUTSIDE the transaction
```
DuckflipXYZ: 3. PlannedExecutionManager.213: Maybe a transaction completion is not the best we could do here, as there is no garantee that VIP is slower than the transaction completion. The VIP execution creation HTTP request is performed more or less at the middle of the transaction, VIP can "theorically" starts the jobs and the data download before the transaction is committed. An idea would be to entityManager.flush() just after the processingResourceRepo.save(pr), that way we ensure the resources are visible to VIP at any time.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionServiceImpl.java
```diff
@@ -74,6 +75,14 @@ public class PlannedExecutionServiceImpl implements PlannedExecutionService {
     @Autowired
     private ObjectCodec objectCodec;
 
+    /**
+     * Client the auto-exec offline/refresh token was issued to. Must match the client used by
+     * {@link org.shanoir.ng.shared.security.KeycloakServiceAccountUtils#refreshUserToken} and the one
+     * VIP uses for its own background token refresh, otherwise Keycloak rejects the refresh with invalid_grant.
+     */
+    @Value("${user-token.client.id:shanoir-ng-front}")
```
DuckflipXYZ: 4. PlannedExecutionServiceImpl.83: I don't understand the tag, how do you want to get a token value from applications.yml, env vars or whatever ??? (Here, shanoir-ng-front is a default value, it's not leading at all to the shanoir-ng-front clientId in keycloack)

shanoir-ng-ms-common/src/main/java/org/shanoir/ng/shared/security/KeycloakServiceAccountUtils.java
```diff
@@ -49,11 +49,41 @@ public class KeycloakServiceAccountUtils {
     private String clientId;
     @Value("${service-account.client.credential-secret:'SECRET'}")
     private String clientSecret;
+    @Value("${user-token.client.id:shanoir-ng-front}")
```
DuckflipXYZ: 5. KeycloakServiceAccountUtils.java.52: same as 4

DuckflipXYZ (commented):
6. Global: Do we have to manage offline token lifespan in keycloak settings ? Atm it's 30 days max until it turns deprecated (to find the config in keycloak, go to realm settings, sessions, and scroll down a bit to "offline sessions" section)

At first sight, an option would be to do a runner to reset existing offline tokens every month, maybe by using a CRON or something like

fragkiadamis:
Hello @DuckflipXYZ 

You probably see teh PR being updated but it's not ready yet... I need to test some stuff and fix some more from your comments.  Iwill let you know asap when it will be ready for review.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
fragkiadamis: The new String[1] wasn't really a list - it was a one-element holder to smuggle the vip identifier out of the runInTransaction lambda, since that lambda returns void and Java won't let me reassign a plain local  from inside it.

Agreed it's awkward though - I added a value-returning callInTransaction to TransactionRunner so the lambda just returns the identifier now, and the holder is gone.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
fragkiadamis: I understood the issue now. flush() though, won't actually fix it. flush just pushes the inserts on our own connection, but they're still inside the uncommitted transaction, and VIP's /carmin-data/path callback reads on a separate connection — so under InnoDB's default isolation it still can't see them until we commit. It'd look safe but wouldn't be.

What actually fixes it is ordering: commit the processing resources first, then send the POST to VIP, then update the monitoring with VIP's identifier in a second transaction. That way the lookup rows are guaranteed visible before VIP ever calls back.

fragkiadamis:
> 6. Global: Do we have to manage offline token lifespan in keycloak settings ? Atm it's 30 days max until it turns deprecated (to find the config in keycloak, go to realm settings, sessions, and scroll down a bit to "offline sessions" section)
> 
> At first sight, an option would be to do a runner to reset existing offline tokens every month, maybe by using a CRON or something like

Checked the realm: offline-session idle is 30 days, there's no max lifespan, and refresh-token rotation is off. So a token only dies if a template sits unused for 30+ days. Added a weekly scheduled job that. It refreshes each stored offline token to reset Keycloak's idle timer.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionServiceImpl.java
fragkiadamis: Dropped the @Value. The client is now read straight from the offline token's own azp claim

shanoir-ng-ms-common/src/main/java/org/shanoir/ng/shared/security/KeycloakServiceAccountUtils.java
fragkiadamis: same as 4

fragkiadamis:
Hello @DuckflipXYZ 

I have done the fixes. Have a look please while I test everything in order to move a bit faster with this PR.

DO NOT MERGE though yet, since I am still testing. Generally everything seems good

DuckflipXYZ (approved):
All good :D

PR #3566, opened 2026-06-02
feat/allowing-papaya-on-bids-dataset

jbardet:
Following discussion at: https://github.com/fli-iam/shanoir-ng/discussions/3549

Allowing Papaya Viewer for BIDS dataset so that we can visualize NIfTI files following BIDS structure. e.g. here it works with a NIfTI returned by VIP because Modality Type is not BIDS.

image


michaelkain (approved):

michaelkain:
Hi @jbardet, perfect, thank you!!!

PR #3573, opened 2026-06-04
Turning the clean DB and the constraint addition generic

DuckflipXYZ:
Hi @michaelkain, I switch from a light clean + specific instance deletion (visible in the PR #3568) to a generic clean + constraint addition. 

As it's about deletion, for some objects we can consider it's masse deletion, I prefer to let you check, and evengtually accept or refuse that PR. Deletion are needed to respect the new constraints added. To give you a size order of the number of objects deleted, I precised it at each line of the sql script. (I tried multiple and multiple time to ensure it was consistent). 

From my point of view, everything seems relevant and coherent with what we want/have, so let me know what will happen to that PR :) see you


michaelkain:
Dear @fragkiadamis, I hope you are doing fine. As you know the database model now quite well, could you please test the PR and give a feedback to Guéwen? Thank you and with kind regards, Michael

fragkiadamis:
Hello @DuckflipXYZ 

I tried to execute some of those queries directly in the databse and I noticed the following:

I created the garbage_dataset table in neurinfo and indeed I got 2 garbage datasets because they had a NULL subject_id to the related examination as we can see in the following 4 images

image
image
image
Screenshot From 2026-06-08 14-58-21

The issue I have with this is that the subject_id is not NULL in the dataset table. And indeed if we move to the `studies` database, we will notice that the data are consistent - check the following two screenshots. The subject exists and is assigned to a study.

Screenshot From 2026-06-08 14-59-28
Screenshot From 2026-06-08 15-00-33

So in that case the subject_id those dataset, dataset_acquisition and examination are not left over dangle entities. The only issue is that the examination is not consistent, everything else seems to be fine.

So just by patching up the NULL values, everything works great as you can see in the following screenshot
image

Do we need these datasets to actually be deleted ??

michaelkain:
Hi! Could we have a meeting on this before merge please?

DuckflipXYZ:
Hi @fragkiadamis thanks for that reply. I just based my self on what we see on GUI, and as it's hardly depending on dataset DB (more than studies DB), that's why I proceeeded that way. But y, if all the rest is consistent, we might do your way.

HI @michaelkain , when you want !

michaelkain:
Dear @DuckflipXYZ, dear @fragkiadamis, thank you for the good meeting!
As discussed, please try a fix-before-delete approach to repair the tree, but then clean it up please and in coordination with Céline, thank you!

DuckflipXYZ:
@fragkiadamis Hi ! 

I did a fix to resolve  what you mentioned, it works but sadly only for one case among the 2. The one which is not solvable has no mention of the subject_id neither in datasets.dataset, in dataset.examination and in studies.study_examination tables, so it's kinda dead.

fragkiadamis:
> @fragkiadamis Hi !
> 
> I did a fix to resolve what you mentioned, it works but sadly only for one case among the 2. The one which is not solvable has no mention of the subject_id neither in datasets.dataset, in dataset.examination and in studies.study_examination tables, so it's kinda dead.

@DuckflipXYZ I think that's fine. I do not think it's a problem to delete the completely dead entities. My only issue was on the incosistent entities and not the completely dead.

Let me know when you you have talked with Céline in order to review and merge.

However I will merge only after the upcoming release is deployed on prod in order to avoid any unforeseen consequences.

fragkiadamis:
Hello @chomo-ofsep 

@DuckflipXYZ told me that I should wait an update from you concerning this PR before merging. Let me know when you have the time to have a look at it.

Best,
Adam

chomo-ofsep:
It's all good for me. I've checked the pre and post clean export that @DuckflipXYZ provided me and it's identical. Thanks !

fragkiadamis:
> It's all good for me. I've checked the pre and post clean export that @DuckflipXYZ provided me and it's identical. Thanks !

Ok thanks, i am merging it then.

Have a good day

fragkiadamis (approved):

PR #3574, opened 2026-06-08
feature#3565: Let admins enable/disable Keycloak 2FA (TOTP) from the user form

### LINKED ISSUE #3565: checkbox in Shanoir GUI to activate OTP in Keycloak
as discussed with @michaelkain, to finalise issue 2807, adding a checkbox in Shanoir that when selected sends the information to Keycloack to activate the OTP (similar to the can import from PACS checkbox).

fragkiadamis:
## Summary

Adds an **admin-only** way to enable or disable Keycloak two-factor authentication (TOTP)
for a user, integrated directly into the existing user profile form ? it behaves exactly
like the "Can import from PACS" field (read-only icon in view mode, checkbox in edit mode,
applied on Save).

## How it works

2FA state lives entirely in **Keycloak**, not in the Shanoir database:

- **Enable** ? adds the `CONFIGURE_TOTP` required action to the user, forcing them to
  register an authenticator app at next session refresh.
- **Disable** ? removes the `CONFIGURE_TOTP` required action **and** deletes any existing
  `otp` credential, so users who already set it up actually lose their second factor.
- **Enabled flag** is derived live from Keycloak: `true` if an `otp` credential exists
  **or** the `CONFIGURE_TOTP` required action is set, `false` otherwise.

The flag travels on the existing user GET/PUT rather than a dedicated endpoint, exactly
like `canAccessToDicomAssociation`.

## Testing

- [x] As an admin, edit another user ? tick the Two-factor authentication checkbox --> Save --> `CONFIGURE_TOTP` appears under the user's Required Actions in the Keycloak admin console.
- [x] Reopen the user in view mode ? the field shows a green tick.
- [x] The user logs in, completes OTP setup ? the field still shows a green tick.
- [x] Edit the user --> untick the checkbox --> Save --> the `otp` credential is removed in the console, the user is no longer prompted, and the field shows a red X.

Closes #3565


shanoir-ng-users/src/main/java/org/shanoir/ng/user/service/UserServiceImpl.java
```diff
@@ -339,6 +348,19 @@ public User update(final User user) throws EntityNotFoundException {
         ShanoirEvent event = new ShanoirEvent(ShanoirEventType.UPDATE_USER_EVENT, user.getId().toString(), KeycloakUtil.getTokenUserId(), "", ShanoirEvent.SUCCESS);
         eventService.publishEvent(event);
 
+        // Two-factor authentication is managed in Keycloak and only by admins.
+        if (KeycloakUtil.isAdmin() && user.getTwoFactorEnabled() != null) {
```
michaelkain, line 364: Hi @fragkiadamis, thank you for your PR!
It might be interesting to extend the creation of a new user with role_admin as well for the flag?
What should extend the create(User user) method?

michaelkain:
Hi @fragkiadamis, the PR looks totally promising by code review.
Could you please ask Céline to test it on qualif with an user on your PR?
I would approve, my only maybe interesting point is the creation of a new user, where we might see the flag in the GUI and maybe not propagate it to Keycloak? With kind regards, Michael

michaelkain:
ah, okay, you only display the flag in !create mode, but it might be interesting for an instance server admin, once operating with OTP to add new users with OTP directly? the issue did maybe not foresee this.

fragkiadamis:
Hello @michaelkain 

We can enable the 2FA in the user creation too now.

michaelkain (approved):
Hi @fragkiadamis, thank you for this PR and your corrections.
I approve by code review only. Please ask Céline to have a test on qualif, starting from tomorrow. With kind regards, Michael

PR #3598, opened 2026-06-11
Fix for MultiframeExtractor (Enhanced DICOM)

michaelkain:
**Problem**
This PR fixes the following exception, that can today occur in MS Import and MS Datasets:
**Unsupported SOP Class: 1.2.840.10008.5.1.4.1.1.4.3** [org.dcm4che3.emf.MultiframeExtractor.implFor(MultiframeExtractor.java:268), org.dcm4che3.emf.MultiframeExtractor.extract(MultiframeExtractor.java:261),

**Background**
Version 5.31.1 and even latest version 5.34.3 of dcm4che library do not support the above SOP Class:
**Enhanced MR Color Image Storage 1.2.840.10008.5.1.4.1.1.4.3**
What leads to the above exception.

**Solution**
To avoid the above exception we use MultiframeExtractor only with SOP classes, that are supported.
Other SOP classes can be imported and stored in Shanoir, only their meta-data extraction remains limited,
but the DICOM files are fully stored.


michaelkain:
Successfully tested locally with corresponding data, import and download worked.

PR #3599, opened 2026-06-11
Pr 3471 bis

DuckflipXYZ:


DuckflipXYZ:
@a-ba Hi Anthony ! 

Tests are fixed, but keycloak access from web interface as from other ms is not working at all

a-ba:
> but keycloak access from web interface as from other ms is not working at all

In which environment is it broken? (dev,  qualif, ...?)

DuckflipXYZ:
> > but keycloak access from web interface as from other ms is not working at all
> 
> In which environment is it broken? (dev, qualif, ...?)

In local only, it works well on remote (at least on qualif)

a-ba:
Ok this is tricky.

Your immediate issue is the value of SHANOIR_KEYCLOAK_INTERNAL_URL, which should be `http://keycloak:8080/auth`. This will fix the communication between keycloak and shanoir.

But there is a second issue : the PR also migrates the keycloak hostname v2 config which introduces strict hostname checking. This breaks the access to the keycloak admin console at `http://localhost:8080/auth/admin/`.

I pushed two commits at https://github.com/a-ba/shanoir-ng/tree/keycloak-fixes to address these issues.

DuckflipXYZ:
Ok, all good ! 

Par contre t'as oublié de push nginx/shanoir.template.keycloak-full.conf et nginx/shanoir.template.keycloak-limited.conf o.o j'aurais bien tenter de les composer moi-même mais je bite rien .. 

A+

a-ba:
oups

je viens de les ajouter

docker-compose-dev.yml
```diff
@@ -301,11 +305,12 @@ services:
       - SHANOIR_PREFIX
       - SHANOIR_URL_SCHEME
       - SHANOIR_URL_HOST
+      - SHANOIR_KEYCLOAK_INTERNAL_URL
       - SHANOIR_MIGRATION
       - SHANOIR_STORAGE_TYPE
       - S3_ENDPOINT
       - S3_REGION
-      - S3_ACCESS_KEY_ID
+      - S3_ACCESS_KEY_IDdev-front
```
a-ba: this looks like a glitch

a-ba (approved):

PR #3602, opened 2026-06-12
feature#3125: Allow attaching extra-data files to dataset acquisitions

### LINKED ISSUE #3125: Arbitrary file import & data relationship
This proposal aims to enable the ingestion of MRI raw measurement data—whether in proprietary formats or open standards like ISMRMRD—into Shanoir. These raw datasets are the precursor of reconstructed images (usually in DICOM format) and spectroscopic files.

Many technical and methodological research projects are using these type of data for improving reconstruction methods or are using prototype reconstruction methods.  

Enabling arbitrary file ingestion in Shanoir highlights the need for data lineage—tracking relationships between files throughout their lifecycle. By implementing this feature, we can establish clear links between raw data and processed outputs (e.g., reconstructed images), enhancing the platform role in research reproducibility.

This data lineage concept extends to existing Shanoir files. For example, in MP2RAGE contrast processing, the two inversion-time images used to generate the final uniform denoised image could be linked using this feature.

fragkiadamis:
Adds the ability to attach arbitrary "extra-data" files (PDFs, RAW images, etc.)
directly to a **dataset acquisition**, mirroring the existing examination extra-data
feature.

Closes #3125


DuckflipXYZ (approved):
Hey ! Some points : 
- maybe it should not, but files with same name are not managed
- maybe fix a file size limit ? 
- would be great to refresh only the acq instead of the whole tree when updating the acq, but since the partial tree refresh method does not exist yet, I guess it would be in an other issue/PR

PR #3612, opened 2026-06-17
change shanoir tutorial link and switch to storage volume TB if needed

LaurentPV:
This PR is not linked to an issue as it is very small.

it corrects the "open tree/close tree" tooltip of the study tree that was inverted.

it changes the unit of total storage volume to Terabyte when the value is superior to 1000 Gigabyte.

It updated the link to Shanoir Tutorial in the Home Documentation menu.


fragkiadamis (approved):
lgtm

PR #3617, opened 2026-06-22
fix null file & download links [Celine bugs]

julien-louis:
* fix download "null" file : content disposition header must not have spaces and must have double quotes
* fix event download bad management


fragkiadamis (approved):
Tested locally and works!

PR #3618, opened 2026-06-22
#2768 : study history does now take into account subject deletion

DuckflipXYZ:

PR #3620, opened 2026-06-23
fix(subject): update `@JsonIgnoreProperties` to fine-tune property handling

### LINKED ISSUE #3608: Import failure
During DICOM dataset importation:

- I create a new subject
- I save the new subject
- The process fails with the following message

Image

The bug was was spotted in neurinfo-qualif

fragkiadamis:
## Problem
  
When importing a DICOM dataset, the subject is created in MS Studies but fails
to propagate to MS Datasets, with the import breaking on: 

jakarta.validation.ConstraintViolationException: Validation failed for classes
[org.shanoir.ng.shared.model.Subject] during persist time ...
propertyPath=study, interpolatedMessage='must not be null'

## Root cause

`SubjectMapper.subjectToSubjectDTO` populates `studyId` but leaves the `study`
(`StudyTagsDTO`) field null (`@Mapping(target = "study", ignore = true)`). Jackson
serializes fields in declaration order, so the message sent to MS Datasets looks like:

```json
{ ..., "studyId": 123, ..., "study": null, ... }
```

The MS Datasets Subject entity has both setters. During deserialization they
fire in JSON order:
  
1. setStudyId(123) -> resolves and sets study
2. setStudy(null) -> overwrites it back to null

Fix
  
Ignore the redundant study property on input in the MS Datasets Subject,
making studyId the sole entry path so it can no longer be clobbered:

@JsonIgnoreProperties(value = "study", ignoreUnknown = true, allowGetters = true)

Closes #3608

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ server directly:
    `SHANOIR_KEYCLOAK_USER`/`SHANOIR_KEYCLOAK_PASSWORD' (default is `admin`/`&a1A&a1A`)
 3. go to the **shanoir-ng** realm
 4. create/edit the new user and grant the relevant role (eg. `ROLE_ADMIN`). By
-   default, new user accounts are created in Keycloak by the users microservice
+   default, new user accounts are created in Keycloak by the users' microservice
    with temporary passwords, you may reset the password in keycloak's admin
    interface and receive the new password is by e-mail. In development, if you
    do hot have a configured SMTP relay, then you may choose to overide the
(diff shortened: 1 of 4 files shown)


fragkiadamis:
Hello @DuckflipXYZ 

This PR fixes an error that was introduced from a PR that I reviewed. It is necessary to be merged in order for me to be able to test your PR #3618 

So please have a look and review it in order to move on with the review on your PR.

Bye!

README.md
```diff
@@ -324,7 +324,7 @@ server directly:
    `SHANOIR_KEYCLOAK_USER`/`SHANOIR_KEYCLOAK_PASSWORD' (default is `admin`/`&a1A&a1A`)
 3. go to the **shanoir-ng** realm
 4. create/edit the new user and grant the relevant role (eg. `ROLE_ADMIN`). By
-   default, new user accounts are created in Keycloak by the users microservice
+   default, new user accounts are created in Keycloak by the users' microservice
```
DuckflipXYZ, line 327: is that a miss  ?

DuckflipXYZ (approved):
All good by my side

README.md
fragkiadamis: https://wallstreetenglish.fr/fiches-anglais/grammaire/apostrophe-s-possession

DuckflipXYZ: LEL

for me, "users" is a name, not an owner, but whatever

PR #3626, opened 2026-06-24
rewrite nginx expiration rules for static content

a-ba:
We experienced multiple caching issues while testing #3599 (eg: env.js would take too long to be refreshed), also the current config does not appear consistent (the *reduced* lifetime is actually longer that the default lifetime).

This commit applies the following changes:

- use a short lifetime for files that may break the app if not up-to-date (5mn is enough to guarantee that the file is refreshed across a reboot of shanoir), this covers:
  - html/js/css files that do not have a fingerprint in their name
  - the default page (index.html)
- use a longer lifetime for images and fonts (no a big deal if they are not up-to-date)
- use a very long lifetime for fingerprinted files produced by esbuild (they never need to be refreshed)
- serve the default page (index.html) through a named location so that the short lifetime is applied event when served by default
- remove the *index* directive (which has no effects here)


DuckflipXYZ (approved):
all good

Issue #3615, opened 2026-06-19
Welcome page: statistics 0 query run


michaelkain:
Hi @LaurentPV,
I hope you are doing fine.
In prod, it takes a long time to load the real number of queries, until loading we display 0 queries,
what gives the images as people go fast over the page, that the platform is not used, what is sad.
Could you please replace the 0 query, with a loading text, while the query runs?
With kind regards, Michael

PR #3636, opened 2026-07-01
last events query improved

LaurentPV:
linked to https://github.com/fli-iam/shanoir-ng/issues/3615

This PR improves the performance of the countByLastUpdateAfter(Date expiryDate) query that is used to retrieve users events count on the welcome page.

field last_update on table events is now indexed and query was improved in ShanoirEventRepositoryImpl.java

Performance was improved from 2,2s to 50ms


michaelkain:
Hi @LaurentPV, thank you for this PR and for improving the performance of the query, what is always good.
My issue was less the time, but more the actual display, that gives a first impression: I do not see the most important change for me on the PR to change from "no" query to something like "loading" query. With kind regards, Michael

LaurentPV:
Ok @michaelkain , now the sentence "and X querie(s) were run during the last month." is shown only when eventsCount is loaded, which is pretty fast now.

michaelkain (approved):

michaelkain:
Perfect, thank you @LaurentPV !

PR #3641, opened 2026-07-02
issue/3640: Fix study deletion

### LINKED ISSUE #3640: Study deletion does not work
When you delete a study, the study never actually gets deleted. It temains in the list, it remains accessible, and it appears in the DB.

No error message.

fragkiadamis:
Deleting a study returned success but left the study in the database, with no error shown.

The deleteStudy endpoint ran its guard checks, cleaned up DUAs/storage and published DELETE_STUDY_EVENT, but never called studyService.deleteById ? so the row was never removed. The catch block also returned 204 No Content for any failure, masking the problem.

Fix
- Call studyService.deleteById(studyId) before storage cleanup, so a DB failure rolls back cleanly.
- Return 404 on not-found and 500 on error instead of a silent 204.

Closes #3640


DuckflipXYZ (approved):
All good

PR #3642, opened 2026-07-02
issue/3579: Fix OHIF data loading issues

### LINKED ISSUE #3579: [1] Reliable visualization of saved OHIF segmentations in the same DICOM study
## Problem:
Users can create and save segmentations in Shanoir (OHIF → DICOM SEG → PACS + SegmentationDataset in DB), but cannot reliably view them again on the source images in the same examination. Review often requires downloading DICOM and using an external tool. 

## Why: 
- Save works: Export SEG → STOW-RS → PACS + SegmentationDataset, linked to the source dataset. The SEG reuses the parent acquisition in the DB (no new acquisition for the SEG series). 
- Reload fails: OHIF reloads via DICOMweb GET .../studies/{examination}/series. Shanoir queries the PACS, then filters the series list: 
  - Allowed list = DatasetAcquisition.sortingIndex only → series numbers of imported image series (MR/CT/…), set at import. 
  - Filter = keep only series whose DICOM SeriesNumber (0020,0011) is in that list (DICOMWebApiController.filterAndSortSeriesBySeriesNumber). 
  - SEG series are excluded: each export creates a new PACS series with its own SeriesNumber. OHIF often assigns a high number (e.g. 99, or 1 + source series number) — not in the allowed list. Shanoir does not set 99 in code; the mismatch is filter vs export behavior. 
  - No SEG exception (modality SEG is not kept). OHIF never sees those series → overlays do not come back. 
  - No UI to open a Segmentation dataset in the viewer; study-tree “View DICOM” opens acquisitions (source images), not SEG datasets. 

## Proposal (MVP):  
- Backend: Expose saved SEG in DICOMweb for the examination — e.g. always include Modality == SEG in series results, and/or add SEG series numbers from SegmentationDataset / PACS paths to the allow-list. 
- UI: “View segmentations” on examination; “View on source image” on segmentation datasets → OHIF segmentation mode with source volume + all linked SEGs loaded. 
- Done when: After several exports on one series, reopening the examination shows all segmentations on the correct images, with normal study rights.

fragkiadamis:
## Problem

1. Segmentations could not be visualized in the OHIF viewer
2. No matter the scope of the viewer (examination/acquisition), all the data of a subject were loaded always

## What this PR does

**Series scoping and SEG/SR inclusion**
- The DICOMWeb facade now returns exactly the series belonging to the examination/acquisitions the viewer was opened for, instead of everything the PACS holds for the subject.
- SEG/SR series attached to a requested acquisition are automatically included, so saved segmentations reappear on their source images after reopening the examination.

**Series UID virtualization**
- OHIF never sees real Study/SeriesInstanceUIDs. The facade rewrites UIDs both ways:
  - acquisition series --> `ROOT.<acquisitionId>`
  - dataset series (e.g. SEG exports, processing outputs) --> `ROOT.0.<datasetId>`
- Incoming requests with virtual UIDs are resolved back to the real PACS UIDs (`SeriesInstanceUIDHandler`), with an exam-level series map cached (30 s TTL) to avoid re-querying the PACS on every request.

**Bulkdata proxying**
- BulkDataURIs in metadata responses are rewritten to the public facade URL (new `viewer.ohif.url.base` property in `application.yml`) with virtual UIDs, and a new endpoint proxies the bulkdata bytes from the PACS. This fixes OHIF request failures on pixel data.

Closes #3579


fragkiadamis:
Hello @michaelkain 

The PR is ready for review.

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dicom/web/DICOMWebApi.java
```diff
@@ -148,6 +148,22 @@ ResponseEntity findFrameOfStudyOfSerieOfInstance(
...
 
+    @Operation(summary = "", description = "Returns the bulkdata of a DICOM instance, e.g. overlay data, of a study and a serie")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "found bulkdata"),
+            @ApiResponse(responseCode = "404", description = "no bulkdata found"),
+            @ApiResponse(responseCode = "401", description = "unauthorized"),
+            @ApiResponse(responseCode = "403", description = "forbidden"),
+            @ApiResponse(responseCode = "500", description = "unexpected error") })
+    @GetMapping(value = "/studies/{examinationUID}/series/{serieInstanceUID}/instances/{sopInstanceUID}/bulkdata/**")
+    @PreAuthorize("hasRole('ADMIN') or (hasAnyRole('EXPERT', 'USER') and @datasetSecurityService.hasRightOnExamination(#examinationUID, 'CAN_SEE_ALL'))")
```
michaelkain: Hi @fragkiadamis, thank you for this great PR.
As we return the actual pixel data image here, as far as I know, I would require CAN_DOWNLOAD rights, because it would actually allow to download the pixel data and put them somewhere else?
With kind regards, Michael

michaelkain: I see, that this should be adapted on some other methods too, to align with the idea

michaelkain:
Hi @fragkiadamis, very good PR by the way.
As discussed, please test with DicomEnhanced and see if a findBulkDataOfStudyOfSerieOfInstanceOfFrame might be required. I will test locally. With kind regards, Michael

michaelkain:
Hi @fragkiadamis, thank you for having added the unit tests.
One point to note please, unit tests are not integration tests via ShUp, that drive the REST-API from outside.
I can understand, that at some point we should finish this PR, but I think there was a misunderstanding on the type of test to propose. Anyway, any test helps and with kind regards, Michael

fragkiadamis:
Hello @michaelkain everything is tested on local and qualif. You can have a look again.

michaelkain:
Hi @fragkiadamis, thank you a lot for this complexe and challenging PR.
My local tests worked fine, I played quite a lot with segmentations and all looks fine.
With kind regards, Michael

michaelkain (approved):

PR #3645, opened 2026-07-03
Removing Login policies updated from logins screen

DuckflipXYZ:


fragkiadamis (changes requested):
Also remove the line `updatePasswordPolicies=For security reasons the password policies have evolved.` from the file `/home/afragkia/Workspace/shanoir-ng/docker-compose/keycloak/themes/shanoir-theme/login/messages/message_en.properties`

fragkiadamis (approved):

PR #3656, opened 2026-07-10
#3651: set up mail supplier black list + related env var

### LINKED ISSUE #3651: Create a env Var to black/whitelist gmail or similar mail for account request
All in title, the idea is to use (or not) a list of mail supplier as a blacklist one, and to invalidate the mail field according to it. It has to be related to a env var that enable/disable its usage

DuckflipXYZ:


a-ba:
It seems weird (and less flexible) to hardcode the lists in the sources. Why not just have the blacklisted domains listed directly in the environment variable? eg like the SHANOIR_ALLOWED_ADMIN_IPS env var

```
# SHANOIR_ALLOWED_ADMIN_IPS: comma-separated list of IP addresses or networks
#   eg: SHANOIR_ALLOWED_ADMIN_IPS=192.0.2.1,2001:db8:1::/64
SHANOIR_ALLOWED_ADMIN_IPS=
```

a-ba:
also the name of the var should be prefixed with `SHANOIR_` (to be consistent with the other vars)

PR #3662, opened 2026-07-15
S3 for dcm4chee

AxelCouturierCIBM:
Add support for dcm4chee s3 on 1st startup of the container


jbardet:
@michaelkain we tested the S3 integration in one of the CIBM institutions. It worked perfectly with those small changes made in this PR. Let us know if you need anything else from our side.

jbardet:
@michaelkain @fragkiadamis @a-ba Could you have a look at this?

michaelkain (changes requested):
Dear @AxelCouturierCIBM, I hope you are doing fine and thank you for this PR!
When I review your PR, I have the impression, maybe I am wrong, that it is a hard
switch to S3 for dcm4chee-arc, without any flag, that is set in the .env to make a switch. E.g. as on our local dev machines, we normally have not S3 and neither have all our production instances, it would be great to have a switch somewhere in .env, to activate or deactivate it. file-system for dicom storage should be the default. With kind regards, Michael

jbardet:
Hi @michaelkain , the PR already had opt-in logic in ldap-first-start.sh (it only changes LDAP when storage env vars are set), but the defaults turned S3 on anyway: STORAGE_DIR: s3://mysupertest in docker-compose.yml, plus LDAP_STORAGE_URI=s3://test and active S3 properties in variables.env.

I fixed that by removing the hardcoded S3 override, clearing LDAP_STORAGE_URI, and commenting out the S3 property examples. Filesystem (/storage/fs1) is now the default; S3 is opt-in via variables.env when someone sets both STORAGE_DIR and LDAP_STORAGE_URI to an s3://… URI. Local dev and non-S3 deployments stay unchanged unless configured explicitly.

michaelkain (approved):
Hi @jbardet, tested, all good, cheers, Michael

PR #3683, opened 2026-07-20
issue/3543: Refresh Keycloak token before submitting VIP execution

### LINKED ISSUE #3543: [4] VIP execution launch should refresh Keycloak tokens before embedding them in shanoir:/ URIs
## Summary

After keeping the Shanoir UI open for a long time without reloading, users can still navigate the application, but launching a VIP pipeline fails on the VIP side during Shanoir input download with a Keycloak token refresh error. A hard refresh often forces logout; logging in again and submitting quickly works.

The VIP job embeds the user's `token` and `refreshToken` in `shanoir:/...` URIs at submission time. If those tokens are stale or the Keycloak session is no longer active, VIP cannot download datasets from Shanoir.

## Steps to reproduce

1. Log in to Shanoir and leave the tab open for a long period (no F5 / no navigation that renews the session).
2. Continue using the UI (browsing still appears to work).
3. Select datasets and run a VIP pipeline (e.g. deface pipeline) without re-authenticating.
4. Observe VIP job failure; optionally hard-refresh Shanoir → often redirected to login; re-login and submit again quickly → job can succeed.

## Actual behavior

VIP workflow fails during `inputs_download` before the pipeline container runs.

From VIP `workflow.err` / `sh.err`:

```text
TOKEN_REFRESH_ERROR - error while refreshing the token: exit=0, status=400, message: Token is not active
Exiting with return value 61
```
Also seen:
```text
TOKEN_REFRESH_TOO_LONG - Token refreshing is taking too long. Aborting the process.
Exiting with return value 60
```
app.out and app.err for the job are empty (failure before container execution).

## Expected behavior
Before building VIP execution inputs and embedding token / refreshToken in Shanoir URIs, the frontend (or backend at launch time) should:

1. Force a Keycloak token refresh (e.g. keycloak.updateToken(minValidity) or equivalent).
2. If refresh fails, block pipeline launch and show a clear message (e.g. "Session expired — please log in again").
3. Only submit to VIP with tokens that Keycloak will accept for grant_type=refresh_token on the token endpoint.

## Suggested fix
- In the VIP execution launch path (shanoir-ng-front and/or datasets VIP services), refresh tokens immediately before creating ExecutionCandidate / VIP inputs.
- Disable "Run" when refresh fails; prompt re-login.
- Optionally surface token/session expiry in the UI before the user invests time configuring a pipeline.

## Additional context (deployment)
Keycloak admin may also tune SSO session idle / refresh token lifespan for shanoir-ng-front, but that does not replace a refresh-before-submit check.

fragkiadamis:
The refresh token embedded in the VIP shanoir:/ input URIs was captured at form load and never refreshed, so a stale or rotated token caused VIP to fail during inputs_download with "Token is not active".

This forces a Keycloak token refresh at submit time (updateToken(-1)), embedding a freshly-rotated token and validating the session. If the session has expired, the launch is blocked with a "Session expired -> please log in again" message instead of submitting a job VIP can't authenticate.

Closes #3543


DuckflipXYZ (approved):
Can not test it atm, (and dunno how to test it) but looks good by code review

PR #3690, opened 2026-07-21
docs: document the keycloak URL variables

jbardet:
SHANOIR_KEYCLOAK_URL, SHANOIR_KEYCLOAK_INTERNAL_URL, SHANOIR_KEYCLOAK_ADAPTER_MODE and SHANOIR_KEYCLOAK_API were missing from the configuration table in the README, which documented only SHANOIR_KEYCLOAK_USER/PASSWORD. The only description of them lived in .env comments.

Add them to the table, and add a note explaining why Keycloak needs two URLs (browsers and containers do not reach it the same way) and what the public one means for token validation -- the issuer has to match exactly, which is the detail that makes or breaks a non-default setup.

The note also spells out what pointing these at an external Keycloak actually entails beyond changing the URLs: importing the realm, installing the shanoir-ng-post-auth provider, updating client redirect URIs, and two-way network reachability.


fragkiadamis (approved):
Thanks for this PR @jbardet

PR #3702, opened 2026-07-27
fix(users): canImportFromPACS Keycloak attribute becomes the literal string null when canAccessToDicomAssociation is unset

jbardet:
## Summary

`canImportFromPACS` is written to Keycloak as the literal string `"null"` whenever a user is created without `canAccessToDicomAssociation` in the request body. Omitting that field is legal — it is a nullable `Boolean` with no `@NotNull` and no default — so this is reachable from the front end, from scripts, and from any integration.

This PR applies the null-safe pattern the codebase already uses for this exact field elsewhere.

## The defect

Both call sites that build the Keycloak attribute concatenate the field directly onto an empty string:

| File | Line | Code |
|---|---|---|
| `shanoir-ng-users/.../user/utils/KeycloakClient.java` | 339 | `attributes.put("canImportFromPACS", Arrays.asList("" + user.isCanAccessToDicomAssociation()));` |
| `shanoir-ng-users/.../ShanoirUsersManagement.java` | 210 | `attributes.put("canImportFromPACS", List.of("" + user.isCanAccessToDicomAssociation()));` |

`isCanAccessToDicomAssociation()` returns the raw `Boolean` (`User.java:184`, no null-coalescing). String concatenation with a `null` object is well-defined Java — it calls `String.valueOf(null)` — so the attribute Keycloak actually stores is the four-character **string `"null"`**, not a boolean and not the field's absence.

This is deterministic. It does not depend on Keycloak version or configuration.

## Why it matters

The create-user call returns success, so nothing anywhere in the chain reports a problem. But any consumer that later parses `canImportFromPACS` as a boolean (`"true"` vs anything else, or an explicit `"true"`/`"false"` check) will silently treat that user as **not** allowed to import from PACS, regardless of intent — and the value keeps reading `"null"` until an explicit update overwrites it.

Silent, permanent, and invisible to the caller.

## The fix

The codebase already has the correct null-safe pattern for this same field, on the update path:

```java
// UserServiceImpl.java:446
userDb.setCanAccessToDicomAssociation(
    user.isCanAccessToDicomAssociation() != null && user.isCanAccessToDicomAssociation());
```

This PR applies that same expression at the two attribute-building call sites — the only places in `shanoir-ng-users` that were missing it. An unset field now correctly becomes `"false"` instead of `"null"`.

A stricter alternative — rejecting the create request outright when the field is absent — is also defensible, but it changes the API contract for existing clients that legitimately omit the field. The null-safe default matches what the update path already does today and is the smaller, backward-compatible change. Happy to switch approach if maintainers prefer the strict one.

## Verification

- `mvn -f shanoir-ng-parent/pom.xml -pl ../shanoir-ng-users -am compile` → **BUILD SUCCESS** (JDK 21, Maven 3.9.12).
- Behavioural check: `POST /users/users` with `canAccessToDicomAssociation` omitted from the JSON body, then read the resulting user's `canImportFromPACS` attribute in Keycloak (Admin Console or `kcadm.sh get users/<id>`). Before this change it reads `null`; after, `false`.
- Both call sites confirmed to be the only ones in the module: `grep -rn "isCanAccessToDicomAssociation" shanoir-ng-users/src/main` returns exactly these two plus the getter and the already-correct `UserServiceImpl` update path.

## How this was found

Found by a downstream release-validation pipeline (CHUV/UniGE Shanoir deployment) during `NG_v2.12.1` → `NG_v2.13.0` upgrade testing on 23 July 2026, and reproduced on every subsequent run. A CI functional test that omitted this optional field triggered a trace into `KeycloakClient.java`.

The string-concatenation defect itself is confirmed directly from the source above and is independent of that pipeline. Whether the *specific* CI symptom on 23 July was Keycloak rejecting the literal `"null"` value versus some unrelated cause in the chain has not been isolated separately from this fix, and is not claimed here.

I could not find an existing issue or PR covering this; happy to open a tracking issue instead if that fits your workflow better.


michaelkain:
Hi @jbardet, I hope you are doing fine and thank you for this PR!
I wonder, if setting private Boolean canAccessToDicomAssociation =  Boolean.FALSE; might be a more elegant solution, to keep the solution already inside User? So not != null checks necessary outside anymore. With kind regards, Michael

jbardet:
Thanks @michaelkain. Defaulting canAccessToDicomAssociation to Boolean.FALSE (and adding @NotNull like the other flags) is cleaner and keeps the invariant on User instead of at every Keycloak sync site. I’ve updated the PR accordingly. For existing rows with NULL in the DB, I didn't add a migration: a null-safe getter and setter on the entity is enough to keep loaded entities consistent when syncing to Keycloak, without touching the database.

michaelkain (approved):
Dear @jbardet, perfect, thank you a lot!

PR #3704, opened 2026-07-28
Fix RTSTRUCT/RTDOSE/RTPLAN import failing with a NullPointerException on missing acquisition date

jbardet:
## Problem

Importing an RTSTRUCT fails with a `NullPointerException`, so RT structure sets can never be imported.

`DicomUtils.checkSerieIsIgnored` explicitly accepts `RTSTRUCT`, `RTDOSE` and `RTPLAN`, so the import proceeds. `DatasetAcquisitionContext.selectModalityStrategy` only has a dedicated strategy for MR, CT, PT and XA, so these modalities fall through to `GenericDatasetAcquisitionStrategy`, which did:

```java
datasetAcquisition.setAcquisitionStartTime(
        LocalDateTime.of(DateTimeUtils.pacsStringToLocalDate(attributes.getString(Tag.AcquisitionDate)),
        DateTimeUtils.stringToLocalTime(attributes.getString(Tag.AcquisitionTime))));
```

`AcquisitionDate`/`AcquisitionTime` (0008,0022 / 0008,0032) are not defined by the RT IODs, so they are always absent. Both `DateTimeUtils` helpers return `null`, and `LocalDateTime.of` does `Objects.requireNonNull(date)` → `NullPointerException`, which fails the whole import.

The generic strategy was the only one doing this. MR, CT, PET and XA all read the acquisition time through the null-safe `DicomProcessing.parseAcquisitionStartTime`, which is why they do not crash.

This was reported against a real cohort: 6478 RTSTRUCT files across 1135 patients, multiple vendors, 2012–2021, with `AcquisitionDate`/`AcquisitionTime` and `ContentDate`/`ContentTime` absent in 100% of files. The data is standards-conformant; the missing tags are supposed to be missing.

## Fix

- `GenericDatasetAcquisitionStrategy` now goes through `DicomProcessing.parseAcquisitionStartTime`, like the four other strategies, and falls back to the closest equivalent timestamps when the acquisition date/time is absent:
  1. `AcquisitionDate`/`AcquisitionTime` (unchanged, still preferred)
  2. `StructureSetDate`/`StructureSetTime` (3006,0008 / 3006,0009) for RTSTRUCT
  3. `ContentDate`/`ContentTime` for the other non-image IODs
  4. `SeriesDate`/`SeriesTime` as a last resort

  The acquisition start time stays optional: a serie with no usable date/time gets `null` rather than failing the import.

- `DicomProcessing.parseAcquisitionStartTime` is hardened as well. A DICOM date/time tag can be present but empty (type 2), in which case both helpers return `null` and `LocalDateTime.of` threw an NPE that the existing `catch (DateTimeParseException)` did not cover. This affected all five strategies, not just the generic one, and it also occurs in the reported cohort (18 files with an empty `StructureSetDate`).

## Tests

Adds `GenericDatasetAcquisitionStrategyTest`, covering the RTSTRUCT fallback, absent date/time, empty (type 2) date/time, an unparsable time, the precedence of `AcquisitionDate`/`AcquisitionTime`, and the `SeriesDate`/`SeriesTime` fallback.

It is a plain JUnit test with no Spring context, so it runs without any infrastructure.

Verified to reproduce the bug: with the two main-code changes reverted, 5 of the 6 tests fail with `java.lang.NullPointerException: date`. With the fix, all 6 pass.

```
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
```

Note: `ImporterServiceTest` and `MrProtocolStrategyTest` fail in my local environment both with and without this change (`Could not resolve placeholder 'SHANOIR_KEYCLOAK_INTERNAL_URL'` — the Spring context needs infrastructure I do not have locally), so they are unaffected by this PR.


michaelkain:
Dear @jbardet, thank you for this complete PR!! Sorry, but would you mind introducing a RtDataset and a corresponding strategy, that takes care of the problem you have and clearly identify the Datasets?
I think, it is a good idea to solve the NPE in general, even on Generic, but for the future, it would be great
to already set up a specific RtStrategy to extract all meta-data wanted? Now the Generic defaults to some
aspects, that only Rt has. With kind regards, Michael

jbardet:
@michaelkain thank you for the review, I agree. I have updated the PR accordingly:

- DicomProcessing.parseAcquisitionStartTime still has the null-safety fix (empty type-2 date/time tags no longer cause an NPE). This applies to all modalities, including Generic.
- GenericDatasetAcquisitionStrategy is simplified again: it only reads AcquisitionDate/AcquisitionTime, like CT and XA. No RT-specific fallbacks remain there.
- New RT import path: RtDataset, RtDatasetAcquisition, RtDatasetStrategy, and RtDatasetAcquisitionStrategy. RTSTRUCT, RTDOSE, and RTPLAN are routed in DatasetAcquisitionContext to this strategy instead of Generic.
- RT acquisition start time uses DicomProcessing.parseRtAcquisitionStartTime with the fallbacks we need (StructureSetDate/Time for RTSTRUCT, then ContentDate/Time, then SeriesDate/Time). Import still succeeds with null when no usable timestamp exists.
- DB migration 0078_create_table_rt_dataset.sql plus the usual enum/JSON/mapper registrations.
- RT datasets are now explicitly typed as Rt / RT_DATASET, which should make it straightforward to extend RT-specific metadata extraction later without polluting Generic.

michaelkain:
Hi @jbardet,
thank you for this great PR! I tested it and the import worked, what is already perfect.
2 potential optimizations:
- The newly created dataset does not show up after import in the Solr search, what might require an extension here: ShanoirMetadataRepositoryCustom, so RtDatasets will become searchable.
- Normally I updated the front, but when I click in the tree on RtDatasetAcquisition or RtDataset I get this frontend error Error: Received dataset acquisition has no valid "type" property.
With kind regards,
Michael

michaelkain:
Error: Received dataset acquisition has no valid "type" property
    getNewDAInstance dataset-acquisition.utils.ts:36
    getEntityInstance dataset-acquisition.service.ts:50
    mapEntity dataset-acquisition.service.ts:54
    _n entity.abstract.service.ts:112
    Angular 22
    RxJS 63
    Z Angular
    ge RxJS
    get entity.abstract.service.ts:111
    e entity.component.abstract.ts:233

jbardet:
Hi, thanks @michaelkain. I did the improvements, now import works end-to-end (tested with real data). Also fixed RT_QUERY to include qualityTag (required for Solr indexing after the quality-tag migration).

michaelkain (approved):
Hi @jbardet, PERFECT!!! Thank you!

PR #3708, opened 2026-07-30
Accept the full DICOM TM range in DateTimeUtils.stringToLocalTime

jbardet:
## Problem

The DICOM VR TM is `HH[MM[SS[.FFFFFF]]]`, with one to six fractional digits, and values are padded with a trailing space to an even length. `DateTimeUtils.stringToLocalTime` only tried two patterns:

```java
DateTimeFormatter fullFormatter = DateTimeFormatter.ofPattern("HHmmss.SSSSSS");
DateTimeFormatter shortFormatter = DateTimeFormatter.ofPattern("HHmmss");
```

So legal values raise a `DateTimeParseException`:

| Value | Legal TM | Before |
|---|---|---|
| `120000.00` | yes, 2 fractional digits | throws |
| `120000.5` | yes, 1 fractional digit | throws |
| `1221` | yes, HHMM | throws |
| `12` | yes, HH | throws |
| `1221 ` | yes, padded to even length | throws |

The consequences differ by caller. `DicomProcessing.parseAcquisitionStartTime` catches `DateTimeParseException` and returns null, so the MR, CT, PET and XA strategies silently lose the acquisition time on such a value. `GenericDatasetAcquisitionStrategy` inlines the same call with no guard, so the import fails outright.

This came up while investigating RTSTRUCT import failures: `120000.00` is a common value in the wild, and the reporter identified the parser range as the second of the two ways that line can fail.

## Fix

Replace the two patterns with a single formatter built from optional sections covering the whole TM range, defaulting the components that are not present to 0, and trim the padding.

Values that are not legal TM still raise a `DateTimeParseException`, so callers keep being able to tell a malformed value from an absent one. A blank value is now treated like an empty one and returns null, consistent with a type 2 tag being present but empty.

The change is strictly more permissive: everything that parsed before parses identically.

## Scope

`stringToLocalTime` has two production callers, both in `shanoir-ng-datasets`:

- `DicomProcessing.parseAcquisitionStartTime` — catches the exception, unaffected
- `GenericDatasetAcquisitionStrategy` — stops failing the import on a badly formatted time

Neither relies on the exception to detect an absent value.

## Relationship to #3704

Independent and complementary; they touch different files and can merge in either order.

#3704 fixes the case where the acquisition date/time tags are **absent**, which is what makes RTSTRUCT fail 100% of the time. This PR fixes the case where they are **present but in a shape the parser did not accept**. Together they close both ways `GenericDatasetAcquisitionStrategy` could fail on a timestamp.

## Tests

Adds `DateTimeUtilsTest`, covering every TM form, one to six fractional digits, the previously accepted formats, trailing padding, null/empty/blank, and invalid values still raising. It also covers `pacsStringToLocalDate`.

This is the first test in `shanoir-ng-ms-common`. No pom change was needed: `spring-boot-starter-test` is already inherited from the parent.

```
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
```

Checkstyle passes on the module with `includeTestSourceDirectory` (0 violations).

### DIFF
--- a/shanoir-ng-ms-common/src/main/java/org/shanoir/ng/shared/dateTime/DateTimeUtils.java
+++ b/shanoir-ng-ms-common/src/main/java/org/shanoir/ng/shared/dateTime/DateTimeUtils.java
@@ -23,7 +23,8 @@ import java.time.LocalTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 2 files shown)


michaelkain (approved):
Hi @jbardet, nice PR! Thank you. I approve by code review. With kind regards, Michael

PR #3709, opened 2026-07-30
add missing indexes

julien-louis:
Table examination was missing indexes on its foreign keys.

Could lead to slow queries when retrieving studies volumes. 
This query was once listed in slow-queries.log.


michaelkain:
Hi @julien-louis, I hope you are doing fine. Thank you for your PR!
Great, that it fixes the indexes in prod, do you think it would be good to add it to JPA/Spring as well, if somebody generates a new database-scheme? With kind regards, Michael

julien-louis:
@michaelkain It's done !

PR #3710, opened 2026-07-30
issue/3586: Fix preclinical examination extra-data upload, download & display

### LINKED ISSUE #3586: Preclinical examination extra data issues
During the creation of a preclinical examination, there are the following issues:
1. When uploading a file on the examination as extra data, it uploads normally and we can see its name (check no 1 in screenshot)
2. When uploading extra data in the anesthetic field, the filename does not appear after the upload (is it uploaded or not?? we do not know, check no 2 nad 3)
Image

After the examination creation, if you open the examination:
1. You can see normally the examination extra data and download them (check no 1)
2. You can also see the extra data related to the anesthetci fieldset, bu you cannot actually download them (check no 2 and 3)
Image

fragkiadamis:
Fixes preclinical extra-data (physiological & blood-gas) file handling: corrects the storage/download path so files download, shows the selected filename with a remove button in the form, and makes the update services null-safe to stop the 500 when a record has no file.

Closes #3586


DuckflipXYZ (approved):
Looks ok by code review

PR #3712, opened 2026-07-31
issue/3706: Let OHIF display compressed DICOM exams

### LINKED ISSUE #3706: OHIF - selected TOF is not displayed on the viewer
On some examinations all the series are not displayed on OHIF and we stays on "Loading..." message with the error pop up showing "une erreur s'est produitefalse"

Image

fragkiadamis:
The OHIF frames proxy dropped the Content-Type returned by the PACS and never forwarded the viewer's Accept header, so compressed exams reached OHIF without the transfer syntax its decoder needs -- they failed with "Couldn't decode". PACS errors were also returned as 200 OK, hiding the real cause.

Closes #3706


fragkiadamis:
Hello @LaurentPV 

This is the fix for the OHIF issue you had at the exam:
https://shanoir.irisa.fr/shanoir-ng/examination/details/38260

The PR is currently deployed in qualif so you can actually test it if you want.

Best,
Adam

LaurentPV (approved):

PR #3714, opened 2026-08-03
task/3638: Remove empty dataset acquisitions after their datasets are deleted

### LINKED ISSUE #3638: Remove empty entities after children removal
Like, if we delete the unique dataset of an acquisition, the acq is not removed and stays empty.

fragkiadamis:
## Description

Deleting the last dataset of an acquisition used to leave that acquisition behind, empty and unreachable: nothing but the import creates an acquisition, and nothing can refill one once its last dataset is gone. It still showed up in the tree, in BIDS exports and as an empty series in the DICOMweb responses.

The deletion now offers to remove the acquisition too.

The removal is proposed, never silent. Before deleting, the front asks the backend which acquisitions the deletion would leave empty, names them in the confirmation dialog, and only then sends deleteEmptyAcquisitions=true. The flag defaults to false, so existing API clients (ShUp included) keep their current behaviour.

An acquisition is only removable when it holds nothing of its own:

no extra-data file uploaded on the acquisition itself, which the removal would destroy;
no copy pointing to it as its source.
Acquisitions failing either check are kept and never proposed for removal.

Only acquisitions are concerned. Examinations, subjects and studies are legitimately empty in the normal workflow (an examination is created before the import that fills it, and carries its own metadata and attached files), so they are left untouched.

## API

| Endpoint | Description |
|---|---|
| `POST /datasets/emptyAcquisitionsPreview` | acquisitions the deletion of the given datasets would leave empty |
| `DELETE /datasets/{id}?deleteEmptyAcquisitions=true` | delete a dataset, remove its acquisition if left empty |
| `DELETE /datasets/delete?deleteEmptyAcquisitions=true` | same for a batch of datasets |
| `GET /datasetacquisition/empty?studyId=` | admin only, the empty acquisitions already in database |
| `DELETE /datasetacquisition/empty?studyId=` | admin only, removes them |

The two last ones clean up the acquisitions emptied before this change: the fix alone only
prevents new ones. The `GET` is a dry run, and the front exposes them as an
*Administration > Clean up empty acquisitions* menu entry that lists what would be deleted before
asking for confirmation.

Closes #3638


fragkiadamis:
Hello @DuckflipXYZ 

This PR is the implementation of the issue you proposed concerning the auto dataset-acquisition deletion if a deleted dataset is the only child of that entity.

It has been tested only locally

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/controler/DatasetApi.java
```diff
@@ -82,9 +85,23 @@ ResponseEntity<Void> deleteDataset(
...
 
+    @Operation(summary = "", description = "Returns the dataset acquisitions that the deletion of the given datasets would leave empty, and that may then be removed")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "found dataset acquisitions"),
+        @ApiResponse(responseCode = "401", description = "unauthorized"),
+        @ApiResponse(responseCode = "403", description = "forbidden"),
+        @ApiResponse(responseCode = "500", description = "unexpected error")})
+    @PostMapping(value = "/emptyAcquisitionsPreview", produces = {"application/json"}, consumes = {"application/json"})
+    @PreAuthorize("hasRole('ADMIN') or (hasRole('EXPERT') and @datasetSecurityService.hasRightOnEveryDataset(#datasetIds, 'CAN_ADMINISTRATE'))")
+    ResponseEntity<List<ExaminationDatasetAcquisitionDTO>> findAcquisitionsLeftEmptyBy(
```
DuckflipXYZ: Usually, we locate an API method in the APIfile of the returned type, not in the APIfile of the request body. That's a minor thing

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/repository/DatasetRepository.java
```diff
@@ -61,6 +61,13 @@ public interface DatasetRepository extends PagingAndSortingRepository<Dataset, L
 
     Iterable<Dataset> findByDatasetAcquisitionId(Long acquisitionId);
 
+    int countByDatasetAcquisitionId(Long acquisitionId);
+
+    @Query("SELECT ds.datasetAcquisition.id, COUNT(ds.id) FROM Dataset ds "
+            + "WHERE ds.datasetAcquisition.id IN (:acquisitionIds) "
+            + "GROUP BY ds.datasetAcquisition.id")
+    List<Object[]> countByDatasetAcquisitionIdIn(List<Long> acquisitionIds);
```
DuckflipXYZ: Since it's not a part of code that much regularly, it's not required to optimize it as much as possible. It would be cleaner to use the method countByDatasetAcquisitionId above in a for loop, in a transactionnal method (can be readOnly). The time loss would be negligeable. This is a minor thing too

shanoir-ng-front/src/app/datasets/dataset/dataset.component.ts
```diff
@@ -60,6 +62,31 @@ export class DatasetComponent extends EntityComponent<Dataset> {
...
+        if (!this.goToExaminationOfRemovedAcquisition()) super.goBack();
+    }
+
+    /**
+     * Deleting the last dataset of an acquisition removes that acquisition too, so its page does
+     * not exist anymore : the examination is then where to go once the dataset is deleted.
+     *
+     * @return true when the acquisition was removed and the examination reached
+     */
+    private goToExaminationOfRemovedAcquisition(): boolean {
```
DuckflipXYZ, line 83: Since it's a behaviour that should be generic when this context of deletion occurs, maybe it would be interesting to manage it in a generic way instead. In the tree.service.ts, there is the method below which is called at entity deletion from front : 

    goToParent() {
        if (this.selectedNode instanceof StudyNode) {
            this.router.navigate(['/study/list']);
        } else {
            this.router.navigate([this.selectedNode?.parent?.route]);
        }
    }

I'm not sure how to deal with it, but maybe it would be relevant to at least think of it.

This is a minor thing too.

shanoir-ng-front/src/app/datasets/shared/dataset.service.ts
```diff
@@ -52,12 +53,59 @@ export class DatasetService extends EntityService<Dataset> {
...
+            body: JSON.stringify(ids),
+            params: new HttpParams().set('deleteEmptyAcquisitions', deleteEmptyAcquisitions)
+        }));
+    }
+
+    /**
+     * The acquisitions that the deletion of the given datasets would leave empty, and that the
+     * backend would then be able to remove. An empty list means the deletion leaves nothing behind.
+     */
+    getAcquisitionsLeftEmptyBy(datasetIds: number[]): Promise<IdName[]> {
```
DuckflipXYZ: Same as for the API, should be more common to call the acquisitionService which is implementing this method. Still minor issue

shanoir-ng-front/src/app/datasets/shared/dataset.service.ts
```diff
@@ -52,12 +53,59 @@ export class DatasetService extends EntityService<Dataset> {
...
+                    .then(deleted => {
+                        // the acquisitions are gone as well : whoever displays them has to know
+                        if (deleted) this.onAcquisitionsRemoved.next(emptied.map(acquisition => acquisition.id));
+                        return deleted;
+                    });
+            });
+    }
+
+    getEmptyAcquisitionsMessage(acquisitions: IdName[]): string {
+        if (acquisitions.length == 1) {
```
DuckflipXYZ: Same as above

shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/controler/DatasetAcquisitionApi.java
```diff
@@ -86,6 +87,30 @@ ResponseEntity<Void> deleteDatasetAcquisition(
...
+            @Parameter(description = "id of the study to clean up, all of them when not given")
+            @RequestParam(value = "studyId", required = false) Long studyId);
+
+    @Operation(summary = "", description = "Deletes the dataset acquisitions that hold no dataset at all and that may be removed")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "dataset acquisitions deleted"),
+        @ApiResponse(responseCode = "401", description = "unauthorized"),
+        @ApiResponse(responseCode = "403", description = "forbidden"),
+        @ApiResponse(responseCode = "500", description = "unexpected error")})
+    @RequestMapping(value = "/datasetacquisition/empty", produces = {"application/json"}, method = RequestMethod.DELETE)
```
DuckflipXYZ: Would it be a @DeleteMapping there ? In that way of use, you would still be able to access to the APIController file without any automatic routing due to the tag usage

shanoir-ng-datasets/src/main/java/org/shanoir/ng/datasetacquisition/controler/DatasetAcquisitionApiController.java
```diff
@@ -253,6 +256,30 @@ public ResponseEntity<Void> deleteDatasetAcquisition(
...
+    public ResponseEntity<List<ExaminationDatasetAcquisitionDTO>> findEmptyDatasetAcquisitions(Long studyId) {
+        List<DatasetAcquisition> empty = datasetAcquisitionService.findEmptyAcquisitions(studyId);
+        if (CollectionUtils.isEmpty(empty)) {
+            return new ResponseEntity<>(Collections.emptyList(), HttpStatus.OK);
+        }
+        return new ResponseEntity<>(examDsAcqMapper.datasetAcquisitionsToExaminationDatasetAcquisitionDTOs(empty), HttpStatus.OK);
+    }
+
+    @Override
+    public ResponseEntity<List<Long>> deleteEmptyDatasetAcquisitions(Long studyId) {
```
DuckflipXYZ, line 278: Maybe an Shanoir event for that method would be nice.

DuckflipXYZ (approved):
All good, some possible improvements proposed

shanoir-ng-front/src/app/datasets/dataset/dataset.component.ts
fragkiadamis: It will not work. When you delete the last dataset of an acquisition, the acquisition gets deleted. The dataset's parent is that deleted acquisition therefore you have an error. I implemented it like that at first and I was having errors since the entity was deleted. So it has to go a level above parent.

PR #3722, opened 2026-08-04
feature/2758: Create a task when deleting a subject

### LINKED ISSUE #2758: Delete subjects should create a task

fragkiadamis:
Add a job in the job list when deleting a subject.

Closes #2758


shanoir-ng-datasets/src/main/java/org/shanoir/ng/examination/service/ExaminationServiceImpl.java
```diff
@@ -112,6 +112,7 @@ public class ExaminationServiceImpl implements ExaminationService {
     private StorageService storageService;
 
     @Override
+    @Transactional
```
DuckflipXYZ: What's the point of that @Transactionnal ?

DuckflipXYZ (approved):
All good

Issue #3049, opened 2025-09-25
Add new metadata sliceThickness and Number of slices to quality control rules


LaurentPV:
In Shanoir Dataset Metadata list of a quality card condition, add : 

- Slice Thickness
- Number of slices

As those metadata are more recent we need to find a way to apply quality cards on dataset that do not have those values without setting an ERROR tag by default and damaging the quality control results.

PR #3756, opened 2026-08-18
Issue#3049 quality card acq condition new fields

LaurentPV:
Linked to https://github.com/fli-iam/shanoir-ng/issues/3049

This PR adds the two Shanoir metadata fields Slice Thickness and Number Of Slices to the Quality Card condition fields.
When selected, only numerical comparators are available.

Capture d’écran 2026-08-18 à 11 29 04

### DIFF
--- a/shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
+++ b/shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
@@ -17,7 +17,7 @@ import { FormsModule } from '@angular/forms';
 
 import { Mode } from '../../../shared/components/entity/entity.component.abstract';
 import { Option, SelectBoxComponent } from '../../../shared/select/select.component';
-import { MetadataFieldScope, StudyCardAssignment } from '../../shared/study-card.model';
+import { FieldType, MetadataFieldScope, StudyCardAssignment } from '../../shared/study-card.model';
 import { AutoAdjustInputComponent } from '../../../shared/auto-ajust-input/auto-ajust-input.component';
 
 
@@ -169,7 +169,7 @@ export class ShanoirMetadataField {
             public field: string,
             public scope: MetadataFieldScope,
             options?: Observable<Option<any>[]> | Option<any>[],
-            public numeric: boolean = false) {
+            public type: FieldType = 'String') {
 
         if (options instanceof Observable) {
             this.options = options;
--- a/shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
+++ b/shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
@@ -165,10 +165,11 @@ export class ShanoirMetadataField {
     public options?: Observable<Option<any>[]>;
 
     constructor(
-            public label: string, 
+            public label: string,
             public field: string,
-            public scope: MetadataFieldScope, 
-            options?: Observable<Option<any>[]> | Option<any>[]) {
+            public scope: MetadataFieldScope,
+            options?: Observable<Option<any>[]> | Option<any>[],
+            public numeric: boolean = false) {
 
         if (options instanceof Observable) {
             this.options = options;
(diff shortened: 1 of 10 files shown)


LaurentPV:
For now the behavior is unchanged on missing fields. As those fields came afterwards a lot of data have them set to null, in this case the condition is not fulfilled.

shanoir-ng-front/src/app/study-cards/study-card-rules/condition/condition.component.ts
```diff
@@ -104,6 +104,8 @@ export class StudyCardConditionComponent implements OnInit, OnDestroy, OnChanges
             validators.push(Validators.pattern(this.buildArrayPattern(vm, 'float')));
         } else if ( type == 'IntArray') {
             validators.push(Validators.pattern(this.buildArrayPattern(vm, 'int'))); // comma separated integers
+        } else if (!type && this.fields?.find(f => f.field == this.condition?.shanoirField)?.numeric) {
```
julien-louis: then just
```typescript
const type: TagType = undefined;
if (this.condition?.dicomTag) {
    type = this.condition.dicomTag.type;
} else if (this.condition?.shanoirField) {
    type = this.condition.shanoirField.type;
}
```

shanoir-ng-front/src/app/study-cards/study-card-rules/condition/condition.component.ts
```diff
@@ -370,7 +372,17 @@ export class StudyCardConditionComponent implements OnInit, OnDestroy, OnChanges
                 this.operations.forEach(op => op.disabled = false);
             }
         } else { // Shanoir fields
-            if (this.shanoirFieldOptions?.length > 0) { // with option list such as coils
+            const shanoirField: ShanoirMetadataField = this.fields?.find(f => f.field == this.condition.shanoirField);
```
julien-louis: could we deal with the 'type' as intended so we can avoid these workarounds ?

julien-louis, line 374: why not directly this.condition.shanoirField ?

shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
```diff
@@ -165,10 +165,11 @@ export class ShanoirMetadataField {
...
 
     constructor(
-            public label: string, 
+            public label: string,
             public field: string,
-            public scope: MetadataFieldScope, 
-            options?: Observable<Option<any>[]> | Option<any>[]) {
+            public scope: MetadataFieldScope,
+            options?: Observable<Option<any>[]> | Option<any>[],
+            public numeric: boolean = false) {
```
julien-louis: maybe use TagType instead of `numeric: boolean` as I think it's a simplification of DICOM types and could be applied to metadata fields as well. You could rename TagType as CardValueType or something like that. The default could be String.

julien-louis (changes requested):
Hi @LaurentPV I did a suggestion but it's up to you. The only point that bothers me is the `his.fields?.find(f => f.field == this.condition?.shanoirField)?.numeric` instead of just `this.condition?.shanoirField`

shanoir-ng-front/src/app/study-cards/study-card-rules/condition/condition.component.ts
LaurentPV: I wanted to do that at first but this.condition.shanoirField has no attribute .type, it is a String corresponding to the field selected by the user. For the syntax line 107 I just copied what was already done for conditionField line 206.

shanoir-ng-front/src/app/study-cards/study-card-rules/condition/condition.component.ts
LaurentPV: Same reason as before, it is because shanoirField is a String and not the ShanoirMetadataField object.

shanoir-ng-front/src/app/study-cards/study-card-rules/action/action.component.ts
LaurentPV: This modification was implemented in the last commit.

julien-louis (approved):

PR #3775, opened 2026-08-27
#3758 : increasing thread parallelization management for dcm4chee

### LINKED ISSUE #3758: Time out and HTTP 401 when too much download simultaneously from VIP
On 200 parallelized SIMS jobs with OSIV false, after 30 secs of zipOutputStream open for VIP download, we got a time out + http 401 for some.

It looks like it's coming from Shanoir. Here are some possibilities : 
- Max number of parallelized HTTP request to Shanoir limited
- Max number of parallelized HTTP request to DCM4Chee limited (logs in /opt/wildfly/standalone/log
- Disk saturation (at least on qualif)

DuckflipXYZ:


shanoir-ng-datasets/src/main/java/org/shanoir/ng/vip/executionTemplate/service/PlannedExecutionManager.java
```diff
@@ -51,7 +51,7 @@ public class PlannedExecutionManager {
     private static final Logger LOG = LoggerFactory.getLogger(PlannedExecutionManager.class);
 
     private  int maxStatusRetries = 3;
-    private long statusSleepSeconds = 20;
+    private long statusSleepSeconds = 2000;
```
LaurentPV: This is used twice, on line 144 where it will wait 2 seconds when the queue is empty and in line 229 where the value is set in seconds (so 33 minutes) in the status polling loop of VIP, is it what you wanted ? it seems long. maybe set 2 different variables or set line 229 in milliseconds.

docker-compose/dcm4chee/tune-wildfly-io-worker.sh
```diff
@@ -0,0 +1,14 @@
...
+# by the parent script (see entrypoint-wrapper.sh).
+#
+# dcm4chee-arc.xml ships with an unbounded Undertow IO worker
+# (<worker name="default"/>), so io-threads/task-max-threads default to values
+# computed from the container's visible CPU count -- under-provisioned on a
+# CPU-constrained host, and not controllable via env vars unlike the PacsDS
+# pool (WILDFLY_PACSDS_MAX_POOL_SIZE). This pins explicit, tunable values.
+CONF="$JBOSS_HOME/standalone/configuration/dcm4chee-arc.xml"
+if [ -f "$CONF" ] && ! grep -q 'io-threads=' "$CONF"; then
+    sed -i 's#<worker name="default"/>#<worker name="default" io-threads="${env.WILDFLY_IO_THREADS:16}" task-max-threads="${env.WILDFLY_IO_TASK_MAX_THREADS:200}"/>#' "$CONF"
```
LaurentPV: minor comment here : if dcm4chee image change in the future this sed command could fail without warning and the tuning wouldn't apply. Maybe set a warning log if the sed command fails or if 'io-threads=' is not in the file after the sed.

LaurentPV (changes requested):

a-ba (changes requested):
My overall feeling is that we are not optimising the right thing here.

The main bottleneck that makes the massive download slow is the loop in `downloadDicomFilesForURLsAsZip()`. It gets the pacs objects one by one, and therefore it is impacted by the latency of the pacs. To make it efficient, the download process should be pipelined so a to prefetch a few objects in advance and to ensure that the output buffer is never cleared. This is what should be optimised **FIRST**. This would make the massive download feature faster for all users and remove the need to handle hundreds of parallel downloads.

Other notes regarding memory consumption:
- `SHANOIR_PACS_MAX_IN_MEMORY_BYTES` configures the buffer size of the WebClient to 64Mo (instead of the default 256Ko). Combined with `WILDFLY_IO_TASK_MAX_THREADS` which is set to 1000, this may potentially consume 62.5Go of RAM (the current production instances are configured with 12Go, now i understand why we regularly have memory peaks)

- `DicomInputStream` also consumes memory because it makes a copy of the dicom data in the returned structure.

- the MIME handler also consumes memory because it buffers objects

Increasing the number of massive downloads implies duplicating the whole chain multiple times, with lot of memory allocated and all its components stalled most of the time.

On the other hand if we correctly parallise the loop in `downloadDicomFilesForURLsAsZip()` we only need a few tasks per connection (because the connections are immediately reused: getting an object is a matter of a few dozen of milliseconds) and keep the number of DicomInputStream and mime parsers low (and the other tuned parameters will be irrelevant).

PR #3780, opened 2026-08-28
Fix quality tags severity order

LaurentPV:
This fixes 2 bugs identified with Angular 22 and OSIV testing : 

- Quality card rules were overwriting each others so only ordering them in the frontend could alter the quality tag applied in case of concurrent conditions. 
Now the more severe tag is applied if the condition is fulfilled.

- There was a bug due to Angular 22 migration where the quality card condition subForm if changed or deleted stayed in the parent FormGroup as invalid instead of being replaced. The result is that quality cards were impossible to modify and the update button was always greyed out without any error being visible.


DuckflipXYZ (approved):
All good from code review

PR #3784, opened 2026-09-01
ShUp - Fixes tmp folder deletion and improve write import-job.json method

### LINKED ISSUE #3665: ShUp : import failure error "folder already exists."
an OFSEP user having the import failure error "folder already exists." noticed that, after searching with the exam date (he is connected to the hospital PACS), if he selects the exam, the import fails and downloading the exam is really long. If he selects all the series of the exam, the import works.

LaurentPV:
Fixes https://github.com/fli-iam/shanoir-ng/issues/3665

This PR fixes the previous use of file.delete() in cleanTempFolders(), this method is not returning an Exception but simply false if executed on a non empty folder. In case of interrupted write of .dcm files (long download that fails like described in issue) the folder is not cleaned and exists in case of reattempt at downloading the same dicom study.

This PR contains also a modification of the writeImportJobJson() method that is now atomic to avoid also corrupted .json files (file is now written as a tmp file in same path and then moved once as import-job.json file)

### DIFF
--- a/shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/NominativeDataImportJobManager.java
+++ b/shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/NominativeDataImportJobManager.java
@@ -16,8 +16,6 @@ package org.shanoir.uploader.nominativeData;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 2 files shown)


shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/NominativeDataImportJobManager.java
```diff
@@ -94,13 +96,22 @@ public ImportJobBase readImportJob() {
...
 
+    /**
+     * Writes the import job to a temp file next to the target, then atomically moves it into place.
+     * Made to avoid parsing failure on the next readImportJob() call if an event (crash, forced kill, sleep/shutdown)
+     * left it truncated/corrupted.
+     */
     public void writeImportJob(ImportJobBase importJob) {
         ReentrantLock lock = getLock();
         lock.lock();
+        File tmpFile = new File(this.nominativeDataJobFile.getParentFile(), this.nominativeDataJobFile.getName() + ".tmp");
```
michaelkain: Hi @LaurentPV, this would probably create an import-job.json.tmp, what might offend Windows? In general I am not a big fan of this approach, I would solve it on the readImportJob side and better ignore corrupted json-files. The file delete in case of an exception, what if ShUp crashes in between, it leaves a tmp-file or a half-full tmp file, that is never copied. With kind regards, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/utils/FileUtil.java
```diff
@@ -124,10 +124,15 @@ public static void copyFile(File sourceFile, File destFile) {
     //     }
     // }
 
-    public static void cleanTempFolders(File workFolder, String studyInstanceUID) {
+    public static void cleanTempFolders(File workFolder, String studyInstanceUID) throws IOException {
         File tempStudyInstanceUIDFolder = new File(workFolder, studyInstanceUID);
         if (tempStudyInstanceUIDFolder.exists()) {
-            tempStudyInstanceUIDFolder.delete();
```
michaelkain, line 130: Hi @LaurentPV, same here, I do not real see any problem solving here, only more code, and I do not achieve to link to the original problem, what is it exactly. With kind regards, Michael

michaelkain: Hi @LaurentPV, SORRY! I went back into the JavaDoc and you are right, the tree deletion is much better, sorry, this one we have to merge. Cheers, Michael

shanoir-uploader/src/main/java/org/shanoir/uploader/nominativeData/NominativeDataImportJobManager.java
LaurentPV: Ok, i rolled back this modification with the last commit.

michaelkain (approved):
Hi @LaurentPV, thank you! I approve by code review only. Cheers, Michael

PR #3798, opened 2026-09-07
issue/3764: Do not publish ShanoirEvent in cascade deletions

### LINKED ISSUE #3764: Delete examinaton flood the side bar with lots of acquisition deletion events
For me, it's related to that PR : https://github.com/fli-iam/shanoir-ng/pull/3714

It's coming from a mail from Elise, which is saying : 

"This is a very very minor point but when suppressing an examination, the following left column has one line per series. This may not be necessary since the series are not distinguished. Maybe this could also be confusing. When import is ongoing the progress for the whole examination is tracked, not series by series. Hence my surprise. 
No major topic but maybe this could be simplified ? "

Image

fragkiadamis:
This PR introduces code to avoid the event publishing of cascade deletions.

For example, when we delete an examination we only get the `DELETE_EXAMINATION` event and not the `DELETE_DATASET_ACQUISITION` and `DELETE_DATASET` events.

Closes #3764


fragkiadamis:
Hello @DuckflipXYZ 

A small PR that fixes teh issue you gave me 2 weeks ago. Review and merge when you have the time.

Best,

fragkiadamis:
Hello @DuckflipXYZ and @LaurentPV 

This is also a fix for the cascade delete of an examination. I have already tested it, so if one of you could have a review to include it in the hotfix release, that would be great

Thanks!

shanoir-ng-datasets/src/main/java/org/shanoir/ng/dataset/service/DatasetService.java
```diff
@@ -56,6 +56,19 @@ public interface DatasetService {
...
+     *
+     * @param id dataset id.
+     * @param cascade true when the deletion is driven by the deletion of a parent (acquisition,
+     *                examination...), which already reports its own progress: no deletion event
+     *                is then published for this dataset.
+     * @throws EntityNotFoundException
+     * @throws ShanoirException
+     */
+    @PreAuthorize("hasRole('ADMIN') or (hasRole('EXPERT') and @datasetSecurityService.hasRightOnDataset(#id, 'CAN_ADMINISTRATE'))")
+    void deleteById(Long id, boolean cascade) throws EntityNotFoundException, ShanoirException, SolrServerException, IOException, RestServiceException;
```
LaurentPV: why not mutualizing with the deleteByIdCascade() below already called by deleteDatasetAcquisition() ?

fragkiadamis: My main concern is that in the deleteByIdCascade in DatasetAcquisitionService.java, is that we do not have the following control:

`@PreAuthorize("hasRole('ADMIN') or (hasRole('EXPERT') and @datasetSecurityService.hasRightOnDataset(#id, 'CAN_ADMINISTRATE'))")`

LaurentPV (approved):

PR #3815, opened 2026-09-10
issue/3806: Restore DICOM metadata node on dataset processing outputs 

### LINKED ISSUE #3806: DatasetProcessing: We cannot access the DICOM metadata for the datasets of the DatasetProcessing entities anymore from the UI
Normally, when collapsing the datasets of a DatasetProcessing entity, we should be able to access the DICOM metadata. From the new release this is not possible anymore

It should look like this:

Image

But it looks like this:

Image

fragkiadamis:
Fixes #3806


fragkiadamis:
Hello @LaurentPV and @DuckflipXYZ 

This is my final fix for the upcoming hotfix. It is tested on neurinfo qualif, please review and merge to include it.

Best,

LaurentPV (approved):

PR #3816, opened 2026-09-10
passing userId to child event and adding unit test

### LINKED ISSUE #3812: RabbitMQ not passing the username to child events when deleteSubject()
When deleting a subject, the event has the parameter userId but it is not transmitted to  examinationService.deleteById() and child events deleting datasetAcquisition, resulting in history events difficult to trace.

Image

LaurentPV:
Fixes #3812


fragkiadamis (approved):
Approved by code review.

PR #3817, opened 2026-09-10
issue/3813: Fix VIP execution monitoring reporting to the wrong user

### LINKED ISSUE #3813: ExecutionMonitoring: wrong user relation
While a user A runs an execution, and another user B starts an execution in the meantime, the execution monitoring of user B will be related to user A that already runs an execution.

User B does not get any infromation on his execution
User A sees random executions that are not theirs
If User B uses py_noir and he needs the execution monitoring for the tracking file, they will get nothing and the tracking file will have null values.

fragkiadamis:
The VIP monitoring loop is shared by all users, but it is started by whoever queues the first execution and the async executor pins *that* user's security context to the loop thread for its whole life. Every execution queued afterwards had its ShanoirEvent created under that foreign identity -- so user B saw nothing and user A saw executions that were not theirs.                                                                                                                                                              

Each job is now polled under a system context, with the owner's id carried explicitly on the queue entry for the sole purpose of addressing its event. Also fixes a race where an execution queued exactly as the loop exited was left in the queue, never polled at all.

Fixes #3813


fragkiadamis:
Fixed by #3818

PR #3819, opened 2026-09-10
task/dataset-processing-dto: Include parentId in the DatasetProcessing DTO

fragkiadamis:
Fixes errors concerning the datasetProcessing donwload for eCAN through py_noir.

If any questions, ask me!

@DuckflipXYZ I am going to need this PR asasp, so please review when you can
Thanks

### DIFF
--- a/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/repository/DatasetProcessingRepository.java
+++ b/shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/repository/DatasetProcessingRepository.java
@@ -16,6 +16,7 @@ package org.shanoir.ng.processing.repository;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.hibernate.Hibernate;
 import org.shanoir.ng.processing.model.DatasetProcessing;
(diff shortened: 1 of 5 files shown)


shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/repository/DatasetProcessingRepository.java
```diff
@@ -147,11 +147,31 @@ public interface DatasetProcessingRepository extends CrudRepository<DatasetProce
...
+     *
+     * @param ids
+     * @return
+     */
+    @Transactional(readOnly = true)
+    default List<DatasetProcessing> findByIdsWithInputsAndOutputs(List<Long> ids) {
+        List<DatasetProcessing> processings = findByIdsWithInputs(ids);
+        findByIdsWithOutputs(ids);
+        return processings;
+    }
```
DuckflipXYZ, line 187: Curious to proceed that way, but why not i guess

DuckflipXYZ (approved):
Looks correct by code review

shanoir-ng-datasets/src/main/java/org/shanoir/ng/processing/repository/DatasetProcessingRepository.java
fragkiadamis: Reworked it to be less cryptic — you were right that the discarded call needed explaining. It's now a Hibernate.initialize on outputDatasets after the input fetch, same idiom as findAllByInputDatasets_IdInWithOutputsAndDatasetFile.

PR #3824, opened 2026-09-15
ci: free disk space before building docker images

LaurentPV:
The ubuntu-latest runner only has ~14GB of free disk space, which is regularly not enough to build all the shanoir-ng docker images, causing intermittent "no space left on device" failures. 

This PR reclaim space (Android/.NET/Haskell runtimes, large apt packages, unused docker images, swap) before the build with the community jlumbroso/free-disk-space action.


DuckflipXYZ (approved):
All good, at least it does not break the build step

Issue #3520, opened 2026-05-11
Turn open-in-view to false in dataset MS


DuckflipXYZ:
The objective is to free DB connections way faster than before, they'll be released at the end of transaction, instead of at the end of java thread.

PR #3826, opened 2026-09-15
fix(examinations): restore read-write transaction on deleteById

jbardet:
## Summary

- Fixes a regression introduced in #3520 (commit 28bbd0415) where `ExaminationServiceImpl.deleteById()` was annotated with `@Transactional(readOnly = true)` while performing DELETE operations.
- Subject deletion triggers asynchronous cleanup in ms datasets: examinations must be deleted before the subject row. With a read-only transaction, examination deletes do not persist, and subject deletion fails with a foreign key constraint on `examination.subject_id`.

## Impact

Observed on NG_v2.14.0: subject delete tasks report success in ms studies ("Subject X removed") but fail in ms datasets with:
```
Cannot delete or update a parent row: a foreign key constraint fails (datasets.examination, ... FOREIGN KEY (subject_id) REFERENCES subject (id))
```

Examination rows remain in the datasets DB; the task report often contains only the subject error (no per-examination errors).

This previously worked: commit 6c383b67d had correctly restored `@Transactional` on this method before the #3520 regression.


fragkiadamis:
Hey @jbardet thanks for this.

It's already solved in non-merged PR, but we will keep yours in order to keep the PRs scoped.

Best,

fragkiadamis (approved):

PR #3828, opened 2026-09-15
feature/3827: Dataset Processing deletion event

### LINKED ISSUE #3827: Create an event for the DatasetProcessing entity deletion
When we delete a DatasetProcessing, there is not event in the task list.

fragkiadamis:
Gives dataset processing deletion an event of its own (`deleteDatasetProcessing.event`)

Closes #3827


DuckflipXYZ (approved):
All good !

PR #3832, opened 2026-09-16
issue/3831: Refresh the tree when a dataset processing is deleted

### LINKED ISSUE #3831: Tree is not refreshing after DatasetPRocessing entity deletion
When we delete a dataset processing entity, the tree does not refresh and as a result the entity is still visible (even though it's deleted).

fragkiadamis:
Deleting a dataset processing from its details page left the node in the tree, unlike every other entity. The tree service refused the selection outright for any processing without a parent monitoring, so `selectedNode` kept pointing at the previously selected node -- which `removeCurrentNode()` then removed instead, or silently did nothing.

Fixes #3831


fragkiadamis:
Hello @DuckflipXYZ 

This is the PR that fixes the tree during the dataset processing deletion.

PR #3838, opened 2026-09-17
fix DatasetAcquisitionWithDatasetsMapper to retrieve centerId/studyId

LaurentPV:
This PR fixes the application of Study Cards from SolR Search menu when user has not role ADMIN.

Test : 
- in SolR Search menu select datasets from a study and center you have rights on
- click on Apply Study Card and select the study card you want to apply, compatible or not
- The application should run fine and the fields of the da/datasets must be modified according to the study card rules.


DuckflipXYZ (approved):
All good

